PR #1, opened 2016-12-13
A timer which logs slow requests

shuzhang1989:
A timer which can be used in services needing to log slow requests.

### DIFF
--- a/common/slow_log_timer.h
+++ b/common/slow_log_timer.h
@@ -1,0 +1,75 @@
+/// Copyright 2016 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author shu (shu@pinterest.com)
+//
+
+#pragma once
+
+#include <glog/logging.h>
+#include <sys/time.h>
+#include <stdlib.h>
+#include <string>
+
+#include "common/stats/stats.h"
+#include "common/timer.h"
+
+namespace common {
+
+// A timer which uses GLOG to log slow requests (long Timer object life cycle)
+// and logs the content with a life cycle time threshold and a sample rate.
+class SlowLogTimer : public common::Timer {
+ public:
+  explicit SlowLogTimer(const uint32_t metric,
+                        const std::string& log_message,
+                        uint64_t log_latency_threshold=-1,
+                        double log_sample_rate=0)
+    : Timer(metric), log_message_(log_message),
+      log_latency_threshold_(log_latency_threshold),
+      log_sample_rate_(log_sample_rate) {}
+
+  explicit SlowLogTimer(const std::string& metric,
+                        const std::string& log_message,
+                        uint64_t log_latency_threshold=-1,
+                        double log_sample_rate=0)
+    : Timer(metric), log_message_(log_message),
+      log_latency_threshold_(log_latency_threshold),
+      log_sample_rate_(log_sample_rate) {}
+
+  // stop the clock and report the delta through metric_[str|int]_
+  // also log the string content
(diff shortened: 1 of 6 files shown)


shuzhang1989:
@newpoo i will need this to log slow requests.

common/slow_log_timer.h
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  bool shouldLog() {
+    auto elapsed_time = getElapsedTimeMs();
+    return generatereRandomDouble0To1() <= log_sample_rate_ &&
+            elapsed_time > log_latency_threshold_;
+  }
+
+  const std::string log_message_;
+  const uint64_t log_latency_threshold_;
```
newpoo: append "ms_"

common/slow_log_timer.h
```diff
@@ -0,0 +1,75 @@
...
+#include "common/timer.h"
+
+namespace common {
+
+// A timer which uses GLOG to log slow requests (long Timer object life cycle)
+// and logs the content with a life cycle time threshold and a sample rate.
+class SlowLogTimer : public common::Timer {
+ public:
+  explicit SlowLogTimer(const uint32_t metric,
+                        const std::string& log_message,
```
newpoo: change it to "std::string log_message" and use "std::move()" at line 39 so that we could potentially use move semantic.

common/slow_log_timer.h
```diff
@@ -0,0 +1,75 @@
...
+#include <string>
+
+#include "common/stats/stats.h"
+#include "common/timer.h"
+
+namespace common {
+
+// A timer which uses GLOG to log slow requests (long Timer object life cycle)
+// and logs the content with a life cycle time threshold and a sample rate.
+class SlowLogTimer : public common::Timer {
```
shuzhang1989: done

common/slow_log_timer.h
shuzhang1989: done

common/slow_log_timer.h
```diff
@@ -0,0 +1,75 @@
...
+  // also log the string content
+  ~SlowLogTimer() {
+    if (shouldLog()) {
+      LOG(WARNING) << "Slow request: " << log_message_;
+    }
+  }
+
+ protected:
+  double generatereRandomDouble0To1() {
+    return rand() / (RAND_MAX + 1.);
```
shuzhang1989: done

common/slow_log_timer.h
```diff
@@ -0,0 +1,83 @@
...
+
+  // stop the clock and report the delta through metric_[str|int]_
+  // also log the string content
+  virtual ~SlowLogTimer() {
+    if (shouldLog()) {
+      LOG(WARNING) << "Slow request: " << log_message_;
+    }
+  }
+
+ protected:
```
newpoo: can we make it private?

common/slow_log_timer.h
```diff
@@ -0,0 +1,83 @@
...
+
+namespace common {
+
+// A timer which uses GLOG to log slow requests (long Timer object life cycle)
+// and logs the content with a life cycle time threshold and a sample rate.
+class SlowLogTimer : public common::Timer {
+ public:
+  explicit SlowLogTimer(const uint32_t metric,
+                        std::string log_message,
+                        uint64_t log_latency_threshold_ms=-1,
```
newpoo: nit: can we add a space before and after = ? and also the other three lines below.

common/slow_log_timer.h
```diff
@@ -0,0 +1,83 @@
...
+    thread_local uint64_t should_log_count = 1;
+    should_log_count ++;
+    auto elapsed_time = getElapsedTimeMs();
+    return (should_log_count % log_sample_trigger_ == 0) &&
+            elapsed_time > log_latency_threshold_ms_;
+  }
+
+  const std::string log_message_;
+  const uint64_t log_latency_threshold_ms_;
+  const uint64_t log_sample_trigger_;
```
newpoo: Can we just let client to pass it in, and rename it to something like "log_one_for_every_n_slow_requests_".

common/timer.h
```diff
@@ -50,7 +50,7 @@ class Timer {
     }
   }
 
- private:
+ protected:
```
newpoo: do we have to make it protected? anything we want to use in it's extended class.

shuzhang1989: another is now() function

common/slow_log_timer.h
shuzhang1989: in test we need to inherit the methods

PR #8, opened 2016-12-19
Replication

newpoo:
This PR implement two new modes for replication.
Mode 1: ack client after committing to the master and write the update to the tcp connection to at least one of the slaves.
Mode 2: ack client after committing to the master and at least one of the slaves.

Besides the new unittests, I also manually tested it on the test cluster.

### DIFF
--- a/rocksdb_replicator/max_number_box.cpp
+++ b/rocksdb_replicator/max_number_box.cpp
@@ -61,6 +61,10 @@ void MaxNumberBox::post(const uint64_t num) {
 }
 
 bool MaxNumberBox::wait(const uint64_t num, const uint64_t timeout_ms) {
+  // We use thread local waiter because 1) avoid constructor and destructor
+  // overhead of Waiter::cv; 2) It is possible that the thread calling wait()
+  // returns from wait() before the thread calling post() calls notify_one()
+  // on Waiter::cv.
   thread_local Waiter me;
 
   std::unique_lock<std::mutex> lk(mtx_);
--- a/rocksdb_replicator/max_number_box.cpp
+++ b/rocksdb_replicator/max_number_box.cpp
@@ -1,0 +1,102 @@
+/// Copyright 2016 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


newpoo:
kick build

rocksdb_replicator/max_number_box.h
```diff
@@ -0,0 +1,84 @@
...
+    , waiters_() {}
+
+  ~MaxNumberBox() {
+    // it's client's responsibility to ensure there is no pending calls when the
+    // destructor is called.
+    CHECK(waiters_.empty());
+  }
+
+  // no copy or move
+  MaxNumberBox(const MaxNumberBox&) = delete;
```
shuzhang1989: if it is move constructor, should it be MaxNumberBox&&

rocksdb_replicator/max_number_box.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "folly/Likely.h"
+
+namespace replicator { namespace detail {
+
+void MaxNumberBox::post(const uint64_t num) {
+  std::vector<Waiter*> waiters_to_notify;
+  std::vector<Waiter*> waiters_to_wait;
+
+  {
+    std::unique_lock<std::mutex> lk(mtx_);
```
shuzhang1989: in highqps condition, will mutex be less performant than other lock implementations?

rocksdb_replicator/max_number_box.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // still holding mtx_.
+  // This could make a pending thread return true from wait() or wait_for() due
+  // to spurious wakeup or timeout. This is benign. Because we use the pred versions
+  // of wait() and wait_for(). Also concurrent calling notify_one() is safe.
+  std::for_each(waiters_to_notify.begin(), waiters_to_notify.end(),
+                [] (Waiter* w) { w->cv.notify_one(); });
+}
+
+bool MaxNumberBox::wait(const uint64_t num, const uint64_t timeout_ms) {
+  thread_local Waiter me;
```
shuzhang1989: add some comment on the reason of thread local?

rocksdb_replicator/max_number_box.h
newpoo: move constructor is implicilty deleted if the copy constructor is explicitly deleted.

rocksdb_replicator/max_number_box.cpp
newpoo: Will explain it a little in the comment

rocksdb_replicator/max_number_box.cpp
newpoo: Will explain it a little in the comment

PR #14, opened 2017-01-05
Improvements to C++ status server:

anshulpundir:
1.  Support HTTP GET with params.
	2. Json formatting by default. ".txt" for text files.

### DIFF
--- a/common/stats/status_server.cpp
+++ b/common/stats/status_server.cpp
@@ -23,6 +23,7 @@
 #include <stdio.h>
 #include <stddef.h>
 
+#include <folly/Uri.h>
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 #include <microhttpd.h>
@@ -81,11 +82,13 @@ void StatusServer::StartStatusServerOrDie(EndPointToOPMap op_map) {
 StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
     : port_(port), d_(nullptr), op_map_(std::move(op_map)) {
 
-  op_map_.emplace("/stats.txt", [this]() {
+  // Text
+  op_map_.emplace("/stats.txt", [this]
+      (const std::vector<std::pair<std::string, std::string>>* ) {
     std::string ret = common::Stats::get()->DumpStatsAsText();
     auto itor = op_map_.find("/rocksdb_info.txt");
     if (itor != op_map_.end()) {
-      ret += itor->second();
+      ret += itor->second(nullptr);
     }
 
     return ret;
@@ -98,12 +101,22 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
 }
 
 std::string StatusServer::GetPageContent(const std::string& end_point) {
-  auto itor = op_map_.find(end_point);
-  if (itor == op_map_.end()) {
-    return "Unsupported http path.\n";
+  // Add dummy url to allow it to be parsed by folly:Uri.
+  folly::Uri u("http://blah.blah" + end_point);
+  auto params = u.getQueryParams();
+
+  auto iter = op_map_.find(u.path().toStdString());
+  if (iter == op_map_.end()) {
+    return  "Unsupported http path.\n";
   }
 
-  return itor->second();
+  std::vector<std::pair<std::string, std::string>> args;
+  for (const auto& p : params) {
+    args.push_back(std::make_pair(p.first.toStdString(),
+                                  p.second.toStdString()));
+  }
+
+  return iter->second(&args);
 }
 
 void StatusServer::Stop() {
--- a/common/stats/status_server.h
+++ b/common/stats/status_server.h
@@ -26,6 +26,7 @@
 #include <functional>
 #include <map>
 #include <string>
+#include <vector>
 
 struct MHD_Daemon;
 
@@ -35,15 +36,40 @@ class Stats;
 
 class StatusServer {
  public:
+  /*
+   * \brief Endpoint/function name to std::function mapping.
+   *
+   * Stores the mapping of an end-point/function to the the
+   * actual std::function to be executed when requested.
+   *
+   * The std::function takes as input the list of arguments to
+   * the function in the form of a vector of [key, value] pairs
+   * (arrays can be passed in as delimited values) and returns a
+   * std::string.
+   */
   using EndPointToOPMap =
-      std::map<std::string, std::function<std::string(void)>>;
+      std::map<std::string, std::function<std::string(
+              const std::vector<std::pair<std::string, std::string>>* )>>;
 
-  // Instantiate a StatusServer.
+  /*!
+   * \brief Instantiate a StatusServer.
+   *
+   * The end-point handler in the EndPointToOPMap is an
+   * std::function that take as input the list of arguments to
+   * the function in the form of a vector of [key, value] pairs
+   * (arrays can be passed in as delimited values) and returns a
+   * std::string.
+   */
   static void StartStatusServer(EndPointToOPMap op_map = EndPointToOPMap());
   // Instantiate a StatusServer or Die.
   static void StartStatusServerOrDie(
       EndPointToOPMap op_map = EndPointToOPMap());
 
+  /*!
+   * \brief Executes the function corresponding to the given end_point.
+   * @param end_point The requested end point, including the query.
+   * @return The result of the funciton call.
+   */
   std::string GetPageContent(const std::string& end_point);
 
   // Stop the HTTP serving daemon. Not currently used.
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build
 .cproject
 .project
 .idea
+cmake-build-debug/
--- a/common/stats/tests/CMakeLists.txt
+++ b/common/stats/tests/CMakeLists.txt
@@ -5,7 +5,7 @@ file(GLOB TEST_SOURCES ./*.cpp)
 foreach(testsourcefile ${TEST_SOURCES})
   get_filename_component(testname ${testsourcefile} NAME_WE)
   add_executable(${testname} ${testsourcefile})
-  target_link_libraries(${testname} gtest stats)
+  target_link_libraries(${testname} gtest stats curl)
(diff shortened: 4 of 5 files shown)


common/stats/status_server.h
```diff
@@ -35,15 +38,32 @@ class Stats;
...
 class StatusServer {
  public:
+  /*
+   * \brief Endpoint/function name to std::function mapping.
+   *
+   * Stores the mapping of an end-point/function to the the
+   * actual std::function to be executed when requested.
+   *
+   * The std::function takes as input the list of arguments to
+   * the function in the form of a vector of [key, value] pairs
```
newpoo: mentioning that if the endpoint ends with ".txt", the server will return the content generated by op(), otherwise, the server will serialize it to json before returning it?

newpoo: Another option is to let op() decide if it should return json or text. The server only returns whatever output by op(). This seems more reasonable to me.

common/stats/status_server.cpp
```diff
@@ -98,12 +101,33 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
+    }
+  }
+
+  auto iter = op_map_.find(u.path().toStdString());
+  if (iter == op_map_.end()) {
+    return  "Unsupported http path.\n";
+  }
+
+  std::string text_ext = ".txt";
+  if (u.path().substr(u.path().length() - text_ext.length(), text_ext.length())
```
newpoo: check if u.path().length() >= text_ext.length().

We can use the single parameter version of std::string::sub_str()?

common/stats/status_server.cpp
```diff
@@ -98,12 +101,33 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
+                                    p.second.toStdString()));
+    }
+  }
+
+  auto iter = op_map_.find(u.path().toStdString());
+  if (iter == op_map_.end()) {
+    return  "Unsupported http path.\n";
+  }
+
+  std::string text_ext = ".txt";
```
newpoo: nit: const

common/stats/status_server.cpp
```diff
@@ -98,12 +101,33 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
 std::string StatusServer::GetPageContent(const std::string& end_point) {
-  auto itor = op_map_.find(end_point);
-  if (itor == op_map_.end()) {
-    return "Unsupported http path.\n";
+  // Add dummy url to allow it to be parsed by folly:Uri.
+  folly::Uri u("http://blah.blah" + end_point);
+  auto params = u.getQueryParams();
+
+  std::vector<std::pair<std::string, std::string>> args;
+  if (!params.empty()) {
```
newpoo: nit: remote this check?

common/stats/status_server.cpp
```diff
@@ -98,12 +101,33 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
+    for (auto p : params) {
+      args.push_back(std::make_pair(p.first.toStdString(),
+                                    p.second.toStdString()));
+    }
+  }
+
+  auto iter = op_map_.find(u.path().toStdString());
+  if (iter == op_map_.end()) {
+    return  "Unsupported http path.\n";
```
newpoo: do this before constructing args vector?

common/stats/status_server.h
anshulpundir: The way it is now is to return a dynamic, which is always converted to json unless it ends with .txt. The benefit of this that we avoid repeating the call to serialize to json in all handlers.

newpoo: okay, I am fine with either way.

common/stats/status_server.cpp
```diff
@@ -98,12 +101,33 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
 }
 
 std::string StatusServer::GetPageContent(const std::string& end_point) {
-  auto itor = op_map_.find(end_point);
-  if (itor == op_map_.end()) {
-    return "Unsupported http path.\n";
+  // Add dummy url to allow it to be parsed by folly:Uri.
+  folly::Uri u("http://blah.blah" + end_point);
+  auto params = u.getQueryParams();
```
newpoo: I forgot one thing.
accoding to https://github.com/facebook/folly/blob/master/folly/Uri.h#L96-L98

getQueryParams() is not completely thread safe. Can we make a first call in its constructor? This requirement looks funny tho

common/stats/status_server.cpp
```diff
@@ -98,12 +101,33 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
+
+  auto iter = op_map_.find(u.path().toStdString());
+  if (iter == op_map_.end()) {
+    return  "Unsupported http path.\n";
+  }
+
+  std::string text_ext = ".txt";
+  if (u.path().substr(u.path().length() - text_ext.length(), text_ext.length())
+      == text_ext) {
+    return iter->second(&args).asString();
```
hrjaco: What's the format of folly::dynamic::asString()? Is it a pretty print which includes nested indentation?

common/stats/status_server.h
hrjaco: I think the current way of implementation is good. Let op return folly::dynamic, and the status server handles the formatting(text or json).

common/stats/status_server.cpp
```diff
@@ -81,11 +83,12 @@ void StatusServer::StartStatusServerOrDie(EndPointToOPMap op_map) {
 StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
     : port_(port), d_(nullptr), op_map_(std::move(op_map)) {
 
-  op_map_.emplace("/stats.txt", [this]() {
+  op_map_.emplace("/stats.txt", [this]
+      (const std::vector<std::pair<std::string, std::string>>* ) {
     std::string ret = common::Stats::get()->DumpStatsAsText();
```
hrjaco: Do you want to add new API to Stats() to return folly::dynamic?

anshulpundir: Sorry, not sure I understand. Can you please clarify ?

common/stats/status_server.cpp
anshulpundir: This function can only be used to access ints and strings. For arrays, objects etc, the op will need to return a formatted string as a dynamic.

common/stats/status_server.h
```diff
@@ -35,15 +39,40 @@ class Stats;
...
+              const std::vector<std::pair<std::string, std::string>>* )>>;
+
+  /*!
+   * \brief Instantiate a StatusServer.
+   *
+   * The end-point handler in the EndPointToOPMap is an
+   * std::function that take as input the list of arguments to
+   * the function in the form of a vector of [key, value] pairs
+   * (arrays can be passed in as delimited values) and returns a
+   * std::string which is expected to be (json/text) formatted.
```
hrjaco: Maybe remove the comment of "which is expected to be (json/text) formatted". It does not have much meaning now as the status server does not care about the format of the response.

common/stats/status_server.cpp
```diff
@@ -98,12 +114,22 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
+  folly::Uri u("http://blah.blah" + end_point);
+  auto params = u.getQueryParams();
+
+  auto iter = op_map_.find(u.path().toStdString());
+  if (iter == op_map_.end()) {
+    return  "Unsupported http path.\n";
+  }
+
+  std::vector<std::pair<std::string, std::string>> args;
+  for (auto p : params) {
```
newpoo: const auto&

common/stats/status_server.cpp
```diff
@@ -81,11 +82,26 @@ void StatusServer::StartStatusServerOrDie(EndPointToOPMap op_map) {
...
     if (itor != op_map_.end()) {
-      ret += itor->second();
+      ret += itor->second(nullptr);
+    }
+
+    return ret;
+  });
+
+  // Json
+  op_map_.emplace("/stats", [this]
```
newpoo: @hrjaco Where are we going to use this end point for json formatted stats?

common/stats/stats.h
```diff
@@ -149,6 +151,9 @@ class Stats {
   // ostrich library dumps stats).
   std::string DumpStatsAsText();
 
+  // Return stats as folly::dynamic.
+  folly::dynamic DumpStatsAsDyn();
```
newpoo: If we really need json formated stats, I think we should return string here, and rename the function to DumpStatsAsJson(). and remove folly/json.h folly/dynamic.h from stats.h and status_server.h, but to put the "constructing dynamic" and "convert dynamic to json" logic in stats.cpp.

common/stats/status_server.cpp
hrjaco: I do not need json format for the stats. But may need it for other endpoint.

newpoo: Then we should remove the support for "/stats" endpoint.

common/stats/stats.h
```diff
@@ -57,6 +57,8 @@
 #pragma once
 
 #include <folly/detail/CacheLocality.h>
+#include <folly/dynamic.h>
+#include <folly/json.h>
```
newpoo: This two lines can be removed now?

common/stats/status_server.h
```diff
@@ -26,6 +26,10 @@
 #include <functional>
 #include <map>
 #include <string>
+#include <vector>
+
+#include <folly/dynamic.h>
+#include <folly/json.h>
```
newpoo: This two lines can be removed now?

common/stats/status_server.h
```diff
@@ -64,5 +93,11 @@ class StatusServer {
   MHD_Daemon* d_;
 
   EndPointToOPMap op_map_;
+
+  static std::string JsonString(folly::dynamic d) {
+    folly::json::serialization_opts opts;
+    opts.pretty_formatting = true;
+    return folly::json::serialize(d, opts);
+  }
```
newpoo: remove this function

PR #15, opened 2017-01-06
Add getDBs() method to ApplicationDBManager

shuzhang1989:
RealPin will need to iterate all Dbs to set up a index map for all shards in startup time. currently no such interface is avaiable. the method is simply returning a snapshot.


newpoo:
What index map?

newpoo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[a discussion](https://reviewable.io:443/reviews/pinterest/rocksplicator/15#-KaJIK2EZCamZg79xDI8:-KaJIK2EZCamZg79xDI9:b5os7he) (no related file):*
Since we don't need this change, can you close it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/15)*

PR #18, opened 2017-01-09
[rocksplicator] add the root pom file for the controller

newpoo:
add the root pom file for the controller


newpoo:
@fangjian601 This is currently an empty project. As we adding more content to the project, the ide will add dependencies automatically to the pom file.

PR #20, opened 2017-01-11
Add service skeleton for controller

angxu:
- Add http service prototype for controller
- Add necessary pom.xml change to build controller server


shuzhang1989:
@angxu one high level thing - how about you put this into controller/src/java/main and make a package for controller-http? we will also put different components there so it is easier to share code in future

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -0,0 +1,50 @@
...
+  @Produces(MediaType.APPLICATION_JSON)
+  public TaskBean get(@PathParam("id") String id) {
+    throw new UnsupportedOperationException("method not implemented");
+  }
+
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  public List<TaskBean> getAll(@QueryParam("state") TaskBean.State state) {
+    throw new UnsupportedOperationException("method not implemented");
+  }
```
shuzhang1989: we may also need to support querying all tasks for a given cluster

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * @author Ang Xu (angxu@pinterest.com)
+ */
+@Path("/v1/clusters/{clusterName : [a-zA-Z0-9\\-_]+}")
+public class Clusters {
+
+  @POST
```
shuzhang1989: any reason not putting it into Tasks class?

angxu: This method only abort tasks for a given cluster, thus I wanted to model it as a sub-resource of the "Clusters" resource.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
angxu: We can also model it as a sub-resource of "Clusters" resource. What do you think?

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,70 @@
...
+  @Path("/abortTask")
```
newpoo: Can you add detailed documents for each API? Like semantics and the meaning of each parameters, etc.

Since this is just a skeleton, so you may add it in subsequent PRs.

Do we have a complete list of APIs we need to implement, and the subset we want to implement for the first version?

angxu:
Review status: 0 of 13 files reviewed at latest revision, 6 unresolved discussions.

---

*[controller/controller-http/bin/run.sh, line 2 at r2](https://reviewable.io:443/reviews/pinterest/rocksplicator/20#-KaEoRFpoET4__RVMAU_-r2-2:-KaEpmOH0lgFNLc6I6TC:b-80tlqk) ([raw file](https://github.com/pinterest/rocksplicator/blob/7d274c55397e08a81d9f3f9988073f0a6379485e/controller/controller-http/bin/run.sh#L2)):*

Can you also add the copyright section to this file?

Can you add a few sample commands to describe how to run this script?
</blockquote></details>

Sure.

---

*[controller/controller-http/bin/server.yaml, line 1 at r2](https://reviewable.io:443/reviews/pinterest/rocksplicator/20#-KaEoRFpoET4__RVMAUa-r2-1:-KaEq-VvR-y_2FNUOrjo:b1dgasw) ([raw file](https://github.com/pinterest/rocksplicator/blob/7d274c55397e08a81d9f3f9988073f0a6379485e/controller/controller-http/bin/server.yaml#L1)):*

Can we put it in a config dir? because it seems all other service at Pinterest are doing it this way.

Another question is what's the preferred format for config files? Is .yaml better than properties files? I am not familiar with the best practice of the Jave world. I saw Niagara is using properties
</blockquote></details>

Sure. But I was thinking that these files are just for demonstration purpose to OS consumers and probably shouldn't be used directly within Pinterest. Internally, we can use whatever config format that is desired.

---

*[controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java, line 36 at r2](https://reviewable.io:443/reviews/pinterest/rocksplicator/20#-KaEoRFpoET4__RVMAUh-r2-36:-KaEtNL6GVrA5K5gPdhp:bsz60xo) ([raw file](https://github.com/pinterest/rocksplicator/blob/7d274c55397e08a81d9f3f9988073f0a6379485e/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java#L36)):*

Can you add detailed documents for each API? Like semantics and the meaning of each parameters, etc.

Since this is just a skeleton, so you may add it in subsequent PRs.

Do we have a complete list of APIs we need to implement, and the subset we want to implement for the first version?
</blockquote></details>

Yes. All these APIs here are listed in @shuzhang1989's design doc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/20)*

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
shuzhang1989: we can prioritize for Q1 goal: healthcheck, and being able to replace a host via HTTP api.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Retrieves all the tasks match the given {@link com.pinterest.rocksplicator.controller.bean.TaskBean.State state}.
+   *
+   * @param state   state of the task being queried.
+   * @return        a list of {@link TaskBean}s
+   */
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  public List<TaskBean> getAll(@QueryParam("state") TaskBean.State state) {
```
newpoo: remove this api, and add an optional state parameter to the clusters getAllTasks API?

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -0,0 +1,60 @@
...
+public class Tasks {
+
+  /**
+   * Retrieves a task by the task id.
+   *
+   * @param id  task id
+   * @return    {@link TaskBean} or null if task doesn't exist
+   */
+  @GET
+  @Path("/{id : [a-zA-Z0-9\\-_]+}")
```
newpoo: task id will always be a number. because it will be genereated automatically by mysql auto increament

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Hosts.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Removes a host from a given cluster.
+   *
+   * @param clusterName name of the cluster
+   * @param host        host to remove, in the format of "ip:port"
+   */
+  @DELETE
+  @Path("/{hostName : [a-zA-Z0-9\\-_]+}")
+  public void removeHost(@PathParam("clusterName") String clusterName,
```
newpoo: remove this api, it's too low level.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Hosts.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Adds a list of hosts to a given cluster to make it complete.
+   *
+   * @param clusterName name of the cluster
+   * @param hosts       a list of hosts to add, in the format of "ip:port"
+   */
+  @POST
+  @Path("/addHosts")
+  @Consumes(MediaType.APPLICATION_JSON)
+  public void addHosts(@PathParam("clusterName") String clusterName,
```
newpoo: I think we can also remove this API, it's too low level. We should provide something in clusters endpoint like expanding or shrinking, etc (we don't need to add them now).

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * @author Ang Xu (angxu@pinterest.com)
+ */
+@Path("/v1/clusters/{clusterName : [a-zA-Z0-9\\-_]+}")
+public class Clusters {
+
+  /**
+   * Retrieves all tasks belong to a given cluster by cluster name.
+   *
+   * @param clusterName name of the cluster
```
newpoo: add an optional state parameter? if it's not set, we retrive all, otherwise we retrieve the ones in the specified state. and rename the function to getTasks.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,114 @@
...
+  @Produces(MediaType.APPLICATION_JSON)
+  public List<TaskBean> getAllTasks(@PathParam("clusterName") String clusterName) {
+    throw new UnsupportedOperationException("method not implemented.");
+  }
+
+  /**
+   * Aborts the current running tasks on a given cluster.
+   *
+   * @param clusterName name of the cluster
+   * @param retry       whether to retry abort or not if the first attempt failed
```
newpoo: I don't think we need a retry parameter

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,114 @@
...
+  @POST
+  @Path("/abortTask")
+  @Produces(MediaType.APPLICATION_JSON)
+  public List<TaskBean> abortTask(@PathParam("clusterName") String clusterName,
+                                  @QueryParam("retry") boolean retry) {
+    throw new UnsupportedOperationException("method not implemented.");
+  }
+
+  /**
+   * Initializes the shard config and controller config for a given cluster.
```
newpoo: what is "controller config"?

angxu:
Review status: 0 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java, line 39 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/20#-KaEoRFpoET4__RVMAUh-r3-39:-KaFGrnE4ptGqDac1rRL:bgvs52y) ([raw file](https://github.com/pinterest/rocksplicator/blob/3cc45cad9b45c68d0c3c1eb19ac30f080b3ddd40/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java#L39)):*

add an optional state parameter? if it's not set, we retrive all, otherwise we retrieve the ones in the specified state. and rename the function to getTasks.
</blockquote></details>

will do.

---

*[controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java, line 64 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/20#-KaEoRFpoET4__RVMAUh-r3-64:-KaFFqLtptzMuGGaRmf_:b-kf36nm) ([raw file](https://github.com/pinterest/rocksplicator/blob/3cc45cad9b45c68d0c3c1eb19ac30f080b3ddd40/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java#L64)):*

what is "controller config"?
</blockquote></details>

copy-pasted from controller design doc. @shuzhang1989 could you explain it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/20)*

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Expands or shrinks the cluster to the given size.
+   *
+   * @param clusterName name of the cluster
+   * @param size        target cluster size
+   */
+  @POST
+  @Path("/resize")
```
shuzhang1989: I think the behavior will be expanding / shrinking the cluster to move shards evenly. it will be useful for us to scale cluster.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+
+  /**
+   * Recovers the config of a cluster to the last snapshot.
+   *
+   * @param clusterName name of the cluster
+   */
+  @POST
+  @Path("/recoverConfig")
```
shuzhang1989: in case something goes wrong, we need a way to rollback to last correct shard mapping. i feel this api will be useful if some tasks enter bad state and we want to rollback to last good config

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
shuzhang1989: +1 for providing a state parameter

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Hosts.java
shuzhang1989: expanding/shrinking will be done in "resize" cluster.
this is similar to "add_without_move_shards" in realpin which add back missing hosts.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
shuzhang1989: probably query tasks with states is good enough.

PR #21, opened 2017-01-13
Framework skeleton for Worker service

shuzhang1989:
No tests yet. just for high level review to ensure the architecture is good.


angxu:
Review status: 0 of 10 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 2 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLK4LJcfPLJOXXzu1_:-KaLK4LKJMDnwEIEZ6Ni:b-p8p8hp) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L2)):*
> ```Java
> /**
>  * Licensed to the Apache Software Foundation (ASF) under one
> ```

We probably should pick one copyright template first and stick with it...

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 43 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJIKU6X3zNYZtEM4s:-KaLJIKV2TgaL7orqZz2:b-erh092) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L43)):*
> ```Java
>   private TaskDispatcher() {}
> 
>   public static TaskDispatcher getInstance() {
> ```

I wouldn't make this class singleton unless you don't expect it to take any parameters.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 61 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJZU3LEKjZb9m2KFQ:-KaLJZU3LEKjZb9m2KFR:b-8uewin) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L61)):*
> ```Java
>    */
>   public boolean start(long dispatcherPollInterval) {
>     synchronized (TaskDispatcher.class) {
> ```

Don't synchronize on static object. You can just put synchronized keyword in the method signature.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 71 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLQ4gD42QSWmDBVZUk:-KaLQ4gEMfUtMrc2gc_h:b-ovmf9c) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L71)):*
> ```Java
>         }
>       };
>       scheduler.scheduleAtFixedRate(dispatcher, 0, dispatcherPollInterval, TimeUnit.SECONDS);
> ```

One problem with #scheduleAtFixedRate is that it might start the next runnable while the previous one hasn't finished. An alternative is to schedule tasks in a chained fashion, meaning that every runnable will resubmit itself to the scheduler after it finished running.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 27 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLONAfYo5pnzQEpFcJ:-KaLONAgX2LqkxGgZk3-:b-4ezuek) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L27)):*
> ```Java
> import org.slf4j.LoggerFactory;
> 
> public class WorkerConfig {
> ```

maybe consider using http://www.dropwizard.io/1.0.5/docs/manual/core.html#configuration?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 40 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLoOtTcM2inDrRrGi:-KaLLoOtTcM2inDrRrGj:b-e86jsi) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L40)):*
> ```Java
>  *
>  */
> public class WorkerPool extends ThreadPoolExecutor {
> ```

It's better to have WorkerPool implements Executor/ExecutorService interface and have an Executor as its member. That way you can easily experiment with different ExecutorService impl and not mess up the inheritance hierarchy.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 80 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLgclIgj83oGvlTh7:-KaLLgclIgj83oGvlTh8:b-rqxvo7) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L80)):*
> ```Java
>   }
> 
>   public static WorkerPool getInstance() {
> ```

ditto.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java, line 32 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLOgXgzEQ7sW24unko:-KaLOgXhifeoLtt2kq2O:b5q3mbm) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java#L32)):*
> ```Java
>  * @author Shu Zhang (shu@pinterest.com)
>  */
> public abstract class Task implements Callable<TaskExecutionResponse> {
> ```

Use generic? For example, Task<T> implements Callable<TaskExecutionResponse<T>> where T is the result type. Btw, do we expect async task? If so, I'd prefer CompletableFuture over Callable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: 0 of 10 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 2 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLK4LJcfPLJOXXzu1_:-KaO-kaSrF8RL9yLRAG7:b-snhplx) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L2)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

We probably should pick one copyright template first and stick with it...
</blockquote></details>

sure, i will copy yours then.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 43 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJIKU6X3zNYZtEM4s:-KaO0GTD1xkxsgjlUF6m:b-g2i6ky) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L43)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

I wouldn't make this class singleton unless you don't expect it to take any parameters.
</blockquote></details>

i thought singleton can make the intention to have only one dispatcher globally clearer. what do you think?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 61 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJZU3LEKjZb9m2KFQ:-KaO-uWFpqfznsLqKvBJ:bb7oryy) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L61)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

Don't synchronize on static object. You can just put synchronized keyword in the method signature.
</blockquote></details>

sure

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 71 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLQ4gD42QSWmDBVZUk:-KaO0bRQZkrpcYB89W9T:b-efk3bg) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L71)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

One problem with #scheduleAtFixedRate is that it might start the next runnable while the previous one hasn't finished. An alternative is to schedule tasks in a chained fashion, meaning that every runnable will resubmit itself to the scheduler after it finished running.
</blockquote></details>

good point. it's a better approach

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 27 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLONAfYo5pnzQEpFcJ:-KaO13gc1IsJ2F8DfJVu:btjnwxl) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L27)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

maybe consider using http://www.dropwizard.io/1.0.5/docs/manual/core.html#configuration?
</blockquote></details>

the worker will not be a http service. do you think depending on dropwizard just for a config is too much?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 40 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLoOtTcM2inDrRrGi:-KaO1UBUG6h4uIhXoWRL:b4vxfi6) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L40)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

It's better to have WorkerPool implements Executor/ExecutorService interface and have an Executor as its member. That way you can easily experiment with different ExecutorService impl and not mess up the inheritance hierarchy.
</blockquote></details>

sure. 

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 80 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLgclIgj83oGvlTh7:-KaO1i335c8F4Gxo-gNT:b-ic4fav) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L80)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

ditto.
</blockquote></details>

same reason as dispatcher. expect only one worker pool globally so made it a singleton...

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java, line 32 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLOgXgzEQ7sW24unko:-KaO30-7t9L1uafZs1UO:b-kdkj78) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java#L32)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

Use generic? For example, Task<T> implements Callable<TaskExecutionResponse<T>> where T is the result type. Btw, do we expect async task? If so, I'd prefer CompletableFuture over Callable.
</blockquote></details>

sure, i think most of the tasks will be executed in thread sequentially. but will change to completablefuture for flexibility.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: 0 of 10 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 43 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJIKU6X3zNYZtEM4s:-KaO5I07OEo0FDAkDXG2:bux3vib) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L43)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

i thought singleton can make the intention to have only one dispatcher globally clearer. what do you think?
</blockquote></details>

i think the dispatcherPollInterval can be got from Configuration directly, so there should be no intervals parameters need to be passed in

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

newpoo:
I think we should implement something similar to Leader/Follower modeled worker pool. The current model can have tasks in the pending queue of the executor, if all worker threads are running other long running tasks. This will make the system incorrectly thinks the worker is dead, and re-assign the pending tasks to other workers.

---

Review status: 0 of 10 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/pom.xml, line 6 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO5-1FlufbWh0rYsq6:-KaO5-1FlufbWh0rYsq7:bg1mi0n) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/pom.xml#L6)):*
> ```XML
>          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
>     <parent>
>         <artifactId>controller</artifactId>
> ```

use a different artifactid than the API server?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 43 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJIKU6X3zNYZtEM4s:-KaO7tWTlTILCQnYECyB:b-imxxvs) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L43)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

i think the dispatcherPollInterval can be got from Configuration directly, so there should be no intervals parameters need to be passed in
</blockquote></details>

+1 for avoiding singleton if possible

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 44 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO5OFstRFJu0ki69Ar:-KaO5OFstRFJu0ki69As:bpb17kc) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L44)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>
>     if (instance == null) {
>       synchronized (TaskDispatcher.class) {
>         if (instance == null) {
>           instance = new TaskDispatcher();
>           return instance;
>         }
>       }
</details>
Do we have to use such complicated locking mechanism to implement a simple thread-safe singleton?

Can we simply initialize it in static initialization block or something similar, because lazy initialization is not important for us.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 27 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLONAfYo5pnzQEpFcJ:-KaO5sChmkvC5ev1zlxd:b-77pq3l) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L27)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

the worker will not be a http service. do you think depending on dropwizard just for a config is too much?
</blockquote></details>

+1 for minimize dependencies

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/TaskExecutionResponse.java, line 48 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO6PF3-X0ny8UR_P_1:-KaO6PF3-X0ny8UR_P_2:b35razy) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/TaskExecutionResponse.java#L48)):*
> ```Java
> 
>   @Override
>   public boolean equals(Object o) {
> ```

why do we need to override equals() and hashCode() ? I don't see you use it in anything like map, set, or call equals() on TaskExecutionResponse

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-:-KaO6eUlpkcpeI1iHnKf:b-89rmx3)*

angxu:
Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 27 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLONAfYo5pnzQEpFcJ:-KaOCipTce60QM_zvsPi:b-j7e8if) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L27)):*

+1 for minimize dependencies
</blockquote></details>

I see. Thought they were gonna co-exist in the same jvm.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java, line 32 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLOgXgzEQ7sW24unko:-KaODuRCxm2aWNGy7XT3:b-mckxm4) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java#L32)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

sure, i think most of the tasks will be executed in thread sequentially. but will change to completablefuture for flexibility.
</blockquote></details>

I meant ```Task<T> implements Callable<TaskExecutionResponse<T>>```...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java, line 32 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLOgXgzEQ7sW24unko:-KaOEsFCL5rCGcIwDHxr:b-3aulbu) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java#L32)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

I meant ```Task<T> implements Callable<TaskExecutionResponse<T>>```...
</blockquote></details>

i have no problem with that.. i am thinking about CompletableFuture and whether it will be useful

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

angxu:
If the concern is on bounded thread pool and possible deadlock situation, we can use ForkJoinPool and run long running tasks via its #managedBlock method. That way we can ensure there are always enough worker threads to pick up the new tasks(ie, the task to kill previous long running job).

---

Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-:-KaOJqLfQcEU1pjboxES:b-bo4f69)*

shuzhang1989:
based on the offline discussion, i think we can use a semaphore to control the workers not claiming more tasks than the pool size. using semaphore is more elegant than to check the map size / active thread number in dispatcher. the worker can increment the semaphore in AfterExecution() hook and the dispatcher can grab the semaphore. this can minimize the change and still reuse executor service's utils instead of writing again the code. what do you guys think?

---

Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-:-KaOrZSR7lu-QbjQgMOp:b-pehb2n)*

shuzhang1989:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/pom.xml, line 6 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO5-1FlufbWh0rYsq6:-KaOvR9gIqC5osyxSlop:by1rbvi) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/pom.xml#L6)):*

use a different artifactid than the API server?
</blockquote></details>

this is parent artifact id. 

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/TaskExecutionResponse.java, line 48 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO6PF3-X0ny8UR_P_1:-KaOwFpIYyABfbYJmDun:b-fm0mjp) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/TaskExecutionResponse.java#L48)):*

why do we need to override equals() and hashCode() ? I don't see you use it in anything like map, set, or call equals() on TaskExecutionResponse
</blockquote></details>

it is auto generated by intellij. i can remove them

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 43 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLJIKU6X3zNYZtEM4s:-KaPABUOdZATg3HwU_2o:bbtqvys) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L43)):*

+1 for avoiding singleton if possible
</blockquote></details>

will change to not be a singleton.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 44 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO5OFstRFJu0ki69Ar:-KaPAH1EpFf-WALQw6wu:bdjbqdo) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L44)):*

>     if (instance == null) {
>       synchronized (TaskDispatcher.class) {
>         if (instance == null) {
>           instance = new TaskDispatcher();
>           return instance;
>         }
>       }
Do we have to use such complicated locking mechanism to implement a simple thread-safe singleton?

Can we simply initialize it in static initialization block or something similar, because lazy initialization is not important for us.
</blockquote></details>

it's not a singleton anymore, will delete this block.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 40 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLoOtTcM2inDrRrGi:-KaPKQAyzqpKac0qjPmx:b-7jj1hu) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L40)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

sure. 
</blockquote></details>

i may need to stick to ThreadPoolExecutor because it provides afterExeucte() for worker to release the Semaphore. i think this need to be transparent to users who write the task.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

newpoo:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[controller/controller-worker/pom.xml, line 6 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaO5-1FlufbWh0rYsq6:-KaPN_b4x3Vt7oypjSxC:byaivch) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/pom.xml#L6)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

this is parent artifact id. 
</blockquote></details>

Sorry, miss read it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

angxu:
Review status: 3 of 9 files reviewed at latest revision, 11 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 40 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLoOtTcM2inDrRrGi:-KahnfZrfdJ1VGETbbiL:b-f6ys71) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L40)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

i may need to stick to ThreadPoolExecutor because it provides afterExeucte() for worker to release the Semaphore. i think this need to be transparent to users who write the task.
</blockquote></details>

A common practice in java to add pre/post execution hook is to wrap the input runnable. It's also easier to share per task/execution context between the two hooks. For example:

```
void execute(Runnable run) {
  executor.execute(() -> {
    Context ctx = new Context();
    preRun(ctx);
    try {
      run.run();
    } finally {
      postRun(ctx);
    }
  });
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: 3 of 9 files reviewed at latest revision, 11 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 40 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaLLoOtTcM2inDrRrGi:-Kai-OalSMwBV2HRjL23:b-mbalc3) ([raw file](https://github.com/pinterest/rocksplicator/blob/c2cbf9700af41ed764b3be2f6e86f5af0a3e811d/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L40)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

A common practice in java to add pre/post execution hook is to wrap the input runnable. It's also easier to share per task/execution context between the two hooks. For example:

```
void execute(Runnable run) {
  executor.execute(() -> {
    Context ctx = new Context();
    preRun(ctx);
    try {
      run.run();
    } finally {
      postRun(ctx);
    }
  });
}
```
</blockquote></details>

Cool, thanks for the idea :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

angxu:
Reviewed 5 of 7 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 37 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiLkKcR4tcj475qjLX:-KaiLkKcR4tcj475qjLY:bfnwup7) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L37)):*
> ```Java
>   private boolean isRunning = false;
>   private final ExecutorService scheduler = Executors.newSingleThreadExecutor();
>   private long dispatcherPollInterval;
> ```

nitpick: Make it final if you don't need to modify its value. Same below.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 63 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiLIY2p45_pqsGaPdu:-KaiLIY2p45_pqsGaPdv:bpiy4eh) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L63)):*
> ```Java
>           idleWorkersSemaphore.acquire();
>           // TODO: Add task dispatch logic to assign tasks to worker pool
>           Thread.sleep(dispatcherPollInterval * 1000);
> ```

nitpick: ```scheduler.schedule(this, dispatcherPollInterval, TimeUnit.MILLISECONDS);```

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 61 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiMQ-cVULkk3Q37wD_:-KaiMQ-cVULkk3Q37wDa:b3mzwzx) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L61)):*
> ```Java
>     executorService.submit(() -> {
>       task.call();
>       idleWorkersSemaphore.release();
> ```

use try-finally 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 61 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiMQ-cVULkk3Q37wD_:-KaiQPKHs62l0qaomWGE:b-yy34r) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L61)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

use try-finally 
</blockquote></details>

task.call() has try-catch-finally. do we need it here too?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

angxu:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 61 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiMQ-cVULkk3Q37wD_:-KaiRdfIF24F0r3U85XZ:b-ipqi8c) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L61)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

task.call() has try-catch-finally. do we need it here too?
</blockquote></details>

I see. It's safer to do try-finally because the sub-class might still throw plus that you didn't catch non-execption throwables in Task, but I'm Okay either way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

angxu:
:lgtm:

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-:-KaiU0p-ezH8ZleQfyMS:bnfp4nl)*

newpoo:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 39 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiW-U3BAKRezyUbsHM:-KaiW-U3BAKRezyUbsHN:b-pfqf08) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L39)):*
> ```Java
>   private long dispatcherPollInterval;
>   private WorkerPool workerPool;
>   private volatile Semaphore idleWorkersSemaphore;
> ```

don't have to be volatile?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 44 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiUTqFpSW9usFlWf9q:-KaiUTqFpSW9usFlWf9r:b-ee7ztu) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L44)):*
> ```Java
>   public TaskDispatcher(long dispatcherPollInterval, Semaphore idleWorkersSemaphore,
>                         WorkerPool workerPool) {
>     this.dispatcherPollInterval = dispatcherPollInterval;
> ```

nit: append Seconds to the names?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 78 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiVQNDtLB-FPJTssjz:-KaiVQNDtLB-FPJTssk-:bz5toga) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L78)):*
> ```Java
>    * Stop to dispatch tasks.
>    */
>   public synchronized boolean stop() {
> ```

Do we allow a stopped dispatcher to start again? If yes, we need to create scheduler in start(), otherwise, we need to have a flag to indicate termination.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 83 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiVKXqqyTgFq-of8o_:-KaiVKXqqyTgFq-of8oa:b-61rc6p) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L83)):*
> ```Java
>       return false;
>     }
>     scheduler.shutdown();
> ```

using awaitTermination() ?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 32 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiQio9mYUM6EGu0Qdd:-KaiQio9mYUM6EGu0Qde:b2d259h) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L32)):*
> ```Java
>   private static PropertiesConfiguration configuration;
> 
>   public static void initialize() throws ConfigurationException {
> ```

replace initialize() with static initialization block? So that client doesn't need to explicitly call initialize().

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 40 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiQYyKVQDO5sSXpVJv:-KaiQYyKVQDO5sSXpVJw:bf90h31) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L40)):*
> ```Java
> 
>   public static long getDispatcherPollInterval() {
>     return configuration == null? 60 : configuration.getLong(DISPATCHER_POLL_INTERVAL_KEY);
> ```

nit: one space between null and ?, same comment to getWorkerPoolSize().

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 41 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicScVLEZdKh1_DDyM:-KaicScVLEZdKh1_DDyN:b-w8oqci) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L41)):*
>  private volatile Semaphore idleWorkersSemaphore;
>   private volatile ConcurrentHashMap<String, FutureTask> runningTasks;
>   
why volatile?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 43 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicYMEfnqiIiyvsEew:-KaicYMEfnqiIiyvsEex:b5rtr5) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L43)):*
> ```Java
>   private volatile ExecutorService executorService;
> 
>   public WorkerPool(ExecutorService executorService, Semaphore idleWorkersSemaphore) {
> ```

new the hash map?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 54 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiaRwAsjgbkP5PFDLJ:-KaiaRwBuelkEbJbG0GX:bm54582) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L54)):*
> ```Java
>    */
>   public void assignTask(Task task) throws Exception {
>     if (runningTasks.containsKey(task.getCluster())) {
> ```

If this is true, that means something really happened. can you log more info, at least log cluster name.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 61 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiMQ-cVULkk3Q37wD_:-Kai_NjRAaENrQ8U9wvV:b-mn6oov) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L61)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

I see. It's safer to do try-finally because the sub-class might still throw plus that you didn't catch non-execption throwables in Task, but I'm Okay either way.
</blockquote></details>

+1 for using try-finally, task.call() will call preprocess() also, which is designed to be implemented by concreate task authors.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 72 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kai_dmsJQP3F6MJMzJU:-Kai_dmsJQP3F6MJMzJV:bpldtdz) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L72)):*
> ```Java
>   public boolean abortTask(String cluster) throws Exception {
>     FutureTask runningTask = runningTasks.get(cluster);
>     if (runningTask.isCancelled()) {
> ```

It's not that useful to check it, because it may be cancelled after this check but before your cancel().

We can just cancel() it directly, and check the return value of cancel().

Another thing is we need to check if runningTask is null

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 84 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaia5ucFvleRjIltvJE:-Kaia5udDVJqfk0_GiQQ:br319sd) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L84)):*
> ```Java
>    * @return
>    */
>   public boolean hasRunningTask(String cluster) {
> ```

How are we going to use it?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java, line 32 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiOxqAf-hFvhWTAauz:-KaiOxqAf-hFvhWTAav-:b-vw07zk) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java#L32)):*
> ```Java
>  * (2) A pool of worker threads executing the tasks
>  * To make sure:
>  * (1) every worker thread can get task from dispatcher immediately
> ```

Item (1) is not accurate?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java, line 56 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaib_ZDruXnBAZmWRXw:-Kaib_ZDruXnBAZmWRXx:b-z894ct) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java#L56)):*
> ```Java
>       TaskDispatcher dispatcher = new TaskDispatcher(
>           WorkerConfig.getDispatcherPollInterval(), idleWorkersSemaphore, workerPool);
>       dispatcher.start();
> ```

The main thread will exit after calling start()?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java, line 82 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiboN6qg_DsBtR0uZe:-KaiboN6qg_DsBtR0uZf:b-v8ef36) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java#L82)):*
> ```Java
> 
>   @Override
>   public T call() {
> ```

Do we need to return T?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
@newpoo i merged it just now.. your comments will be refected in next diff with unit tests

---

Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 39 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiW-U3BAKRezyUbsHM:-KaieKNq2DWmR5BfVSxS:b-4cq495) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L39)):*

don't have to be volatile?
</blockquote></details>

it's shared by TaskDispatcher and WorkerPool. 

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 44 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiUTqFpSW9usFlWf9q:-KaienPqBt6bodwHb63h:bb7oryy) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L44)):*

nit: append Seconds to the names?
</blockquote></details>

sure

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 78 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiVQNDtLB-FPJTssjz:-KaieocYb8Ye2GgbqyTe:b-1pwsto) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L78)):*

Do we allow a stopped dispatcher to start again? If yes, we need to create scheduler in start(), otherwise, we need to have a flag to indicate termination.
</blockquote></details>

will move scheduler construction to start(). 

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 83 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiVKXqqyTgFq-of8o_:-Kaieuw4WgeDZ_NQQ0i9:b-vqsjng) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L83)):*

using awaitTermination() ?
</blockquote></details>

the graceful shutdown is a TODO. i will do it in next diff

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 32 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiQio9mYUM6EGu0Qdd:-KaifDqlCiH2XHgbjklW:bb7oryy) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L32)):*

replace initialize() with static initialization block? So that client doesn't need to explicitly call initialize().
</blockquote></details>

sure

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 40 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiQYyKVQDO5sSXpVJv:-KaifFs4ORQA-ehBSoCt:bb7oryy) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L40)):*

nit: one space between null and ?, same comment to getWorkerPoolSize().
</blockquote></details>

sure

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 41 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicScVLEZdKh1_DDyM:-KaifHHoFcm63PoTba6f:b-jxhuu0) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L41)):*

>  private volatile Semaphore idleWorkersSemaphore;
>   private volatile ConcurrentHashMap<String, FutureTask> runningTasks;
>   
why volatile?
</blockquote></details>

don't need volatile on runningTasks.
volatile is still needed in idleWorkerSemaphore because different worker threads will access and release it when task finishes

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 43 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicYMEfnqiIiyvsEew:-KaifTm-wa2ucRfppagj:begq0be) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L43)):*

new the hash map?
</blockquote></details>

Good catch. the bad side of not having unit tests

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 54 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiaRwAsjgbkP5PFDLJ:-KaifqOhL5q380HbYU8C:bb7oryy) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L54)):*

If this is true, that means something really happened. can you log more info, at least log cluster name.
</blockquote></details>

sure

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 61 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiMQ-cVULkk3Q37wD_:-Kaift9PktZB9vKGcJ78:bfgsl4c) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L61)):*

+1 for using try-finally, task.call() will call preprocess() also, which is designed to be implemented by concreate task authors.
</blockquote></details>

added try-finally

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 72 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kai_dmsJQP3F6MJMzJU:-Kaifv_seVvb5V6FFP9q:bb7oryy) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L72)):*

It's not that useful to check it, because it may be cancelled after this check but before your cancel().

We can just cancel() it directly, and check the return value of cancel().

Another thing is we need to check if runningTask is null
</blockquote></details>

sure

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 84 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaia5ucFvleRjIltvJE:-KaigI8c_bLM9rleE_EP:b-fn74hn) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L84)):*

How are we going to use it?
</blockquote></details>

plan to use if for unit tests. 

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java, line 32 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiOxqAf-hFvhWTAauz:-Kaih0D18mBMJoyxzAo8:b2nhysk) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java#L32)):*

Item (1) is not accurate?
</blockquote></details>

ack

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java, line 56 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaib_ZDruXnBAZmWRXw:-KaigstUzb_Y8yh_K4-q:b-afp4jz) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java#L56)):*

The main thread will exit after calling start()?
</blockquote></details>

need to block the main thread. will change.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java, line 82 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiboN6qg_DsBtR0uZe:-Kaig43IzeRTZDYdDTci:b-i1uuhu) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/Task.java#L82)):*

Do we need to return T?
</blockquote></details>

T is useful for doing unit tests (to check results)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-:-KaieB2178XZ4_3ONG63:b-srezt7)*

newpoo:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java, line 39 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiW-U3BAKRezyUbsHM:-Kaj0afLrsUQ1UyZVvPF:byrptru) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/TaskDispatcher.java#L39)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

it's shared by TaskDispatcher and WorkerPool. 
</blockquote></details>

We don't need to make it volatile even the object it refers to is shared, right?

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 41 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicScVLEZdKh1_DDyM:-Kaj0qr3P68Jy1eMjMmZ:bgabkgw) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L41)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

don't need volatile on runningTasks.
volatile is still needed in idleWorkerSemaphore because different worker threads will access and release it when task finishes
</blockquote></details>

I think it's fine. We can safely concurrently access ExecutorService and Semaphore without volatile. Actually volatile doesn't do much for us here. It only guarantees memory visibility of the variable itself.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 84 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaia5ucFvleRjIltvJE:-Kaj1i49i2cwqXiTF8Fp:br77gwo) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L84)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

plan to use if for unit tests. 
</blockquote></details>

Does JUnit allow us to access private member of the class being tested? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

angxu:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 41 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicScVLEZdKh1_DDyM:-Kaj56SEBZjMnCYOWDvW:b-ntbe1j) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L41)):*

I think it's fine. We can safely concurrently access ExecutorService and Semaphore without volatile. Actually volatile doesn't do much for us here. It only guarantees memory visibility of the variable itself.
</blockquote></details>

Make it final so that it can be set by constructor only. It also guarantees memory visibility in multi-threading scenario. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

shuzhang1989:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 32 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiQio9mYUM6EGu0Qdd:-Kaj8QINkA0eVQu9zM9S:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L32)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

sure
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java, line 40 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiQYyKVQDO5sSXpVJv:-KakOnEUEGodIzl_eVIE:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java#L40)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

sure
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 41 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicScVLEZdKh1_DDyM:-KajBgxh6vs-Mmw3nteF:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L41)):*
<details><summary><i>Previously, angxu wrote…</i></summary><blockquote>

Make it final so that it can be set by constructor only. It also guarantees memory visibility in multi-threading scenario. 
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 43 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaicYMEfnqiIiyvsEew:-KakOvY7Nj51PulWlgtM:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L43)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

Good catch. the bad side of not having unit tests
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 54 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiaRwAsjgbkP5PFDLJ:-KakPJsPihjBY0_vVHw8:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L54)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

sure
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 61 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-KaiMQ-cVULkk3Q37wD_:-KakPKDJ3DJYZD1YmTHW:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L61)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

added try-finally
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 72 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kai_dmsJQP3F6MJMzJU:-KakRjT_sTyBpCMz9jPS:b-896fix) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L72)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

sure
</blockquote></details>

Done.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 84 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaia5ucFvleRjIltvJE:-Kaj2NrmrcZZTdpqGstl:bsz5dgz) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L84)):*

Does JUnit allow us to access private member of the class being tested? 
</blockquote></details>

i think @VisibleForTesting will help. i can remove this if we dont have concrete use case.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java, line 56 at r3](https://reviewable.io:443/reviews/pinterest/rocksplicator/21#-Kaib_ZDruXnBAZmWRXw:-KamyV1vt58jbGIfBmO2:b-20og6k) ([raw file](https://github.com/pinterest/rocksplicator/blob/7979a6ff80812c78c23bbca88e1b78cefabd6524/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java#L56)):*
<details><summary><i>Previously, shuzhang1989 (Shu Zhang) wrote…</i></summary><blockquote>

need to block the main thread. will change.
</blockquote></details>

Just did a check, the main process will not exit as long as dispatcher is running. The non-daemon thread will prevent process from exisitng so it should be safe here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/21)*

PR #27, opened 2017-02-28
Add GetMetadata() to s3util to get s3 metadata information

shuzhang1989:
to be used in scorpion to get md5 of objects

### DIFF
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -134,7 +134,7 @@ GetObjectMetadataResponse S3Util::getObjectMetadata(const string &key) {
   map<string, string> metadata;
   if (headObjectResult.IsSuccess()) {
     metadata = headObjectResult.GetResult().GetMetadata();
-    return GetObjectMetadataResponse(metadata, "");
+    return GetObjectMetadataResponse(std::move(metadata), "");
   } else {
     return GetObjectMetadataResponse(
             metadata, headObjectResult.GetError().GetMessage());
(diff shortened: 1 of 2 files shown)


common/s3util.cpp
```diff
@@ -124,6 +126,21 @@ GetObjectsResponse S3Util::getObjects(
...
 
+GetObjectMetadataResponse S3Util::getObjectMetadata(const string &key) {
+  HeadObjectRequest headObjectRequest;
+  headObjectRequest.SetBucket(bucket_);
+  headObjectRequest.SetKey(key);
+  auto headObjectResult = s3Client.HeadObject(headObjectRequest);
+  map<string, string> metadata;
+  if (headObjectResult.IsSuccess()) {
+    metadata = headObjectResult.GetResult().GetMetadata();
+    return GetObjectMetadataResponse(metadata, "");
```
chengren311: NIT: should we return something indicating the request is successful instead of an empty string?

shuzhang1989: No, i use str.empty() to check if its a good response or not.

jinruh: nit: std::move(metadata)

newpoo:
please add std::move() as suggested by jinru

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/27#-:-Ke45HEJaxeFXCbuErmm:b96r7w9)*

PR #30, opened 2017-03-10
S3Util returns const ref for members

shuzhang1989:


shuzhang1989:
@newpoo oops, missed your comments..
will do and change audit all usage of this response later

PR #31, opened 2017-03-13
Add application db write

jinruh:
if replicated_db_ does is nullptr. write to db_


rocksdb_admin/application_db.cpp
```diff
@@ -94,7 +94,11 @@ rocksdb::Status ApplicationDB::Write(const rocksdb::WriteOptions& options,
   common::Stats::get()->Incr(kRocksdbWrite);
   common::Stats::get()->Incr(kRocksdbWriteBytes, write_batch->GetDataSize());
   common::Timer timer(kRocksdbWriteMs);
-  return replicated_db_->Write(options, write_batch);
```
newpoo: Can you add a comment saying that ApplicationDBManager can be used to manage DB lifecycles only (no replication involved). In this use case, ApplicationDB has no upstream and its role is set to Slave. Thus its replicated_db_ is set to nullptr.

PR #32, opened 2017-03-14
Add application db write

jinruh:
ApplicationDBManager can be use to manage rocksdb instance lifecycle without replication. In this case, ApplicationDB has the replicator::DBRole as SLAVE and no upstream_addr. Thus the replicated_db_ is nullptr, and we'll write to the local db_


jinruh:
accidentally deleted the branch in previous PR

PR #33, opened 2017-03-17
Refactor TaskBase to support type-safe task construction

angxu:
- Add two new constructs to TaskBase:

  1. `Parameter` class which holds task parameters passed in during construction time. It'll be stored as taskBody in backend.
  2. `Context` class which stores immutable bookkeeping information managed by Controller itself.

- Refactor TaskFactory to use pure reflection.


controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskQueue.java
```diff
@@ -174,25 +176,25 @@ public boolean keepTaskAlive(final long id) {
    * @param state peek tasks in this state only
    * @return the list of tasks found
    */
-  public List<Task> peekTasks(final String clusterName,
-                              final Integer state) {
-    return new ArrayList<Task>();
+  default List<TaskInternal> peekTasks(final String clusterName,
+                                      final Integer state) {
```
newpoo: nit: one more space to fix the indention.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -67,8 +68,8 @@ public Task get(@PathParam("id") Long id) {
    */
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  public List<Task> findTasks(@QueryParam("clusterName") Optional<String> clusterName,
-                              @QueryParam("state") Optional<TaskState> state) {
+  public List<TaskInternal> findTasks(@QueryParam("clusterName") Optional<String> clusterName,
```
newpoo: any specific reason why we still let get() return Task, while findTasks() returns TaskInternal

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskQueue.java
```diff
@@ -25,30 +25,28 @@
 /**
  * @author Bo Liu (bol@pinterest.com)
  */
-public final class TaskQueue {
```
newpoo: Why do we make it an interface? For test only?

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/Cancellable.java
```diff
@@ -0,0 +1,24 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.pinterest.rocksplicator.controller;
+
+/**
+ * @author Ang Xu (angxu@pinterest.com)
+ */
+public interface Cancellable {
```
newpoo: remove this file, since it's not used any where

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java
```diff
@@ -37,40 +42,49 @@
...
+    this.taskQueue = taskQueue;
   }
 
   /**
    * Assign a task to a worker thread.
    * @param task the task to execute
    * @throws Exception if there is a running task for the cluster.
    */
-  public boolean assignTask(TaskBase task) {
+  public synchronized boolean assignTask(TaskInternal task) throws Exception {
```
newpoo: Why we didn't need synchronization previously, but need it now?

angxu:
Because I'd like to change the behavior of TaskQueue in certain cases, e.g. DelayAckTaskQueue.

---

Review status: 0 of 26 files reviewed at latest revision, 6 unresolved discussions.

---

*[controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Task.java, line 21 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/33#-KfTfPSHM3SGO3BBfm_E-r1-21:-KfiGeIa0-izFzLw9pLW:bb7sa0c) ([raw file](https://github.com/pinterest/rocksplicator/blob/df1673651922845ca3b680fa66081c866bd72c31/controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Task.java#L21)):*

"import java.sql.Timestamp;" can be removed now?

Github doesn't allow me to comment on something you didn't change ...
</blockquote></details>

yes.

---

*[controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskInternal.java, line 35 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/33#-KfTfPSHM3SGO3BBfm_F-r1-35:-KfiHZ-OkPEe37QBAL7F:b-dgn1oh) ([raw file](https://github.com/pinterest/rocksplicator/blob/df1673651922845ca3b680fa66081c866bd72c31/controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskInternal.java#L35)):*

Do we really need it?
</blockquote></details>

no, we can get rid of it.

---

*[controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java, line 71 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/33#-KfTfPSHM3SGO3BBfm_K-r1-71:-KfiHSv01QRYlHLC6NnP:b-x97ol3) ([raw file](https://github.com/pinterest/rocksplicator/blob/df1673651922845ca3b680fa66081c866bd72c31/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java#L71)):*

any specific reason why we still let get() return Task, while findTasks() returns TaskInternal
</blockquote></details>

should have used TaskInternal. thanks for the catch.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/Cancellable.java, line 22 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/33#-KfTfPSHM3SGO3BBfm_L-r1-22:-KfiGoC0ZS5HFas3Sf5J:b57e11a) ([raw file](https://github.com/pinterest/rocksplicator/blob/df1673651922845ca3b680fa66081c866bd72c31/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/Cancellable.java#L22)):*

remove this file, since it's not used any where
</blockquote></details>

Oops.. good catch.

---

*[controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java, line 63 at r1](https://reviewable.io:443/reviews/pinterest/rocksplicator/33#-KfTfPSHM3SGO3BBfm_N-r1-63:-KfiGstFAWRbNkZcrbyL:b-usxkp8) ([raw file](https://github.com/pinterest/rocksplicator/blob/df1673651922845ca3b680fa66081c866bd72c31/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java#L63)):*

Why we didn't need synchronization previously, but need it now?
</blockquote></details>

Previously we use atomic putIfAbsent, but it contains a race condition because FutureTask put in map may not be submitted to executor and thus its #cancel() method may be a noop. Using synchronized key word fixes this problem

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pinterest/rocksplicator/33#-:-KfiG0r43TzksyoD8nr0:b-gxnbq7)*

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Task.java
```diff
@@ -16,21 +16,12 @@
...
 package com.pinterest.rocksplicator.controller;
 
-import java.sql.Timestamp;
 
 /**
  * @author Bo Liu (bol@pinterest.com)
  */
-final public class Task {
-  public long id;
+public class Task {
```
shuzhang1989: can we add a comment on the usage of the class?

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskInternal.java
```diff
@@ -0,0 +1,41 @@
...
+package com.pinterest.rocksplicator.controller;
+
+import java.sql.Timestamp;
+
+/**
+ * Internal representation of a Task
+ *
+ * @author Bo Liu (bol@pinterest.com)
+ */
+public class TaskInternal extends Task {
```
shuzhang1989: add a comment where is class is used?

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
shuzhang1989: if task internal will be exposed by http, maybe give it a better name because it will be seen by end user

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java
shuzhang1989: we only have one dispatcher thread doing the assignment. do we need synchronizing here?

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java
```diff
@@ -79,8 +93,8 @@ public boolean assignTask(TaskBase task) {
    * @param cluster the name of the cluster
    * @return
    */
-  public boolean abortTask(String cluster) throws Exception {
-    FutureTask runningTask = runningTasks.get(cluster);
+  public synchronized boolean abortTask(String cluster) throws Exception {
```
shuzhang1989: do we need it? the map is a concurrent map

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java
angxu: There may be different threads calling #abortTask(), thus it's safer to use synchronized.

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerPool.java
angxu: There are two critical sections we synchronize on:
```
Future f = executorService.submit(...);
runningTasks.put(clusterName, f);
```
and
```
Future f = runningTasks.get(clusterName);
f.cancel(...);
```
We don't want two threads interleaving these two critical section.

shuzhang1989: okay, making sense

PR #38, opened 2017-03-27
Implement health check task

angxu:
Implement health check task

### DIFF
--- a/controller/controller-common/pom.xml
+++ b/controller/controller-common/pom.xml
@@ -20,34 +20,28 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.5.0</version>
-        </dependency>
-        <dependency>
-            <groupId>com.twitter</groupId>
-            <artifactId>scrooge-core_2.9.2</artifactId>
-            <version>3.4.0</version>
-        </dependency>
-        <dependency>
-            <groupId>com.twitter</groupId>
-            <artifactId>finagle-thrift_2.9.2</artifactId>
-            <version>6.5.1</version>
+            <version>0.10.0</version>
         </dependency>
     </dependencies>
 
     <build>
         <plugins>
             <plugin>
-                <groupId>com.twitter</groupId>
-                <artifactId>scrooge-maven-plugin</artifactId>
-                <!-- this version must match the version from scrooge-gen.ivy.xml -->
-                <version>3.16.0</version>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.6.1</version>
                 <configuration>
-                    <language>java</language>
-                    <!-- default is scala, can also be java -->
-                    <thriftOpts>
-                        <!-- add other Scrooge command line options using thriftOpts -->
-                        <thriftOpt>--finagle</thriftOpt>
-                    </thriftOpts>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.thrift</groupId>
+                <artifactId>thrift-maven-plugin</artifactId>
+                <version>0.10.0</version>
+                <configuration>
+                    <thriftExecutable>/usr/local/bin/thrift</thriftExecutable>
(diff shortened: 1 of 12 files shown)


controller/controller-common/pom.xml
```diff
@@ -17,6 +17,55 @@
...
+            <artifactId>finagle-thrift_2.9.2</artifactId>
+            <version>6.5.1</version>
+        </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.twitter</groupId>
+                <artifactId>scrooge-maven-plugin</artifactId>
```
shuzhang1989: any reason using scrooge? are we using finagle?

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/HealthCheckTask.java
```diff
@@ -0,0 +1,166 @@
...
+
+  public HealthCheckTask(Param param) {
+    super(param);
+  }
+
+  @Override
+  public void process(Context ctx) throws Exception {
+    final String clusterName = getParameter().getClusterName();
+    final String zkPath = getParameter().getZkPath();
+    final InterProcessLock lock =
```
shuzhang1989: i think we decided to use mysql task queue lock instead of zk lock. @newpoo to confirm

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/HealthCheckTask.java
```diff
@@ -0,0 +1,166 @@
...
+  public AdminClientFactory clientFactory;
+
+  public HealthCheckTask(Param param) {
+    super(param);
+  }
+
+  @Override
+  public void process(Context ctx) throws Exception {
+    final String clusterName = getParameter().getClusterName();
+    final String zkPath = getParameter().getZkPath();
```
shuzhang1989: should we have a convention that we can get a zk path from cluster name? so we dont need to give zk path.

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/HealthCheckTask.java
```diff
@@ -0,0 +1,166 @@
...
+        clientFactory.getClient(hostAddr).ping();
+      }
+
+      LOG.info("All hosts are good");
+
+      ctx.getTaskQueue().finishTask(ctx.getId(),
+          String.format("Cluster %s is healthy", clusterName));
+    } catch (Exception ex) {
+      ctx.getTaskQueue().failTask(ctx.getId(),
+          String.format("Cluster %s is unhealthy, reason = %s", clusterName, ex.getMessage()));
```
shuzhang1989: we will be interested to know which ip is not pingable. do we have ip in the exception?

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/HealthCheckTask.java
newpoo: Yes, I think so.

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/HealthCheckTask.java
angxu: You mean make zkPath a constant?

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/HealthCheckTask.java
```diff
@@ -0,0 +1,161 @@
...
+  private AdminClientFactory clientFactory;
+
+  public HealthCheckTask(Param param) {
+    super(param);
+  }
+
+  @Override
+  public void process(Context ctx) throws Exception {
+    final String clusterName = ctx.getCluster();
+    final String zkPath = getParameter().getZkPath();
```
shuzhang1989: you can call task queue library interface to grab the lock?

angxu: I thought the lock is already hold once you dequeued task?

PR #39, opened 2017-03-27
Add support for chained and retry task

angxu:
Add two useful util methods for task composition.
- Task#andThen()
- Task#retry()

### DIFF
--- a/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/DelayAckTaskQueue.java
+++ b/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/DelayAckTaskQueue.java
@@ -1,0 +1,165 @@
+/*
+ *  Copyright 2017 Pinterest, Inc.
+ *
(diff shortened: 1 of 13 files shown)


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/DelayAckTaskQueue.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * A {@link TaskQueue} wrapper which delays the calls to the actual
+ * {@link #finishTask(long, String)} and {@link #failTask(long, String)}.
+ *
+ * It enables ChainedTask to achieve ack-and-enqueue operation atomically.
+ */
+class DelayAckTaskQueue implements TaskQueue {
```
shuzhang1989: delay is a little confusing naming.
maybe rename to "LocalAckTaskQueue"?

PR #43, opened 2017-03-31
Add RemoveHostTask

angxu:
Add a controller task to remove host from a given cluster.

### DIFF
--- a/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/RemoveHostTask.java
+++ b/controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/RemoveHostTask.java
@@ -88,15 +88,18 @@ public class RemoveHostTask extends TaskBase<RemoveHostTask.Param> {
       return;
     }
     for (SegmentBean segmentBean : clusterBean.getSegments()) {
+      // filter out the host to remove
       List<HostBean> newHostList = segmentBean.getHosts().stream()
-          .filter(hostBean -> !hostBean.equals(toRemove))
+          .filter(hostBean ->
+              !(hostBean.getIp().equals(toRemove.getIp()) &&
+                    hostBean.getPort() == toRemove.getPort()))
           .collect(Collectors.toList());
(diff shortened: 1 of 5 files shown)


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/RemoveHostTask.java
```diff
@@ -0,0 +1,119 @@
...
+ * This task removes a given host from a cluster. It consists of the following steps:
+ *
+ * 1) Check if the host is still up and running. Fail if yes.
+ * 2) Update cluster config by removing that host.
+ *
+ * @author Ang Xu (angxu@pinterest.com)
+ */
+public class RemoveHostTask extends TaskBase<RemoveHostTask.Param> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HealthCheckTask.class);
```
shuzhang1989: RemoveHostTask.class

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/RemoveHostTask.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Override
+  public void process(Context ctx) throws Exception {
+    final String clusterName = ctx.getCluster();
+    final HostBean toRemove = getParameter().getHostToRemove();
+
+    InetSocketAddress hostAddr = new InetSocketAddress(toRemove.getIp(), toRemove.getPort());
+    Admin.Client client = clientFactory.getClient(hostAddr);
+
+    // 1) ping the host to remove to make sure it's not running
```
shuzhang1989: maybe this check is too strict. sometimes we may want to remove a running but having other issue's host.

angxu: Copied this logic from rocksdb_admin.py... 
How about we add a flag like removeEvenIfHostIsRunning to this task and make it less error-prone?

PR #47, opened 2017-04-04
Implement rebalance task

angxu:


shuzhang1989:
let's merge all previous ones and rebase so the review can be easier

PR #49, opened 2017-04-06
[rocksplicator/s3util] reuse s3util client for sst downloading

newpoo:
S3 client is very heavy (internal thread pool), we couldn't afford to create one for every sst downloading request.

This PR has the potential to remove the need for two clusters for p2p and realpinnability.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -539,6 +539,26 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
     std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
       AddS3SstFilesToDBResponse>>> callback,
     std::unique_ptr<AddS3SstFilesToDBRequest> request) {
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -539,6 +539,26 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  if (s3_download_limit_mb != request->s3_download_limit_mb ||
+      s3_bucket != request->s3_bucket) {
+    // TODO(bol) currently we use the parameters in the first request recieved.
+    // Fixing me by upgrading to new s3 sdk and changing the S3Util interface.
+    LOG(ERROR) << "Ignoring download limit: " << request->s3_download_limit_mb
+               << " and s3 bucket: " << request->s3_bucket
+               << " use " << s3_download_limit_mb
+               << " and " << s3_bucket
+               << " instead.";
+  }
```
anshulpundir: should we fail instead of continuing ?

PR #51, opened 2017-04-10
log less in cpu thread pool

shuzhang1989:
the log is too noisy in realpin. make it log every n.

### DIFF
--- a/common/global_cpu_executor.cpp
+++ b/common/global_cpu_executor.cpp
@@ -18,7 +18,12 @@
 
 #include "common/global_cpu_executor.h"
 
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+
 #include "wangle/concurrent/CPUThreadPoolExecutor.h"
+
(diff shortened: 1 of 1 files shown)


common/global_cpu_executor.cpp
```diff
@@ -18,17 +18,23 @@
...
   static const auto n_threads = sysconf(_SC_NPROCESSORS_ONLN);
   static const auto queue_sz = (1 << 14);
 
-  LOG(INFO) << "Running global CPU thread pool with "
-            << n_threads << " threads, and "
-            << queue_sz << " length queue.";
+  LOG_EVERY_N(INFO, FLAGS_cpu_thread_pool_log_evey_n)
+                           << "Running global CPU thread pool with "
+                           << n_threads << " threads, and "
+                           << queue_sz << " length queue.";
```
newpoo: We could either store a pointer at the client side code, or make two anonymouse functions. One to log and return n_threads, one to log and return queue_sz. And call them below.

PR #53, opened 2017-04-12
Implement LoadSSTTask and the corresponding http endpoint

angxu:


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/LoadSSTTask.java
```diff
@@ -0,0 +1,230 @@
...
+      String errMsg = String.format("Segment %s not in cluster %s.", segment, clusterName);
+      LOG.error(errMsg);
+      ctx.getTaskQueue().failTask(ctx.getId(), errMsg);
+      return;
+    }
+
+    final ExecutorService executor = Executors.newFixedThreadPool(getParameter().getConcurrency());
+
+    // first pass load sst to masters
+    List<Future<Boolean>> futures = new ArrayList<>(segmentBean.getHosts().size());
```
shuzhang1989: first pass and second pass code looks pretty similar, put them into one function?

angxu: sure.

PR #54, opened 2017-04-12
Implement AdminClientFactory with caching

angxu:


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java
```diff
@@ -35,10 +35,12 @@
   private static final String WORKER_POOL_SIZE_KEY = "worker_pool_size";
   private static final String DISPATCHER_POLL_INTERVAL_KEY = "dispatcher_poll_interval";
   private static final String ZK_PATH_KEY = "zk_path";
+  private static final String ZK_ENDPOINTS_KEY = "zk_endpoints";
   private static String HOST_NAME;
   private static PropertiesConfiguration configuration;
 
   private static final String DEFAULT_ZK_PATH = "/config/services/rocksdb/";
+  private static final String DEFAULT_ZK_ENDPOINTS = "observerzookeeper010:2181";
```
shuzhang1989: nit: do we support a list of zk endpoint? i am not sure about curator, but for other zk clients, they support pass in endpoints separated by comma.

but given our very low load to zk, probably not a issue for now.

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java
```diff
@@ -55,9 +78,37 @@ public static void main(String[] args) {
...
           WorkerConfig.getDispatcherPollIntervalSec(), idleWorkersSemaphore, workerPool, taskQueue);
       dispatcher.start();
+      SHUTDOWN_HOOK.register(dispatcher::stop);
     } catch (Exception e) {
       LOG.error("Cannot start the worker service", e);
       System.exit(1);
     }
   }
+
+  private static class ShutdownHook implements Runnable {
```
shuzhang1989: let's move this util to common, in case other service needs it

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerConfig.java
angxu: Curator does support that format. It will just pass this endpoint string to the constructor of vanilla zookeeper.

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/WorkerService.java
angxu: sure.

PR #55, opened 2017-04-12
Update s3 util

cpd85:
This will prevent an old bug in AWS SDK from occurring, where libcurl causes the program to crash


common/tests/s3_util_test.cpp
```diff
@@ -51,11 +52,27 @@ TEST(S3UtilTest, ParseS3StringTest) {
   EXPECT_TRUE(std::get<1>(result) == "");
 }
 
+TEST(S3UtilTest, GetObjectS3Test) {
+  static shared_ptr<common::S3Util> myUtil =
+      common::S3Util::BuildS3Util(50, "media.pinterest.com", 60000, 60000);
```
shuzhang1989: let's remove this test.
it exposes some pinterest internal information.

common/s3util.h
```diff
@@ -87,9 +89,12 @@ class S3Util {
   };
 
   // Don't recommend using this directly. Using BuildS3Util instead.
+  // If you must, make sure to call Aws::InitAPI(options); before constructor
   S3Util(const string& bucket,
-         const ClientConfiguration& client_config) :
-    bucket_(std::move(bucket)), s3Client(client_config) {
+         const ClientConfiguration& client_config,
+         SDKOptions options) :
```
shuzhang1989: const SDKOptions&

common/tests/s3_util_test.cpp
shuzhang1989: also this test will be flaky, what if the s3 object is removed?

common/s3util.h
```diff
@@ -36,6 +37,7 @@ using std::string;
 using std::vector;
 using std::tuple;
 using std::shared_ptr;
+using Aws::SDKOptions;
```
newpoo: nit: make it sorted?

common/s3util.h
```diff
@@ -87,9 +89,12 @@ class S3Util {
   };
 
   // Don't recommend using this directly. Using BuildS3Util instead.
+  // If you must, make sure to call Aws::InitAPI(options); before constructor
```
newpoo: In this case, we can make it private?

common/s3util.cpp
```diff
@@ -185,7 +185,9 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
   aws_config.readRateLimiter = 
     std::make_shared<DefaultRateLimiter<>>(
           read_ratelimit_mb * 1024 * 1024);
-  return std::make_shared<S3Util>(bucket, aws_config);
+  SDKOptions options;
+  Aws::InitAPI(options);
```
newpoo: What happens if this gets called twice?

.travis.yml
```diff
@@ -9,4 +9,4 @@ before_install:
 - docker pull aliced3645/rocksplicator-build
```
angxu: change it to cdonaghy/rocksplicator-build?

PR #58, opened 2017-04-14
Complete cluster level queue operations

shuzhang1989:


shuzhang1989:
as chatted with @angxu, using jpa to do orm to avoid directly managing mysql connections and mappings.

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
```diff
@@ -17,71 +17,130 @@
...
+      locked = isClusterLocked(clusterName);
+    } catch (MySQLTaskQueueException e) {
+      LOG.error("Cluster {} hasn't been created", clusterName);
+    } finally {
+      if (locked) {
+        LOG.error("Cluster {} is already locked, cannot double lock", clusterName);
+        return false;
+      }
+    }
+    Query query = entityManager.createNamedQuery("tag.lock").setParameter("name", clusterName);
```
angxu: What's the result of this query is the # of affected row is 0? I think we need to make sure whether the caller indeed acquire the lock or if the lock is already held by others.

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
```diff
@@ -17,71 +17,130 @@
...
+    }
+    return true;
+  }
+
+
+  @Override
+  public boolean removeCluster(final String clusterName) {
+    boolean locked = true;
+    try {
+      locked = isClusterLocked(clusterName);
```
angxu: include this in the transaction?

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
shuzhang1989: the check is performed above - we check if the cluster is existed, or it's already locked

angxu: Then we should make it in the same transaction, right? Otherwise two concurrent calls can both succeed.

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
shuzhang1989: same thought as above, separate for precondition and actual transaction. (written in the doc)

controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/TaskBase.java
```diff
@@ -16,8 +16,6 @@
 
 package com.pinterest.rocksplicator.controller.tasks;
 
-import com.pinterest.rocksplicator.controller.TaskEntity;
```
angxu: Rename this class to something else? For example, AbstractTask?

shuzhang1989: sounds good

PR #60, opened 2017-04-20
Complete remaining task operations

shuzhang1989:


controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Task.java
```diff
@@ -27,10 +27,12 @@
  */
 public class Task extends TaskBase {
 
-  public static int PENDING = 0;
-  public static int RUNNING = 1;
-  public static int DONE = 2;
-  public static int FAILED = 3;
+  public enum TaskState {
```
angxu: There is already a TaskState enum defined in controller-common. You can reuse that one.

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskQueue.java
```diff
@@ -143,9 +143,10 @@ default boolean removeCluster(final String cluster) {
 
   /**
    * Remove finished tasks.
+   * @param secondsAgo tasks finished seconds ago
    * @return the number of finished tasks which are removed by this call.
    */
-  default int removeFinishedTasks() {
+  default int removeFinishedTasks(final int secondsAgo) {
```
angxu: Better to define such API as `#removeFinishedTasks(long, TimeUnit)`.

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
```diff
@@ -138,28 +154,43 @@ public boolean removeCluster(final String clusterName) {
...
-  public boolean enqueueTask(final TaskBase taskBase,
-                             final String clusterName,
-                             final int runDelaySeconds) {
-    entityManager.getTransaction().begin();
+  private long enqueueTaskImpl(final TaskBase taskBase,
+                               final String clusterName,
+                               final int runDelaySeconds,
+                               final Task.TaskState state,
+                               final String claimedWorker,
+                               boolean inTransaction) {
```
angxu: Or we can just make this method non-transactional and wrap it with `entityManager.getTransaction().begin()` and `entityManager.getTransaction().commit()` in the outer method.

Also for transaction, do we need to do `try {...} finally {...}`?

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Task.java
shuzhang1989: sounds good 👍

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/TaskQueue.java
shuzhang1989: i think our default time unit is all seconds if you look at other api and design doc.

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
shuzhang1989: sure, how about put transaction().rollback() in finally

PR #63, opened 2017-04-25
Add direct I/O support to s3util

angxu:


common/s3util.h
```diff
@@ -64,6 +66,61 @@ class S3UtilResponse {
...
+  uint32_t offset_;
+};
+
+/**
+ * A wrapper class of DirectIOFile which can be used together with
+ * boost::iostreams::stream to implement i{o}stream.
+ */
+class DirectIOFileSink {
+ public:
+  typedef char                                        char_type;
```
shuzhang1989: nit: prefer using char_type = char

common/s3util.cpp
```diff
@@ -18,6 +18,12 @@
 
 #include "common/s3util.h"
 
+#include <fcntl.h>
+#include <stdlib.h>
```
newpoo: nit: can you include cstdlib instead ?

common/s3util.cpp
```diff
@@ -18,6 +18,12 @@
...
+#include <unistd.h>
+#include <sys/types.h>
```
newpoo: nit: could you sort them?

common/s3util.h
```diff
@@ -111,10 +168,12 @@ class S3Util {
   }
 
   // Download an S3 Object to a local file
-  GetObjectResponse getObject(const string& key, const string& local_path);
+  GetObjectResponse getObject(const string& key, const string& local_path,
+                              const bool direct_io=false);
```
newpoo: nit: keep one space before and after "=", same foe line 176.

common/s3util.cpp
```diff
@@ -41,11 +50,77 @@ using Aws::S3::Model::ListObjectsRequest;
...
+    , file_size_(0)
+    , buffer_()
+    , offset_(0) {
+  if (posix_memalign(&buffer_, kPageSize, kPageSize) != 0) {
+    LOG(ERROR) << "Failed to allocate memaligned buffer, errno = " << errno;
+    return;
+  }
+
+  int flag = O_WRONLY | O_TRUNC | O_CREAT | O_DIRECT;
+  fd_ = open(file_path.c_str(), flag , 640);
```
newpoo: nit: can we replace 640 with pre-defined flags?

common/s3util.cpp
```diff
@@ -41,11 +50,77 @@ using Aws::S3::Model::ListObjectsRequest;
...
+  }
+  uint32_t remaining = static_cast<uint32_t>(n);
+  while (remaining > 0) {
+    uint32_t bytes = std::min((kPageSize - offset_), remaining);
+    std::memcpy((char*)buffer_ + offset_, s, bytes);
+    offset_ += bytes;
+    remaining -= bytes;
+    // flush when buffer is full
+    if (offset_ == kPageSize) {
+      if (::write(fd_, buffer_, kPageSize) != kPageSize) {
```
shuzhang1989: check errno ?

common/s3util.cpp
```diff
@@ -41,11 +50,77 @@ using Aws::S3::Model::ListObjectsRequest;
...
+}
+
+std::streamsize DirectIOFile::write(const char* s, std::streamsize n) {
+  if (buffer_ == nullptr || fd_ < 0) {
+    return -1;
+  }
+  uint32_t remaining = static_cast<uint32_t>(n);
+  while (remaining > 0) {
+    uint32_t bytes = std::min((kPageSize - offset_), remaining);
+    std::memcpy((char*)buffer_ + offset_, s, bytes);
```
shuzhang1989: s + n - remaining?

common/s3util.h
```diff
@@ -64,6 +66,61 @@ class S3UtilResponse {
...
+
+  DirectIOFileSink(const string& file_path)
+      : file_(std::make_shared<DirectIOFile>(file_path)) {
+  }
+
+  std::streamsize write(const char* s, std::streamsize n) {
+    return file_->write(s, n);
+  }
+
+  std::streamsize read(char* s, std::streamsize n) {
```
newpoo: Do we have to implement read(), as it is a sink?

common/s3util.cpp
angxu: nice catch!

common/s3util.h
```diff
@@ -64,6 +66,61 @@ class S3UtilResponse {
   string error_;
 };
 
+/**
+ * A writable file which uses direct I/O under the hood.
+ */
+class DirectIOFile {
```
newpoo: nit: forgot to mention previously. Can we rename it to DirectIOWritableFile ?

common/s3util.cpp
angxu: I think it's simpler to just fail unconditionally. But we can log the errno so that it's easier to reason failures.

common/s3util.h
angxu: The aws sdk requires a iostream. However, if we don't implement read(), `boost::iostreams::stream<DirectIOFileSink>(local_path)` will return a ostream instead of iostream.

common/s3util.cpp
newpoo: can you change/add a test to catch it?

angxu: sure.

angxu: It seems hard to catch this in test because `boost::iostreams::stream` is probably doing some internal buffering and never write buffer larger 4096...

shuzhang1989: i think one of your use case us using 4097 'g's. you can change that to have different chars in the string

common/s3util.cpp
angxu: I couldn't find pre-defined flags to this. Do you have any pointer?

common/s3util.cpp
angxu: done!

common/s3util.cpp
newpoo: I got these from man 2 open. do they work?

The following symbolic constants are provided for mode:

              S_IRWXU  00700 user (file owner) has read, write and execute permission

              S_IRUSR  00400 user has read permission

              S_IWUSR  00200 user has write permission

              S_IXUSR  00100 user has execute permission

              S_IRWXG  00070 group has read, write and execute permission

              S_IRGRP  00040 group has read permission

              S_IWGRP  00020 group has write permission

              S_IXGRP  00010 group has execute permission

              S_IRWXO  00007 others have read, write and execute permission

              S_IROTH  00004 others have read permission

              S_IWOTH  00002 others have write permission

              S_IXOTH  00001 others have execute permission

angxu: Yeah, they work. Thanks!

PR #65, opened 2017-04-27
Adding flags to do retry for a getObject failures.

anshulpundir:
We see errors of the following nature quite often from AWS when downloading SST files for counter-ads:

"Unable to parse ExceptionName: InternalError Message: We encountered an internal error. Please try again."

Since this is an internal S3 error, retrying is the best bet. We enabled retrying on the python job level, but errors show up on different objects which doesn't improve the chance of success that much. Adding retry to to individual should help.


shuzhang1989:
let's put this retry logic in client side.

PR #66, opened 2017-05-12
Fix a race condition in cached_iter evb

shuzhang1989:
the evb's loopForever() call may not be running when runAfterDelay() is called, hence we see Assertion `isInEventBaseThread()' failed error which made tests flaky.


angxu:
Where did the assertion fail at? Maybe I missed it, but I checked runAfterDelay() method and it doesn't seem to have such check.

shuzhang1989:
https://github.com/facebook/folly/blob/master/folly/io/async/TimeoutManager.cpp#L97

PR #68, opened 2017-05-17
Support Serverset routing in thrift router

shuzhang1989:
Extend thrift_router to be able to parse and select client from serverset.

The motivation is that by reusing the logic we did for shard config, we can ensure we always get good client from the serverset, so we dont need to handle retries in serverset_thrift_client_pool, which is less efficient because we need to send out actual request to know the channel is broken, and try other clients.

### DIFF
--- a/common/thrift_router.h
+++ b/common/thrift_router.h
@@ -22,6 +22,7 @@
 #include <functional>
 #include <map>
 #include <memory>
+#include <random>
 #include <set>
 #include <string>
 #include <vector>
@@ -32,6 +33,7 @@
 #include "common/availability_zone.h"
 #include "common/thrift_client_pool.h"
 #include "folly/RWSpinLock.h"
+#include "folly/ScopeGuard.h"
 #include "folly/SocketAddress.h"
 #include "folly/ThreadLocal.h"
 
@@ -53,7 +55,6 @@ struct Host {
   bool operator<(const Host& other) const {
     return addr < other.addr;
   }
-
   folly::SocketAddress addr;
   std::string az;  // availability zone
 };
@@ -126,6 +127,7 @@ class ThriftRouter {
 
   /*
    * Get clients pointing to the server(s) matching the request conditions.
+   * @note This function is valid when parser function is parseConfig().
    * @note It is encouraged and performance-wise OK to call getClientsFor() for
    *       every RPC.
    * @param segment   The requested segment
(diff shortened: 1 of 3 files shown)


shuzhang1989:
@jinruh yeah, i think that's another way. the motivation doing here is that we can have thrift_router support both stateless and stateful routing modes, making it a centralized routing class. creating another thrift_client_pool in cosmos may be confusing, and needs to do similar thread local work in thrift_router, which also needs some work. what do you think?

common/thrift_router.h
```diff
@@ -207,6 +220,15 @@ class ThriftRouter {
     return itor->second.shard_to_hosts.size();
   }
 
+  uint32_t getHostsCount() {
```
chengren311: getHostCount?

common/thrift_router.h
```diff
@@ -307,6 +329,31 @@ class ThriftRouter {
...
+      for (int i = 0; i < serverset_size; i ++){
+        // Keep searching until we get a good client
+        folly::ScopeGuard round_robin_guard = folly::makeGuard([&] {
+          round_robin_counter ++;
+        });
+        auto selected_host = (*hosts_).at(round_robin_counter % serverset_size);
+        if (!createOrFixClientFor(selected_host)) {
+          continue;
+        }
+        return (*clients_)[selected_host->addr].client;
```
chengren311: should we also log when a good client is found?

shuzhang1989: no

common/thrift_router.h
```diff
@@ -410,28 +464,33 @@ class ThriftRouter {
...
 
-    void createOrFixClientsFor(const std::set<const Host*>& hosts) {
-      for (auto host : hosts) {
-        auto itor = clients_->find(host->addr);
-        if (itor != clients_->end() &&
-            is_client_good(itor->second.client.get())) {
-            // has the client and it is good
-          continue;
-        }
+    bool createOrFixClientFor(const Host* host) {
```
chengren311: maybe it would be helpful to add some method description here on what it returns.

newpoo:
Let's do it in a separate serverset_thrift_router class in cosmos first.
Besides what Jinru said, we have to use different client protocol/type to talk to apache thrift server.

shuzhang1989:
i will do it in cosmos.

PR #71, opened 2017-05-23
Make s3 download adaptive for different ratelimits

shuzhang1989:
We need support s3client for different rates without restarting the service.

### DIFF
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -286,7 +286,8 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
           read_ratelimit_mb * 1024 * 1024);
   SDKOptions options;
   Aws::InitAPI(options);
-  return std::make_shared<S3Util>(bucket, aws_config, options);
+  return std::make_shared<S3Util>(
+          bucket, aws_config, options, read_ratelimit_mb);
 }
 
 }  // namespace common
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -19,6 +19,7 @@
 
 #include "rocksdb_admin/admin_handler.h"
 
+#include <atomic>
 #include <memory>
 #include <string>
 #include <vector>
@@ -26,7 +27,6 @@
 #include "boost/filesystem.hpp"
 #include "common/network_util.h"
 #include "common/rocksdb_glogger/rocksdb_glogger.h"
-#include "common/s3util.h"
 #include "common/thrift_router.h"
 #include "folly/FileUtil.h"
 #include "folly/ScopeGuard.h"
@@ -553,6 +553,12 @@ void AdminHandler::async_tm_clearDB(
   callback->result(ClearDBResponse());
 }
 
+inline bool should_new_s3_client(
+  shared_ptr<common::S3Util> s3_util, AddS3SstFilesToDBRequest* request) {
+  return s3_util == nullptr || s3_util->getBucket() != request->s3_bucket ||
+      s3_util->getRateLimit() != request->s3_download_limit_mb;
+}
+
 void AdminHandler::async_tm_addS3SstFilesToDB(
     std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
       AddS3SstFilesToDBResponse>>> callback,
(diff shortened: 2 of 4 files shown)


common/s3util.cpp
```diff
@@ -273,7 +273,8 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
           read_ratelimit_mb * 1024 * 1024);
   SDKOptions options;
   Aws::InitAPI(options);
```
newpoo: Will there be any side effect if we call it multiple times in a process?

rocksdb_admin/admin_handler.cpp
```diff
@@ -19,14 +19,14 @@
 
 #include "rocksdb_admin/admin_handler.h"
 
+#include <atomic>
```
newpoo: remove it?

rocksdb_admin/admin_handler.cpp
```diff
@@ -605,7 +607,7 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
     return;
   }
 
-  auto responses = s3Util->getObjects(request->s3_path, local_path);
+  auto responses = s3_util_->getObjects(request->s3_path, local_path);
```
newpoo: what if someone changed s3_util_ before this line and after line 608?

I think we can do it in a simpler way. We just have a shared_ptr and a spin lock to protect it.

When we need to use it, we grab the lock, make a local copy of the shared_ptr, and release the lock. and then check if the s3 client is what we need. If yes, we just use it. Otherwise, we use it, and store it to the global shared_ptr (acquire the spin lock again)

rocksdb_admin/admin_handler.cpp
```diff
@@ -538,6 +538,12 @@ void AdminHandler::async_tm_clearDB(
   callback->result(ClearDBResponse());
 }
 
+inline bool should_new_s3_client(
+  shared_ptr<common::S3Util> s3_util, AddS3SstFilesToDBRequest* request) {
```
newpoo: nit: it's more performant to change it to const common::S3Util& s3_util.

common/s3util.cpp
shuzhang1989: no, this InitAPI is there before we do the global single s3client.

rocksdb_admin/admin_handler.cpp
shuzhang1989: sure

rocksdb_admin/admin_handler.cpp
shuzhang1989: sure

rocksdb_admin/admin_handler.cpp
shuzhang1989: these block are all protected by the read lock at line 555, so if we are executing this line, there should be no opportunity for others the change the s3_util_.

rocksdb_admin/admin_handler.cpp
```diff
@@ -546,30 +551,26 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
   // see latency spike when uploading data to production clusters.
-  static const uint32_t s3_download_limit_mb = request->s3_download_limit_mb;
-  static const std::string s3_bucket = request->s3_bucket;
-  static auto s3Util = common::S3Util::BuildS3Util(
-    s3_download_limit_mb, s3_bucket);
+  RWSpinLock::UpgradedHolder upgraded_guard(s3_util_lock_);
+  if (s3_util_ == nullptr || should_new_s3_client(*s3_util_, request.get())) {
+    // Request with different ratelimit or bucket has to wait for old
+    // requests to drain.
+    RWSpinLock::WriteHolder write_guard(s3_util_lock_);
```
newpoo: shouldn't you upgrade from upgraded_guard?
Then the lock will be released in the if() block?

common/s3util.cpp
newpoo: I don't think we have it before we use the global client.
InitAPI() was introduced by the new aws library, which was added by Connell after we switched to the global client.
From AWS page https://aws.amazon.com/cn/blogs/developer/aws-sdk-for-c-simplified-configuration-and-initialization/  it's more like a global initialization function.

rocksdb_admin/admin_handler.cpp
shuzhang1989: awesome, it's a good catch

common/s3util.cpp
shuzhang1989: oh, i remembered it wrong. we need to make it a call once

shuzhang1989: I took a look again, we will do shutdownAPI() at the deconstructor of S3Util. this means when we create a new s3util, the old shutdownAPi will be called. so we still need to call initApi() again

shuzhang1989:
ping - updated the lock issue :)

rocksdb_admin/admin_handler.h
```diff
@@ -35,6 +37,7 @@ namespace admin {
 
 using RocksDBOptionsGeneratorType =
   std::function<rocksdb::Options(const std::string&)>;
+using folly::RWSpinLock;
```
newpoo: nit: can we remove this line, and add folly:: to line 109?

rocksdb_admin/admin_handler.cpp
```diff
@@ -538,6 +537,12 @@ void AdminHandler::async_tm_clearDB(
   callback->result(ClearDBResponse());
 }
 
+inline bool should_new_s3_client(
+  const common::S3Util& s3_util, AddS3SstFilesToDBRequest* request) {
```
newpoo: nit: two more spaces.

shuzhang1989:
@newpoo you can take a look again

rocksdb_admin/admin_handler.cpp
```diff
@@ -559,30 +567,26 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  {
+    std::lock_guard<std::mutex> guard(s3_util_lock_);
+    if (s3_util_ == nullptr || should_new_s3_client(*s3_util_, request.get())) {
+      // Request with different ratelimit or bucket has to wait for old
+      // requests to drain.
+      while (s3_util_ != nullptr && s3_util_.use_count() > 1) {
+        LOG(INFO) << "There are other downloads happening, wait "
+                  << kS3UtilRecheckSec << " seconds";
+        std::this_thread::sleep_for(std::chrono::seconds(kS3UtilRecheckSec));
+      }
```
newpoo: why do we need to wait here?

shuzhang1989: If we do not wait, we will potentially end up calling Aws::initApi() more than Aws::ShutdownApi() when there are multiple concurrent requests. Although from the code and asked aws people if that is allowed to do and they said fine for most cases, i think making the calls always occur in pair will be safer.

PR #72, opened 2017-05-31
Make ThriftRouter support different thrift client pool type

jinruh:


shuzhang1989:
what's the motivation of the change?
i am working on a thrift router of serverset, which can be used to support terrapin serverset. not sure if this change will make the serverset work meaningless because looks you are going to talk to terrapin leaf anyways.

also for the serverset router change, i found there will be a lot of duplicate code (primiary on the logic of the connection check) if i implement in cosmos, however https://github.com/pinterest/rocksplicator/pull/68 looks to have less change.

jinruh:
The main motivation is to save one hop between scorpion and terrapin. This is also what scorpion does when talking to Realpin. Pinacle also directly talk to Terrapin leaf.

I think it's necessary to have serverset based thrift router given that the current serverset thrift pool is not implemented ideally. I believe the new stateless thrift router will perform much better than existing one, and people should use the new one when they want to talk to stateless service.

newpoo:
Besides saving one hop, the current Terrapin root doesn't work well in terms of performance.

I still think we should implement it in a separate class instead of merging it into the current stateful router. Because the current one is already very complicated. Adding more logic to it will make it very hard to maintain in future. Also, it has some logics that the stateless one doesn't need.

shuzhang1989:
ok,  as long as the duplicate code is not a big concern.

PR #89, opened 2017-07-07
Make it possible to set threads for global thread pool and thrift client pool

shuzhang1989:


common/global_cpu_executor.cpp
```diff
@@ -23,10 +23,14 @@
 
 #include "wangle/concurrent/CPUThreadPoolExecutor.h"
 
+DEFINE_int32(global_worker_threads, sysconf(_SC_NPROCESSORS_ONLN),
+             "The number of threads for global CPU executor.");
+
```
newpoo: This won't work reliably when GetThreadsCount() is called before entering main(), such as static variable initialization. Because the gflag could be initialized after that.

common/thrift_client_pool.h
```diff
@@ -40,6 +40,8 @@ DECLARE_int32(channel_max_checking_size);
 
 DECLARE_int32(channel_send_timeout_ms);
 
+DECLARE_int32(thrift_client_pool_threads);
```
newpoo: can we rename it to default_thrift_client_pool_threads?

Because client can pass in a different value to ThriftClientPool() to overwrite the value provided by this flag.

common/global_cpu_executor.cpp
shuzhang1989: yes i thought about this. is there any case we call GetThreadCounts() outside main?

shuzhang1989: should we just remove GetThreadsCount()?

newpoo: I meant to say getGlobalCPUExecutor().
We used to have a gflag for exactly the same purpose, and we hit this tricky bug in Scorpion. So it got removed eventually.

common/global_cpu_executor.cpp
```diff
@@ -23,10 +23,17 @@
...
 namespace {
 
-long GetThreadsCount() {
-  static const auto n_threads = sysconf(_SC_NPROCESSORS_ONLN);
+uint16_t GetThreadsCount() {
+  auto worker_thread_count =
+    FLAGS_global_worker_threads == 0?
```
newpoo: nit: one space between 0 and ?

PR #90, opened 2017-07-14
added more mysql queries to suit the need for peekTasks()

ywang282:
/tasks path assumes either both STATE and CLUSTER_NAME or either one of them is specified in the parameter. However, peekTasks() fails to get any matching rows in the db when only one of the field is present. Because the default value for both STATE and CLUSTER_NAME is NULL and mysql is matching for CLUSTER_NAME = name AND STATE = null (eg, when STATE is not provided).


shuzhang1989:
do you also want to change the http layer to not throw exception?

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
```diff
@@ -369,9 +369,26 @@ public boolean finishTaskAndEnqueuePendingTask(final long id,
 
   @Override
   public List<Task> peekTasks(final String clusterName, final Integer state) {
-    Query query = getEntityManager()
-        .createNamedQuery("task.peekTasks")
+    Query query;
+    if (clusterName != null && state != null){
```
shuzhang1989: nit: space between ) and {

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
```diff
@@ -369,9 +369,26 @@ public boolean finishTaskAndEnqueuePendingTask(final long id,
...
+      query = getEntityManager()
+        .createNamedQuery("task.peekTasksFromClusterWithState")
         .setParameter("state", state).setParameter("name", clusterName);
+    }
+    else if (clusterName == null){
```
shuzhang1989: nit: put it in the last line
if {
} else if {
}

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
ywang282: fixed

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLTaskQueue.java
ywang282: fixed

shuzhang1989:
awesome!

PR #92, opened 2017-07-17
created the Result class and made partial http endpoints user friendly

ywang282:
All the http response will have the following structure
test


angxu:
Just curious, why do we have a separate status code in response body? And would it ever be different from the response status code?

shuzhang1989:
talked with @ywang282 offline, we prefer not to change core library interface for now. we will do the response check and return richer response in http layer

PR #93, opened 2017-07-17
add log4j lib dependency to worker

shuzhang1989:


angxu:
We already have logback dependency coming from dropwizard. Can we use that instead? If not, we need to ban that dependency as otherwise the behavior is undefined.

shuzhang1989:
didnt notice the log back dependency.. need to revert this diff

PR #95, opened 2017-07-18
Upgrade RocksDB version to 5.4.7

animesht:
As title

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -38,6 +38,8 @@ RUN apt-get -q -y update && \
       flex \
       g++-5 \
       gcc-5 \
+      g++-4.7 \
+      gcc-4.7 \
       gdb \
       git \
       hadoop \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -124,16 +126,16 @@ RUN cd /opt && \
 
 # jemalloc
 RUN cd /opt && \
-    wget https://github.com/jemalloc/jemalloc/archive/3.6.0.tar.gz && \
-    tar -zxvf 3.6.0.tar.gz && \
-    (cd jemalloc-3.6.0 && \
+    wget https://github.com/jemalloc/jemalloc/archive/4.5.0.tar.gz && \
+    tar -zxvf 4.5.0.tar.gz && \
```
cpd85: should this be 4.5.7?

cpd85:
could you also push your docker image, and then modify the README so that we instruct users to pull from the new docker image? Currently we pull from cdonaghy/rocksplicator-build, which has the old RocksDB version

shuzhang1989:
have we tested that the all binaries can be compiled locally, and all tests pass?

docker/Dockerfile
animesht: This is for jemalloc, I just ported over the version being used in the cosmos Dockerfile, since that has been updated more recently than the one for Rocksplicator

cpd85: oops, you're right sorry for the confusion

PR #97, opened 2017-07-18
support e2e run healthcheck

shuzhang1989:


cpd85:
should we use CheckDBRequest for each shard to see if the host is capable of serving reads/writes where the config says it is?

shuzhang1989:
@cpd85 i think the logic is written before the new endpoint is added. feel free to add it

PR #98, opened 2017-07-19
make HTTP layer user friendly

ywang282:
as above


controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/util/Result.java
```diff
@@ -0,0 +1,64 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.pinterest.rocksplicator.controller.util;
+
+/**
+ * A data structure for holding the result from TaskQueue so that
```
angxu: Since TaskQueue doesn't use this class anymore, do we still need it?

ywang282: because previously if we return null, the front end layer receives no content, not even null. But I just read a little bit about HTTP status code, we can use no_content_204 as the status code and therefore we don't need the Result class to bring the null to front end anymore. Thanks!

angxu: Does [WebApplicationException](http://docs.oracle.com/javaee/7/api/javax/ws/rs/WebApplicationException.html#WebApplicationException-java.lang.String-java.lang.Throwable-int-) suffice?

It'll give you not only the error message but also the stack trace at where the error happened.

ywang282: the documentation does not say the String message field is available to clients. you cna check the http://docs.oracle.com/javaee/7/api/javax/ws/rs/WebApplicationException.html#WebApplicationException-javax.ws.rs.core.Response-
public WebApplicationException(Response response), this response that will be returned to the client, which the front end can retrieve.

So basically, for all the resource i found, in order to pass the message to front end, i need to create a Response with the message and then do throw WebApplicationException (Response).  https://stackoverflow.com/questions/7998390/how-do-i-set-a-40x-error-with-a-custom-message-on-jax-rs-exception

Also, somethings the response are not necessarily a WebApplicationException. When thats the case, we still need to put the message in a Response object. 

Let me know if you have a better solution

angxu: I haven't tested it but per the doc(http://www.dropwizard.io/0.8.0-rc1/docs/manual/core.html#error-handling) it says `WebApplicationException` will be mapped to a defined `Response` and throwing `Exception` is preferred than returning `Response`. 

I don't have a strong opinion on it though.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -77,18 +79,27 @@ public Clusters(String zkPath,
...
+      String message = String.format("Failed to read from zookeeper: %s", e);
+      LOG.error(message);
+      return Response.status(HttpStatus.INTERNAL_SERVER_ERROR_500)
+                     .entity(result.setMessage(message))
+                     .build();
     }
     if (clusterBean == null) {
-      throw new WebApplicationException(HttpStatus.NOT_FOUND_404);
+      String message = String.format("Znode %s does not exist or failed to parse config", clusterName);
+      return Response.status(HttpStatus.BAD_REQUEST_400)
```
shuzhang1989: Result class is not needed, because we never have a case need to set message and the data.
for the exception, when it is thrown, it's logged into the server log by the framework. 
i think we actually prefer seeing the stacktrace (or at least the detailed error message) in the frontend, which set the response can give us more control.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/util/Result.java
shuzhang1989: Result class is not needed, because we never have a case need to set message and the data.
for the exception, when it is thrown, it's logged into the server log by the framework. 
i think we actually prefer seeing the stacktrace (or at least the detailed error message) in the frontend, which set the response can give us more control.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
ywang282: `HttpStatus.NOT_FOUND_404` means a server resource does not exist. In this case the server resource exists, just the element we need to find does not exist, therefore `400`.

you can check here: https://stackoverflow.com/questions/25378624/should-a-restful-api-return-400-or-404-when-passed-an-invalid-id

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/util/Result.java
ywang282: The Result class is not needed anymore. But let me see if there is a way front end can get the exception trace.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
shuzhang1989: i think 404 is the one to use. the 404 means the URL doesn't exist. if the host does't exist, then the URL is not valid. 
one analogy is if you want to get a non existing object from S3, s3 will return you code 404.

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -169,8 +196,14 @@ public void replaceHost(@PathParam("clusterName") String clusterName,
           .getEntity();
 
       taskQueue.enqueueTask(task, clusterName, 0);
+      return Response.status(HttpStatus.OK_200)
+                     .entity(ImmutableMap.of("data", true))
+                     .build();
     } catch (JsonProcessingException e) {
-      throw new WebApplicationException(e);
+      String message = String.format("JsonProcessingException: %s", e);
```
shuzhang1989: "cannot serialize parameters"

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -196,8 +230,14 @@ public void loadData(@PathParam("clusterName") String clusterName,
       TaskBase task = new LoadSSTTask(segmentName, s3Bucket, s3Prefix,
           concurrency.orElse(20), rateLimit.orElse(64)).getEntity();
       taskQueue.enqueueTask(task, clusterName, 0);
+      return Response.status(HttpStatus.OK_200)
+                     .entity(ImmutableMap.of("data", true))
+                     .build();
     } catch (JsonProcessingException e) {
-      throw new WebApplicationException(e);
+      String message = String.format("JsonProcessingException: %s", e);
```
shuzhang1989: same here

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -223,8 +274,19 @@ public boolean lock(@PathParam("clusterName") String clusterName) {
    */
   @POST
   @Path("/unlock/{clusterName : [a-zA-Z0-9\\-_]+}")
-  public boolean unlock(@PathParam("clusterName") String clusterName) {
-    return taskQueue.unlockCluster(clusterName);
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response unlock(@PathParam("clusterName") String clusterName) {
+    boolean unlocked = taskQueue.unlockCluster(clusterName);
+    if (unlocked) {
```
shuzhang1989: nit: if(taskQueue.unlockCluster()) {
}

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -16,6 +16,7 @@
 
 package com.pinterest.rocksplicator.controller.resource;
 
+import com.google.common.collect.ImmutableMap;
```
shuzhang1989: move it to second part of imports

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -16,6 +16,7 @@
 
 package com.pinterest.rocksplicator.controller.resource;
 
+import com.google.common.collect.ImmutableMap;
```
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -77,18 +78,26 @@ public Clusters(String zkPath,
...
+  public Response get(@PathParam("clusterName") String clusterName) {
     final ClusterBean clusterBean;
     try {
       clusterBean = checkExistenceAndGetClusterBean(clusterName);
     } catch (Exception e) {
-      LOG.error("Failed to read from zookeeper.", e);
-      throw new WebApplicationException(e);
+      String message = String.format("Failed to read from zookeeper: %s", e);
+      LOG.error(message);
+      return Response.status(HttpStatus.INTERNAL_SERVER_ERROR_500)
```
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -122,9 +135,20 @@ public ClusterBean get(@PathParam("clusterName") String clusterName) {
    */
   @POST
   @Path("/initialize/{clusterName : [a-zA-Z0-9\\-_]+}")
-  public boolean initialize(@PathParam("clusterName") String clusterName) {
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response initialize(@PathParam("clusterName") String clusterName) {
     // Create directly, we dont
-    return taskQueue.createCluster(clusterName);
+    boolean created = taskQueue.createCluster(clusterName);
+    if (created) {
```
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -122,9 +135,20 @@ public ClusterBean get(@PathParam("clusterName") String clusterName) {
...
+      return Response.status(HttpStatus.OK_200)
+                     .entity(ImmutableMap.of("data", true))
+                     .build();
+    }else {
```
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -139,13 +163,16 @@ public boolean initialize(@PathParam("clusterName") String clusterName) {
...
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response replaceHost(@PathParam("clusterName") String clusterName,
                           @NotEmpty @QueryParam("oldHost") String oldHostString,
                           @QueryParam("newHost") Optional<String> newHostOp) {
 
     //TODO(angxu) support adding random new host later
     if (!newHostOp.isPresent()) {
-      throw new UnsupportedOperationException("method not implemented.");
+      return Response.status(HttpStatus.BAD_REQUEST_400)
+                     .entity(ImmutableMap.of("message", "method not implemented"))
```
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
ywang282: checked

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
ywang282: checked

PR #99, opened 2017-07-19
add removeCluster http endpoint

ywang282:
as above

### DIFF
--- a/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
+++ b/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
@@ -141,16 +141,20 @@ public class Clusters {
   }
 
   /**
-   * Remove a given cluster. This may include removing designated tag
-   * in DB and/or writing shard config to zookeeper.
+   * Remove tag in DB and shard config in zookeeper.
    *
    * @param clusterName name of the cluster
    */
   @POST
   @Path("/remove/{clusterName : [a-zA-Z0-9\\-_]+}")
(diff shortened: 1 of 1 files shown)


controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -128,6 +128,19 @@ public boolean initialize(@PathParam("clusterName") String clusterName) {
   }
 
   /**
+   * Remove a given cluster. This may include removing designated tag
+   * in DB and/or writing shard config to zookeeper.
```
shuzhang1989: what does `and/or writing shard config to zookeeper.` mean?

ywang282: I copied and pasted from the `initialize()` 's documentation: 
`/**
   * Initializes a given cluster. This may include adding designated tag
   * in DB and/or writing shard config to zookeeper.
   *
   * @param clusterName name of the cluster
   */`

angxu: In this case, I think we can assume the cluster tag should always exist in DB. So you can just word it like "remove tag in DB and shard config in zookeeper".

ywang282: checked

shuzhang1989:
this operations is a little dangerous, we can hide this api in the ui, and only people in our team with enough knowledge can use it.

PR #103, opened 2017-07-21
Join flush_thread in the destructor of Stats class

angxu:
Otherwise the flush thread may access "this" object after it is destroyed.


newpoo:
Thanks for the quick fix!

PR #104, opened 2017-07-24
Create an abstract task framework for recurring tasks

shuzhang1989:
there will be more recurring tasks to be added. move the shared recurring logic to an abstract class.


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/RetryTask.java
```diff
@@ -93,7 +95,11 @@ public TaskBase getTask() {
       return task;
     }
 
-    public RetryTask.Param setTask(TaskBase task) {
+    public RetryTask.Param setTask(TaskBase task) throws Exception {
+      // We don't allow RetryTask wraps a RecurringTask.
+      if (TaskFactory.getWorkerTask(task) instanceof RecurTask) {
```
angxu: nit: Or you can override the retry() method in RecurTask and throw UnsupportedException there.

shuzhang1989: cool, this is more elegant

PR #108, opened 2017-07-26
add parameter for getAllClusters

ywang282:
as above

### DIFF
--- a/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
+++ b/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
@@ -103,9 +103,9 @@ public class Clusters {
    */
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  public Response getAll(@QueryParam("verbose") boolean verbose) {
+  public Response getAll(@QueryParam("verbose") Optional<Boolean> verbose) {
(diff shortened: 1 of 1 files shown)


controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -103,8 +103,11 @@ public Response get(@PathParam("clusterName") String clusterName) {
    */
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  public Response getAll() {
+  public Response getAll(@QueryParam("verbose") Optional<String> verbose) {
```
angxu: Use boolean if you don't need the value of string?

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -103,8 +103,11 @@ public Response get(@PathParam("clusterName") String clusterName) {
    */
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  public Response getAll() {
+  public Response getAll(@QueryParam("verbose") boolean verbose) {
```
angxu: nit: Optional<Boolean> so you can omit it when not needed.

PR #111, opened 2017-07-31
remove task by id

ywang282:
as above

### DIFF
--- a/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
+++ b/controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
@@ -26,8 +26,8 @@ import org.eclipse.jetty.http.HttpStatus;
 
 import java.util.List;
 import java.util.Optional;
-import javax.ws.rs.GET;
 import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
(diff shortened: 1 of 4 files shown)


controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -82,4 +83,22 @@ public Response findTasks(@QueryParam("clusterName") Optional<String> clusterNam
     return Utils.buildResponse(HttpStatus.OK_200, result);
   }
 
+  /**
+   * Remove a task by its id
+   *
+   * @param id id of the task being removed
+   * @return Response object
+   */
+  @POST
```
angxu: why not use DELETE?

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -82,4 +83,22 @@ public Response findTasks(@QueryParam("clusterName") Optional<String> clusterNam
...
+  @Path("/remove/{id : [0-9]+}")
```
angxu: If you use DELETE, the path can be just "/{id: [0-9]+}".

shuzhang1989: +1

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -27,6 +27,7 @@
 import java.util.List;
 import java.util.Optional;
 import javax.ws.rs.GET;
+import javax.ws.rs.DELETE;
```
shuzhang1989: nit: alphabetical order

PR #114, opened 2017-08-01
Support extra_stats_endpoints to status server.

hrjaco:
This way application can write their own stats endpoint and be appended to the stats.txt endpoint. The new stats will be automatically consumed by the stats collector(there will be no need to write a new stats collector for a new endpoint any more).

Tested on one muse leaf server with this new feature.

### DIFF
--- a/common/stats/status_server.cpp
+++ b/common/stats/status_server.cpp
@@ -29,6 +29,7 @@
 #include <microhttpd.h>
 
 #include <memory>
+#include <set>
 #include <string>
 #include "common/stats/stats.h"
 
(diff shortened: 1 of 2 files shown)


common/stats/status_server.cpp
```diff
@@ -92,6 +98,12 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
       ret += itor->second(nullptr);
     }
```
newpoo: can you remove the code for rocksdb_info.txt special handling by adding "rocksdb_info.txt" to extra_stats_endpoints_?

common/stats/status_server.cpp
```diff
@@ -92,6 +98,12 @@ StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map)
...
 
+    for (const auto& endpoint_name : extra_stats_endpoints_) {
```
newpoo: did you test if ostrich can handle the output like "metric..., counters..., metrics..., counters..., " ?

common/stats/status_server.cpp
hrjaco: Done. Also added a check for "/stats.txt" to prevent infinite recursion.

common/stats/status_server.cpp
hrjaco: Tested and looked at the ostrich-fetcher code. Applications should not provide top keys like "metric" or "counters", they will overwrite the old one in the original "/stats.txt" result. I added comments for that.

common/stats/status_server.cpp
```diff
@@ -70,28 +71,41 @@ int ServeCallback(void* param, struct MHD_Connection* connection,
...
+StatusServer::StatusServer(uint16_t port, EndPointToOPMap op_map,
+                           std::set<std::string> extra_stats_endpoints)
+    : port_(port), d_(nullptr), op_map_(std::move(op_map)),
+      extra_stats_endpoints_(std::move(extra_stats_endpoints)) {
 
+  extra_stats_endpoints_.emplace("/rocksdb_info.txt");
+  // prevent infinite recursion...
+  if (extra_stats_endpoints_.find("/stats.txt") != extra_stats_endpoints_.end()) {
+    extra_stats_endpoints_.erase("/stats.txt");
+  }
```
newpoo: we can remove the "if()" check and always do erase().

common/stats/status_server.h
```diff
@@ -59,11 +60,18 @@ class StatusServer {
    * the function in the form of a vector of [key, value] pairs
    * (arrays can be passed in as delimited values) and returns a
    * std::string.
+   * @param extra_stats_endpoints For endpoints in this list, their string
+   * result will be appended to the regular stats.txt endpoint result. Note:
+   * These endpoints' format need to confirm to the existing stats.txt format,
+   * so that the same stats collector can consume it. Also they cannot use keys
+   * including "counters" and "metrics".
```
newpoo: I think it only allows "counter" type stats? Because the last section of the original "stats.txt" endpoint is for counters. We'd document it also.

common/stats/status_server.cpp
hrjaco: Done

common/stats/status_server.h
hrjaco: I was thinking the new endpoint can output sth like:

PR #116, opened 2017-08-05
Add getHostNumberFor(segment, shard) to ThriftRouter.

hrjaco:
Each Muse root cluster uses only 1 shard, and have all the hosts live in that single shard. 

We currently want to support global rate limit for each root cluster. We need this function to get the root cluster's size to compute each muse root's individual rate limit.

@newpoo @jinruh

### DIFF
--- a/common/thrift_router.h
+++ b/common/thrift_router.h
@@ -189,11 +189,7 @@ class ThriftRouter {
   }
 
   uint32_t getShardNumberFor(const std::string& segment) {
-    std::shared_ptr<const ClusterLayout> layout;
-    {
-      folly::RWSpinLock::ReadHolder read_guard(layout_rwlock_);
-      layout = cluster_layout_;
-    }
+    const auto layout = getClusterLayout();
 
     if (UNLIKELY(layout == nullptr)) {
       return 0;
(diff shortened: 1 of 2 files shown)


common/thrift_router.h
```diff
@@ -207,6 +207,31 @@ class ThriftRouter {
     return itor->second.shard_to_hosts.size();
   }
 
+  // Note: This function is AZ unaware. It returns the total number hosts
+  // within the segment and shard.
```
newpoo: nit: can you follow the similar comment style as other functions in the file?
using:
/*
 * 
 */

common/thrift_router.h
```diff
@@ -207,6 +207,31 @@ class ThriftRouter {
...
+  uint32_t getHostNumberFor(const std::string& segment, const ShardID shard) {
+    std::shared_ptr<const ClusterLayout> layout;
+    {
+      folly::RWSpinLock::ReadHolder read_guard(layout_rwlock_);
+      layout = cluster_layout_;
+    }
```
newpoo: now we have three identical code sections in this file.
`{
      folly::RWSpinLock::ReadHolder read_guard(layout_rwlock_);
       layout = cluster_layout_;
 }`
Can you refactor the code a little bit by adding a private member function getClusterLayout()?

common/thrift_router.h
hrjaco: done

common/thrift_router.h
hrjaco: good point! done

hrjaco:
addressed comments

common/thrift_router.h
```diff
@@ -221,6 +240,16 @@ class ThriftRouter {
     local_client_map_.updateClusterLayout(std::move(layout));
   }
 
+  std::shared_ptr<const ClusterLayout> getClusterLayout() {
```
newpoo: We can simply implement this function like below. It's guaranteed that the lock will be released AFTER reading cluster_layout_.

`
{
folly::RWSpinLock::ReadHolder read_guard(layout_rwlock_);
return cluster_layout_;
}
`

common/thrift_router.h
```diff
@@ -221,6 +240,16 @@ class ThriftRouter {
     local_client_map_.updateClusterLayout(std::move(layout));
```
newpoo: Can you update this function also?

`
{
local_client_map_.updateClusterLayout(getClusterLayout());
}
`

hrjaco: done

common/thrift_router.h
hrjaco: good point! done.

hrjaco:
address comments

jinruh:
Do you just want to get the number of all hosts for certain shard, or the number of all the good hosts for certain shard?

hrjaco:
@jinruh I think for now getting the number of all hosts for certain shard is enough. We just need to get an idea about all the hosts in the cluster, and then rate limit on total_rate_limits / num_servers.

PR #118, opened 2017-08-09
added addDB endpoint to admin handler

cpd85:


shuzhang1989:
can we reuse or make minor change to restoreDB()? what is the use case of adddb api?

cpd85:
When we are restoring a DB, we are restoring it from an HDFS backup. AddDBRequest will be for adding a DB that is not currently in the config for a given host, and might not be coming from an HDFS backup. For having some host serve a shard that comes from S3 and not HDFS, you can use AddDBRequest.

shuzhang1989:
so the use case will be adding a new replica, you want to add the empty db first and upload data  from s3?

angxu:
Can we just change the zookeeper config and restart server?

cpd85:
@shuzhang1989 the case is for adding a new DB. To have a host serve a new shard, we can

1) addDB so the db_name exists in applicationdbmanager
2) download the data from s3, to this new db
3) update the cluster config to reflect this change.

Currently, we cannot do step 2) without adding step 1) or else we will get a db not exist error when we try and download the shards from s3

@angxu If we change the zookeeper config and restart the server, then we'll forward requests to a DB before it is filled with data, so there will be some period of bad requests to the shard

Realpin currently has OpenSegmentDBRequest endpoint for this use case.

cpd85:
Alternatively, I could add an O_CREAT flag to AddS3SstFilesToDBRequest, I just didn't know if this endpoint could have more future use

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -30,6 +30,18 @@ exception AdminException {
   2: required AdminErrorCode errorCode,
 }
 
+struct AddDBRequest {
+  # the db to add. If it is not a master, you must provide upstream info
+  1: required string db_name,
+  2: optional string db_role = "SLAVE",
```
shuzhang1989: should we always add as slave? we can do rebalance after

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -30,6 +30,18 @@ exception AdminException {
...
+  3: optional string upstream_ip,
+  4: optional bool overwrite = false,
```
shuzhang1989: is this flag useful?

rocksdb_admin/rocksdb_admin.thrift
shuzhang1989: also, prefer to use enum instead of string as role

cpd85: Hmmm, ChangeDBRoleAndUpstreamRequest uses a string as a role, do I need to do refactoror should we keep consistency?

shuzhang1989: ok, we should refactor later. 
for the request, is the role field needed?

rocksdb_admin/rocksdb_admin.thrift
newpoo: can you add some comment about what does this flag do?

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -148,6 +160,13 @@ service Admin {
 void ping()
 
 /*
+ * Add the DB to the host, throw exception if it already exists
+ * and overwrite is false.
```
newpoo: I thought the idea is that:
1) if the process already manages the db, we should always throw an exception.
2) otherwise, if overwrite is true, we will remove the DB content on the host, which might be leftover previously.

rocksdb_admin/admin_handler.cpp
```diff
@@ -249,6 +249,73 @@ std::unique_ptr<rocksdb::DB> AdminHandler::removeDB(
   return db;
 }
 
+void AdminHandler::async_tm_addDB(
+      std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
+          AddDBResponse>>> callback,
+      std::unique_ptr<AddDBRequest> request) {
+  AdminException e;
```
newpoo: can you add the following code at the beginning of this function.

`db_admin_lock_.Lock(request->db_name);
SCOPE_EXIT { db_admin_lock_.Unlock(request->db_name); };
`

It helps us to synchronize all DB admin operations.

rocksdb_admin/rocksdb_admin.thrift
cpd85: sure. This way the pattern can be
1) AddDB (I will add in some clearing)
2) Download from s3

Versus
1) Download from s3
2) Catch exception
3) AddDB 
4) Download from s3

rocksdb_admin/rocksdb_admin.thrift
cpd85: Sure, lets make it so that we launch as slave, since the user can use admin handler to change the role manually

rocksdb_admin/admin_handler.cpp
```diff
@@ -249,6 +249,69 @@ std::unique_ptr<rocksdb::DB> AdminHandler::removeDB(
...
 
+void AdminHandler::async_tm_addDB(
+      std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
+          AddDBResponse>>> callback,
+      std::unique_ptr<AddDBRequest> request) {
+  db_admin_lock_.Lock(request->db_name);
+  SCOPE_EXIT { db_admin_lock_.Unlock(request->db_name); };
+
+  AdminException e;
+  // validate the request
```
newpoo: remove this line, since it seems not relevant now.

rocksdb_admin/admin_handler.cpp
```diff
@@ -249,6 +249,69 @@ std::unique_ptr<rocksdb::DB> AdminHandler::removeDB(
...
+  auto segment = admin::DbNameToSegment(request->db_name);
+  auto db_path = FLAGS_rocksdb_dir + request->db_name;
+  auto status = rocksdb::DB::Open(rocksdb_options_(segment), db_path, &rocksdb_db);
+  if (!OKOrSetException(status,
+                        AdminErrorCode::DB_ERROR,
+                        &callback)) {
+    return;
+  }
+
+  if (request->overwrite) {
```
newpoo: should we do this before opening db, i.e., move the if() block to line 277?

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -148,6 +161,13 @@ service Admin {
...
```
newpoo: can you update this comment to reflect the new code?
we always throw an exception if the db already exists.

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -30,6 +30,19 @@ exception AdminException {
   2: required AdminErrorCode errorCode,
 }
 
+struct AddDBRequest {
+  # the db to add. the db is added as slave, so upstream_ip must be provided
+  # if the db exists already, a DB_EXIST error is thrown
+  1: required string db_name,
+  2: required string upstream_ip,
+  # if overwrite is true, then we will clear any existing DB if there is already info
```
newpoo: can we replace "existing" with another word? since we have used it for the case where the process already manages the db. Here, we wanted to mean something else.

rocksdb_admin/admin_handler.cpp
cpd85: Wouldn't you consider adding an existing DB an invalid request? But sure I will remove it

PR #119, opened 2017-08-09
Add cluster namespace support

shuzhang1989:
It is a full stack change, motivation is to divide different clusters into namespace and cluster name. so realpin, rocksdb and scorpion will be 3 namespaces, and cluster name will be the next level in zk path.

Change list:
1. Changed DB schema to add (namespace, name) as the composite primary key of tag table, and (namespace, name) as the composite foreign key of task table.
2. Change all reference to String cluster to Cluster class, which contains namespace and name. Change all API interfaces to refer to Cluster class as input parameter.
3. Changed all HTTP apis which needs to specify cluster to 2 levels: namespace/name. 
4. TaskQueue interfaces are updated to use Cluster class.
5. Unit tests and MySQL integ tests have been changed and passed.

### DIFF
--- a/controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
+++ b/controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
@@ -1,0 +1,77 @@
+/*
+ *  Copyright 2017 Pinterest, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.pinterest.rocksplicator.controller;
+
+/**
+ * Storing cluster metadata (not including detailed segment/shard info)
+ *
+ * @author shu (shu@pinterest.com)
+ */
+public class Cluster {
+  public String namespace;
+  public String name;
+
+  public Cluster(final String namespace, final String name) {
+    this.namespace = namespace;
+    this.name = name;
+  }
+
+
+  public String getNamespace() {
+    return namespace;
+
+  }
+
+  public Cluster setNamespace(String namespace) {
+    this.namespace = namespace;
+    return this;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public Cluster setName(String name) {
+    this.name = name;
+    return this;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+
+    Cluster cluster = (Cluster) o;
+
+    if (namespace != null ? !namespace.equals(cluster.namespace) : cluster.namespace != null)
+      return false;
+    return name != null ? name.equals(cluster.name) : cluster.name == null;
+
+  }
+
+  @Override
+  public int hashCode() {
+    int result = namespace != null ? namespace.hashCode() : 0;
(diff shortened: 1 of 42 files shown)


controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
```diff
@@ -0,0 +1,77 @@
...
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+
+    Cluster cluster = (Cluster) o;
+
+    if (namespace != null ? !namespace.equals(cluster.namespace) : cluster.namespace != null)
+      return false;
+    return name != null ? name.equals(cluster.name) : cluster.name == null;
+
```
ywang282: nit: extra line

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
```diff
@@ -76,9 +77,10 @@ public Response get(@PathParam("id") Long id) {
...
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  public Response findTasks(@QueryParam("clusterName") Optional<String> clusterName,
-                              @QueryParam("state") Optional<TaskState> state) {
-    List<Task> result = taskQueue.peekTasks(clusterName.orElse(null),
+  public Response findTasks(@QueryParam("namespace") Optional<String> namespace,
+                            @QueryParam("clusterName") Optional<String> clusterName,
+                            @QueryParam("state") Optional<TaskState> state) {
+    List<Task> result = taskQueue.peekTasks(new Cluster(namespace.get(), clusterName.get()),
                                             state.map(TaskState::intValue).orElse(null));
```
ywang282: what if namespace and clusterName is not set in the query parameter?

```get(): 
If a value is present in this Optional, returns the value, otherwise throws NoSuchElementException.```

**Comment by shuzhang1989** [comment]
@ywang282 it's work in progress, i will let you know when it's ready to review

File: controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+package com.pinterest.rocksplicator.controller;
+
+/**
+ * Storing cluster metadata (not including detailed segment/shard info)
+ *
+ * @author shu (shu@pinterest.com)
+ */
+public class Cluster {
+  public String namespace;
```
jinruh: private?

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
```diff
@@ -0,0 +1,77 @@
...
+  public String name;
```
jinruh: also can we add some comments on what is namespace and what is name in the cluster?

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public Cluster(final String namespace, final String name) {
+    this.namespace = namespace;
+    this.name = name;
+  }
+
+
+  public String getNamespace() {
+    return namespace;
+
```
shuzhang1989: 👍

controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Tasks.java
shuzhang1989: It should return all tasks.
will add some tests to verify

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
shuzhang1989: 👍

controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/Cluster.java
shuzhang1989: sure

PR #120, opened 2017-08-10
added create_db option to AddS3SstFilesToDB

cpd85:
The reason for this pull request, is so that we can download a shard from S3 even if it doesn't exist in the config. The user who does this can afterwards update their cluster config to reflect that this host is now serving this shard.

### DIFF
--- a/rocksdb_admin/rocksdb_admin.thrift
+++ b/rocksdb_admin/rocksdb_admin.thrift
@@ -114,6 +114,7 @@ struct AddS3SstFilesToDBRequest {
   2: required string s3_bucket,
   3: required string s3_path,
   4: optional i32 s3_download_limit_mb = 64,
+  5: optional bool create_db = false,
 }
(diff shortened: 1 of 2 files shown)


rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -114,6 +114,7 @@ struct AddS3SstFilesToDBRequest {
   2: required string s3_bucket,
   3: required string s3_path,
   4: optional i32 s3_download_limit_mb = 64,
+  5: optional bool create_db = false,
```
newpoo: Can you add an openDB() interface instead?

I'd prefer a set of simple and small interfaces to a smaller set of complicated interfaces.

newpoo:
looks like you already had a PR for addDB() interface, any reason to not use addDB() + addS3SstFilesToDB()?

PR #123, opened 2017-08-15
add healthcheck task for scorpion

shuzhang1989:


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/ConsistentHashRingHealthCheckTask.java
```diff
@@ -0,0 +1,100 @@
...
+    super(parameter);
+  }
+
+  @Override
+  public void process(Context ctx) throws Exception {
+    try {
+      ConsistentHashRingsBean consistentHashRingsBean =
+          ZKUtil.getConsistentHashRingsConfig(zkClient, ctx.getCluster());
+      if (consistentHashRingsBean == null) {
+        ctx.getTaskQueue().failTask(ctx.getId(), "Failed to read cluster config from zookeeper.");
```
angxu: why don't we send email here?

PR #124, opened 2017-08-17
Hdfs backup migration

ywang282:
The test can only pass when the correct environment variables are set. It will fail in the current dock container because of missing environment variables.

Should we exclude this migration test from the Makefile? so when we do `make test`, it won't run with the other tests unless we specifically make it to run?

Please let me know if you have any suggestion on how to perfectly merge this test along with the rest of the tests.

### DIFF
--- a/rocksdb_admin/tests/hdfs_backup_migration_test.cpp
+++ b/rocksdb_admin/tests/hdfs_backup_migration_test.cpp
@@ -1,0 +1,114 @@
+/// Copyright 2016 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author evening (evening@pinterest.com)
+//
+
+#include "boost/filesystem.hpp"
+#include "gtest/gtest.h"
+#include "rocksdb/db.h"
+#include "rocksdb/env.h"
+#include "rocksdb/utilities/backupable_db.h"
+
+using boost::filesystem::remove_all;
+using rocksdb::BackupableDBOptions;
+using rocksdb::BackupEngine;
+using rocksdb::DB;
+using rocksdb::Env;
+using rocksdb::NewHdfsEnv;
+using rocksdb::Options;
+using rocksdb::ReadOptions;
+using rocksdb::SequenceNumber;
+using std::string;
+using std::to_string;
+using std::unique_ptr;
+
+// helper function to open a rocksdb instance
+unique_ptr<DB> OpenDB(const string& path, bool error_if_exists = false) {
+  Options options;
+  options.create_if_missing = true;
+  options.error_if_exists = error_if_exists;
+  options.WAL_size_limit_MB = 100; //
+  DB* db;
+  auto status = DB::Open(options, path, &db);
+  EXPECT_TRUE(status.ok());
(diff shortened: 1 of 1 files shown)


angxu:
what are the env variables needed?

rocksdb_admin/tests/hdfs_backup_migration_test.cpp
```diff
@@ -0,0 +1,114 @@
+/// Copyright 2016 Pinterest Inc.
```
shuzhang1989: rename it to "hbase_backup_integration_test"

rocksdb_admin/tests/hdfs_backup_migration_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  options.error_if_exists = error_if_exists;
+  options.WAL_size_limit_MB = 100;
+  DB* db;
+  auto status = DB::Open(options, path, &db);
+  EXPECT_TRUE(status.ok());
+  return unique_ptr<DB>(db);
+}
+
+unique_ptr<DB> CleanAndOpenDB(const string& path) {
+  EXPECT_NO_THROW(remove_all(path));
```
shuzhang1989: let's use destroydb() api

ywang282:
@angxu the env variables are for example JAVA_HOME, CLASSPATH to all the hadoop distributions and c++ dynamically linked libs, etc

ywang282:
@shuzhang1989 yes the new hdfs cluster works fine, you can find all the back used files with `hdfs dfs -ls /migration_test/private`

rocksdb_admin/tests/hdfs_backup_migration_test.cpp
ywang282: checked

rocksdb_admin/tests/hdfs_backup_migration_test.cpp
ywang282: checked

rocksdb_admin/tests/hdfs_backup_migration_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  return OpenDB(path, true);
+}
+
+TEST(HDFSBackupMigrationTest, Basics) {
+
+  string hdfs_name_node = "hdfs://hbasebak-infra-namenode-prod1c01-001:8020";
```
ywang282: checked

rocksdb_admin/tests/hbase_backup_integration_test.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  status = engine->CreateNewBackup(original_db.get());
+  EXPECT_TRUE(status.ok());
+
+  // restore db
+  status = engine->RestoreDBFromLatestBackup(FLAGS_restored_db_path, FLAGS_restored_db_path);
+  EXPECT_TRUE(status.ok());
+
+  // compare the original db with the restored db
+  auto restored_db = OpenDB(FLAGS_restored_db_path);
```
shuzhang1989: why not clean and open db for the restore db?

rocksdb_admin/tests/hbase_backup_integration_test.cpp
```diff
@@ -0,0 +1,123 @@
+/// Copyright 2016 Pinterest Inc.
```
shuzhang1989: make it a manual test and no run by cmake test

rocksdb_admin/tests/hbase_backup_integration_test.cpp
ywang282: Because RestoreDBFromLatestBackup() will create the restored db with restored data. If we clean_and_open_db it will erase all the restored data so the db will be empty. Thus OpenDB() .

rocksdb_admin/tests/hbase_backup_integration_test.cpp
ywang282: screen shot 2017-08-17 at 2 14 43 pm
screen shot 2017-08-17 at 2 20 24 pm

shuzhang1989:
thanks for the integration test!

PR #133, opened 2017-08-29
For recurring job, pass down job body to next enqueued jobs

shuzhang1989:


controller/controller-worker/src/main/java/com/pinterest/rocksplicator/controller/tasks/RecurTask.java
```diff
@@ -118,6 +118,9 @@ public void process(Context ctx) throws Exception {
     try {
       ctx = new Context(id, ctx.getCluster(), localAckTaskQueue, ctx.getWorker());
       task.process(ctx);
+      // in case the current body is changed, and we need to carry the state
+      // to next enqueued job
+      getParameter().getTask().body = task.getParameter().serialize();
```
angxu: can we do getParameter().setTask(task) ?

shuzhang1989: use setBody() instead because Task's other 2 fields are name and priority

PR #137, opened 2017-09-19
Implement MySQL based cluster manager

shuzhang1989:


controller/controller-common/src/main/java/com/pinterest/rocksplicator/controller/mysql/MySQLClusterManager.java
```diff
@@ -16,45 +16,154 @@
...
+
+  private boolean persistTagHostsEntity(TagHostsEntity tagHostsEntity) {
+    beginTransaction();
+    getEntityManager().persist(tagHostsEntity);
+    getEntityManager().getTransaction().commit();
+    return true;
+  }
+
+  @Override
+  public boolean createCluster(final Cluster cluster) {
```
angxu: is this operation protected by cluster lock? if not, i think we need to beginTransaction() from the beginning.

shuzhang1989: good catch.
for other operations like register / unregister and blacklist, i plan to implement them in tasks, and task will automatically hold the cluster lock so they should be fine.

PR #139, opened 2017-09-19
[rocksplicator/timer] remove unused function

newpoo:


newpoo:
@shuzhang1989 I vaguely remember you used to write a function which inherits from Timer class. However, I couldn't find it now. Does it use the Timer::now() function anywhere?

PR #141, opened 2017-09-20
Migrate AddFile() call to IngestExternalFile()

cpd85:
https://github.com/facebook/rocksdb/blob/60beefd6e043967165af2d2d8944fefa580a2383/include/rocksdb/db.h#L1017


rocksdb_replicator/tests/rocksdb_assumption_test.cpp
```diff
@@ -222,11 +222,16 @@ TEST(RocksDBAssumptionTest, SequenceNumber) {
 
 
   // add two SST files to the empty DB, the sequence # should still be zero
```
newpoo: Can we add a test to verify what happens to sequence # after ingesting files into a non-empty DB? Our cluster management script/task may need to be modified accordingly.

PR #149, opened 2017-09-29
Enhance stats support periodic callbacks

shuzhang1989:
Use case: 
ThriftServer has getLoad() to return active tasks, to expose it, we need to periodically call Stats->AddMetric() for this function. 

Compiled a server and run in production, worked as expected

### DIFF
--- a/common/stats/stats.cpp
+++ b/common/stats/stats.cpp
@@ -370,14 +370,14 @@ void Stats::AddMetric(const string& metric, int64_t value) {
 void Stats::RegisterIncr(const std::string &counter,
                          std::function<uint64_t()> callback) {
   lock_guard<mutex> g(lock_counter_callbacks_);
-  counter_callbacks_.emplace(counter, callback);
+  counter_callbacks_.emplace(counter, std::move(callback));
 }
 
 
 void Stats::RegisterAddMetric(const std::string &metric,
                               std::function<int64_t()> callback) {
   lock_guard<mutex> g(lock_metrics_callbacks_);
-  metrics_callbacks_.emplace(metric, callback);
+  metrics_callbacks_.emplace(metric, std::move(callback));
 }
 
 LocalStats* Stats::GetLocalStats() {
--- a/common/stats/stats.cpp
+++ b/common/stats/stats.cpp
@@ -324,6 +324,18 @@ Stats::Stats() : flush_interval_(kFlushIntervalMS)
   flush_thread_ = thread([this] {
     while (!should_stop_) {
       sleep_for(this->flush_interval_);
+      {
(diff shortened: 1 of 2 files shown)


newpoo:
Does it make more sense to make a dedicated class to accept callback registration and call stats::incr() and stats::addMetric periodically?
There are two benefits. 1) it doesn't add more complexities to Stats class, which is already very complicated; 2)

common/stats/stats.cpp
```diff
@@ -355,6 +367,19 @@ void Stats::AddMetric(const string& metric, int64_t value) {
   GetLocalStats()->AddMetric(metric, value);
 }
 
+void Stats::RegisterIncr(const std::string &counter,
+                         std::function<uint64_t()> callback) {
+  lock_guard<mutex> g(lock_counter_callbacks_);
+  counter_callbacks_.emplace(counter, callback);
```
newpoo: std::move(callback)

common/stats/stats.cpp
```diff
@@ -355,6 +367,19 @@ void Stats::AddMetric(const string& metric, int64_t value) {
...
+}
+
+
+void Stats::RegisterAddMetric(const std::string &metric,
+                              std::function<int64_t()> callback) {
+  lock_guard<mutex> g(lock_metrics_callbacks_);
+  metrics_callbacks_.emplace(metric, callback);
```
newpoo: same.

common/stats/stats.h
```diff
@@ -83,6 +88,11 @@ class Stats {
   void AddMetric(const uint32_t metric, int64_t value);
   void AddMetric(const std::string& metric, int64_t value);
 
+  // Register callback for Counter / Metric so they will be reported
+  // periodically by the Stats class.
+  void RegisterIncr(const std::string& counter, std::function<uint64_t()>);
+  void RegisterAddMetric(const std::string& metric, std::function<int64_t()>);
```
newpoo: we may provide a uint32_t version counter and metric APIs if we need to use it now. Otherwise, we may add it later.

newpoo:
Do you need unregister API during gracefully shutdown?

shuzhang1989:
yes, my first thought is also to do it in a class, but it has some disadvantages:
1. I need to implement the same polling thread logic to periodically report stats
2. i need to register the new class into shutdown hook. for this implementation, because status_server is called Stop() at the end of main already, no need to work with shutdown hook.

common/stats/stats.h
shuzhang1989: the stats function is uint64_t() for ThriftServer

shuzhang1989:
actually it has nothing to do with stats server, the stats() object and polling thread will be safely joined() at the deconstructor of the stats singleton

PR #150, opened 2017-09-29
[admin_handler] create gflag for allowing global seqno

cpd85:


cpd85:
the default case (false) is more common

rocksdb_admin/admin_handler.cpp
```diff
@@ -53,6 +53,9 @@ DEFINE_int32(port, 9090, "Port of the server");
 DEFINE_string(shard_config_path, "",
              "Local path of file storing shard mapping for Aperture");
 
+DEFINE_bool(allow_global_seqno, false,
```
newpoo: nit: rename it to rocksdb_allow_global_seqno, and see if we can put them on the same line?

rocksdb_admin/admin_handler.cpp
```diff
@@ -722,7 +725,7 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   rocksdb::IngestExternalFileOptions ifo;
   ifo.move_files = true;
   /* allow for overlapping keys */
```
newpoo: remove this comment now?

cpd85: isn't it useful to say what FLAGS_ rocksdb_allow_global_seqno does?

newpoo:
sure, I was suggesting that the comment is inaccurate now because allowing overlapping keys is now determined by a gflag.

PR #151, opened 2017-10-02
Update seqno flag

cpd85:


rocksdb_admin/admin_handler.cpp
```diff
@@ -725,6 +725,7 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   ifo.move_files = true;
   /* if true, rocksdb will allow for overlapping keys */
   ifo.allow_global_seqno = FLAGS_rocksdb_allow_global_seqno;
+  ifo.allow_blocking_flush = FLAGS_rocksdb_allow_global_seqno;
```
newpoo: should we rename the gflag to something like rocksdb_allow_overlapping_keys?

cpd85: sure, we need to be extra careful with rolling out deploys on clusters when we land the change then

PR #161, opened 2017-10-12
Replace AZ concept in thrift router to Replica

shuzhang1989:
To enable multiple replica in one az, remove the AZ dependency in thrift router and generalize it to be replica name.


shuzhang1989:
will have another diff.

PR #162, opened 2017-10-14
Support multiple groups in one AZ for routing (good version)

shuzhang1989:
The change should cover 3 parts:
1. Foreign host (not in the config) to route using AZ information
2. host in config to talk to other hosts in config, but the config doesn't have group concept. The routing is based on AZ information, only route traffic to local az.
3. host in config to talk to other hosts in config, and the config contains group concept. The routing is based on group information, only route traffic to local group.

1-3 is the behavior when FLAGS_always_prefer_local_host is true.

### DIFF
--- a/common/tests/thrift_router_test.cpp
+++ b/common/tests/thrift_router_test.cpp
@@ -799,11 +799,11 @@ TEST(ThriftRouterTest, ForeignHostGroupTestPreferLocal) {
     EXPECT_EQ(v.size(), 1);
     v[0]->future_ping().get();
   }
-  // Load should be balanced
-  ASSERT_TRUE(handlers[0]->nPings_.load() == 51);
-  ASSERT_TRUE(handlers[1]->nPings_.load() == 51);
+  ASSERT_TRUE(handlers[0]->nPings_.load() > 1);
+  ASSERT_TRUE(handlers[1]->nPings_.load() > 1);
   ASSERT_TRUE(handlers[2]->nPings_.load() == 1);
   ASSERT_TRUE(handlers[3]->nPings_.load() == 1);
+  ASSERT_TRUE(handlers[0]->nPings_.load() + handlers[1]->nPings_.load() == 102);
 
   // stop all servers
   for (auto& s : servers) {
--- a/common/tests/thrift_router_test.cpp
+++ b/common/tests/thrift_router_test.cpp
@@ -23,11 +23,9 @@
 #include <tuple>
 #include <utility>
 #include <vector>
-#include <boost/algorithm/string/replace.hpp>
 
 #include "gtest/gtest.h"
 
-#include "common/network_util.h"
 #include "common/tests/thrift/gen-cpp2/DummyService.h"
 #include "common/thrift_router.h"
 #include "thrift/lib/cpp2/server/ThriftServer.h"
@@ -88,14 +86,14 @@ static const char* g_config_v2 =
   "{"
   "  \"user_pins\": {"
   "  \"num_leaf_segments\": 3,"
-  "  \"127.0.0.1:8090:zone-a\": [\"00002:M\", \"00000:M\"],"
-  "  \"127.0.0.1:8091:zone-b_1\": [\"00000:S\", \"00001:M\", \"00002:M\"],"
-  "  \"127.0.0.1:8092:zone-c\": [\"00001:S\", \"00002:M\"]"
+  "  \"127.0.0.1:8090:zone_a\": [\"00002:M\", \"00000:M\"],"
+  "  \"127.0.0.1:8091:test\": [\"00000:S\", \"00001:M\", \"00002:M\"],"
+  "  \"127.0.0.1:8092:zone_c\": [\"00001:S\", \"00002:M\"]"
   "   },"
   "  \"interest_pins\": {"
   "  \"num_leaf_segments\": 2,"
-  "  \"127.0.0.1:8090:zone-a\": [\"00000:M\"],"
-  "  \"127.0.0.1:8091:zone-b_0\": [\"00001:S\"]"
+  "  \"127.0.0.1:8090:zone_a\": [\"00000:M\"],"
+  "  \"127.0.0.1:8091:test\": [\"00001:S\"]"
   "   }"
   "}";
 
@@ -158,18 +156,15 @@ makeServer(uint16_t port) {
 }
 
 void updateConfigFile(const string& content) {
-  // replace the IP with real local IP.
-  string modify_content = content;
-  boost::replace_all(modify_content, "127.0.0.1", common::getLocalIPAddress());
   ofstream os(g_config_path);
   EXPECT_TRUE(os);
-  os << modify_content;
+  os << content;
 }
 
 TEST(ThriftRouterTest, Basics) {
   updateConfigFile("");
   ThriftRouter<DummyServiceAsyncClient> router(
-    "zone-b", g_config_path, common::parseConfig);
+    "test", g_config_path, common::parseConfig);
 
   EXPECT_EQ(router.getShardNumberFor("user_pins"), 0);
   EXPECT_EQ(router.getShardNumberFor("interest_pins"), 0);
@@ -368,7 +363,6 @@ TEST(ThriftRouterTest, Basics) {
   }
 }
 
-
 void stress(int n_threads, int n_ops) {
   vector<thread> threads(n_threads);
   shared_ptr<DummyServiceTestHandler> handlers[3];
@@ -380,7 +374,7 @@ void stress(int n_threads, int n_ops) {
 
   updateConfigFile(g_config_v2);
   ThriftRouter<DummyServiceAsyncClient> router(
-    "zone-b", g_config_path, common::parseConfig);
+    "", g_config_path, common::parseConfig);
   sleep(1);
 
   for (int i = 0; i < n_threads; ++i) {
@@ -454,7 +448,7 @@ TEST(ThriftRouterTest, LocalAzTest) {
   tie(handlers[1], servers[1], thrs[1]) = makeServer(8091);
   tie(handlers[2], servers[2], thrs[2]) = makeServer(8092);
   sleep(1);
-  
+
   EXPECT_EQ(router.getShardNumberFor("user_pins"), 3);
   EXPECT_EQ(router.getHostNumberFor("user_pins", 0), 3);
   EXPECT_EQ(router.getHostNumberFor("user_pins", 1), 3);
@@ -476,7 +470,7 @@ TEST(ThriftRouterTest, LocalAzTest) {
   for (int i = 0; i < 100; i ++) {
     std::vector<shared_ptr<DummyServiceAsyncClient>> v;
     EXPECT_EQ(
-        router.getClientsFor("user_pins", Role::ANY, Quantity::ONE, 2, &v), 
+        router.getClientsFor("user_pins", Role::ANY, Quantity::ONE, 2, &v),
         ReturnCode::OK);
     EXPECT_EQ(v.size(), 1);
     v[0]->future_ping().get();
@@ -507,7 +501,7 @@ TEST(ThriftRouterTest, ForeignAzTest) {
   tie(handlers[1], servers[1], thrs[1]) = makeServer(8091);
   tie(handlers[2], servers[2], thrs[2]) = makeServer(8092);
   sleep(1);
-  
+
   EXPECT_EQ(router.getShardNumberFor("user_pins"), 3);
   EXPECT_EQ(router.getHostNumberFor("user_pins", 0), 3);
   EXPECT_EQ(router.getHostNumberFor("user_pins", 1), 3);
@@ -527,7 +521,7 @@ TEST(ThriftRouterTest, ForeignAzTest) {
   for (int i = 0; i < 100; i ++) {
     std::vector<shared_ptr<DummyServiceAsyncClient>> v;
     EXPECT_EQ(
-        router.getClientsFor("user_pins", Role::ANY, Quantity::ONE, 2, &v), 
+        router.getClientsFor("user_pins", Role::ANY, Quantity::ONE, 2, &v),
         ReturnCode::OK);
     EXPECT_EQ(v.size(), 1);
     v[0]->future_ping().get();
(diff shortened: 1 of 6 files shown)


shuzhang1989:
that is the way similar to the previous diff. The problem with this is that groups are not az - groups are segment dependent - because when we config a new segment, we ask people to pass in how many groups they want. That's why the group info is stored in SegmentInfo, not the host level.

shuzhang1989:
ok. 
for passing in groups in the constructor, the client need to pass in a map then?

newpoo:
Client only passes in a single group.

shuzhang1989:
thrift router contains all segment information from the config, but why in the constructor it only take in one group name? the caller should give all groups it is in.

For example, in ApertureRouter, we only maintain one ThriftRouter<ApertureAsyncClient>, but it is responsible for routing traffic to multiple segments

newpoo:
The problem is how do we determine the map for external config with multiple segments? I was thinking to pass in one group name as the default routing prefix for all segments. This should be able to cover all of our current use cases, right? We allow a host resides in two different group names within a single config. But we don't allow multiple routing prefix in a single thrift router.

shuzhang1989:
Yes, if we pass in a map of segment-group, it is very hard to let the client know and pass in that.

I think what you are saying the group name of default prefix, is just what we do today as the az. for example, a host is in group us-east-1a_0 for segment1, us-east-1a_1 for segment2, then the passed in group is us-east-1a which is the AZ. 

That's why i keep the AZ concept in the implementation, and if external client want to use another AZ, they can also pass in. The default prefix you talked about I think is more confusing right?

newpoo:
In the new code, az is completely useless. The current way using it is way hacky, and makes strong implicit assumptions. We don't want to make the decision for clients on what group name to use for routing. The case you described (two segments need sub_group routing) is a rare use case, even pinnability doesn't necessary need it. We don't want to make majority clients' life difficult.

For a cluster with two segments. One use az level routing, one use sub_group level routing, we can always pass in the sub_group name as the default routing prefix. It will automatically cover both cases. Because we sort hosts based on common prefix length.

We can easily extend the code to support two segments for sub_group LATER routing by adding a map<segment, routing_prefix> thrift_router constructor parameter if we have to do that. Even in this case, we still want to keep the default routing prefix parameter. Because majority of clients don't have to know all these things which are unrelated to their use cases.

shuzhang1989:
ok. if we have to do that way, i can change that and send another diff. 

I think even passing in the group is an exposure of cluster layout internal info to client. but passing in an AZ is reasonable for client. Imagine if you are scorpion, you know what AZ you are in, and ThriftRouter will pick up best Client for you in GetClientFor() calls. I cannot come up another use case when client need to pretend itself in another zone.

But in the other hand, not all application of rocksplicator has the zone concept, but as long as they are in cloud, zone is a reasonable implication.

Do you have strong opionion your suggested way is better? if yes, i will change to that way

newpoo:
Nothing is blocking clients passing in AZ as the group name with the suggested solution. That is actually how the majority of our current use cases will work with this change.

shuzhang1989:
yes I know functionally the interface is the same, but the difference is the concept changes from az to a prefix string. If i am a client. and i dont know the config, how do i know I should pass in AZ as the prefix? then it means you still need to know the config layout and convention to know passing in AZ is the correct way to constrcut a thrift router

shuzhang1989:
@newpoo the diff has been updated with the prefix length based routing

common/thrift_router.h
```diff
@@ -385,34 +383,47 @@ class ThriftRouter {
...
-          return h1->az == this->local_az_;
-        } else {
-          return h1->az < h2->az;
-        }
+        const unsigned rotation_counter,
+        const std::string& segment) {
+      if (v->size() == 0) return;
+      auto comparator = [this, segment]
+        (const Host* h1, const Host* h2) -> bool {
+        // Descending order
```
newpoo: How about we do something like this, then we can remove all the code after line 402 in this function.

if (h1->groups_prefix_lengths[segment] != h2->groups_prefix_lengths[segment]) {
  return *** > ***;
}

return (rotation_counter & 0x1) ? h1 < h2 : h2 < h1; // this is probably good enough, if it turns out to be an issue, we can add pid info to it.

common/thrift_router.cpp
```diff
@@ -105,26 +121,34 @@ std::unique_ptr<const detail::ClusterLayout> parseConfig(std::string content) {
...
-      if (!parseHost(host_port_az, &host)) {
-        LOG(ERROR) << "Invalid host port az " << host_port_az;
+      if (!parseHost(host_port_group, &host, segment, local_group)) {
+        LOG(ERROR) << "Invalid host port group " << host_port_group;
         return nullptr;
       }
 
-      const detail::Host* pHost = &*(cl->all_hosts.insert(host).first);
-      const auto& shard_list = segment_value[host_port_az];
+      if (cl->all_hosts.find(host.addr) == cl->all_hosts.end()) {
```
newpoo: we can do cl->all_hosts.find(host) instead of find(host.addr). So we don't have to change the map to set above.

common/thrift_router.cpp
```diff
@@ -30,10 +31,22 @@ DEFINE_int32(min_client_reconnect_interval_seconds, 5,
...
+inline uint16_t longest_common_prefix(const std::string& s1,
+                                      const std::string& s2) {
+  auto length = std::min(s1.length(), s2.length());
+  uint16_t count = 0;
+  for (int i = 0; i < length; i ++, count ++) {
+    if (s1[i] != s2[i]) {
+      return count;
+    }
+  }
+  return count;
```
newpoo: nit: can we use std::mismatch() to implement this function?

common/thrift_router.cpp
```diff
@@ -45,8 +58,10 @@ bool parseHost(const std::string& str, common::detail::Host* host) {
   } catch (...) {
     return false;
   }
-
-  host->az = (tokens.size() == 3 ? tokens[2] : "unknown");
+  auto group = (tokens.size() == 3 ? tokens[2] : "unknown");
```
newpoo: This is not introduced by your change. Does it make sense to replace "unknown" with ""?

common/thrift_router.cpp
```diff
@@ -45,8 +58,10 @@ bool parseHost(const std::string& str, common::detail::Host* host) {
...
+  host->groups_prefix_lengths[segment] =
+    longest_common_prefix(group, local_group);
+  tokens.clear();
```
newpoo: remove this line. It's useless.

common/thrift_router.cpp
```diff
@@ -75,7 +90,8 @@ bool parseShard(const std::string& str, common::detail::Role* role,
 
 namespace common {
 
-std::unique_ptr<const detail::ClusterLayout> parseConfig(std::string content) {
+std::unique_ptr<const detail::ClusterLayout> parseConfig(
+  std::string content, const std::string& local_group) {
```
newpoo: too more spaces.

common/thrift_router.cpp
```diff
@@ -105,26 +121,34 @@ std::unique_ptr<const detail::ClusterLayout> parseConfig(std::string content) {
...
-      const auto& shard_list = segment_value[host_port_az];
+      if (cl->all_hosts.find(host.addr) == cl->all_hosts.end()) {
+        cl->all_hosts.emplace(host.addr, host);
+      } else {
+        cl->all_hosts[host.addr].groups_prefix_lengths.insert(
+                host.groups_prefix_lengths.begin(),
+                host.groups_prefix_lengths.end());
+      }
+
+      detail::Host* pHost = &(cl->all_hosts.at(host.addr));
```
newpoo: replace at() with [], we are sure we have it.

common/tests/thrift_router_test.cpp
```diff
@@ -23,9 +23,11 @@
 #include <tuple>
```
newpoo: I think most of the changes to existing tests can be undone, right?
Because we now rely on client passing in local_group for routing.
We just need to add a new test for the new use case.

common/thrift_router.cpp
shuzhang1989: the reason is that i need to update the element in the set in place, the set::find() returns const iterator.

shuzhang1989: Here is what i tried:

set<Host>::iterator it = cl->all_hosts.find(host);
it-> groups_prefix_lengths.insert(host. groups_prefix_lengths.begin(), host. groups_prefix_lengths.end())

gcc returns error that cannot manipulate a const iterator

common/thrift_router.h
```diff
@@ -385,34 +383,47 @@ class ThriftRouter {
...
-      auto comparator = [this] (const Host* h1, const Host* h2) -> bool {
-        if (h1->az != h2->az &&
-            (h1->az == this->local_az_ || h2->az == this->local_az_)) {
-          return h1->az == this->local_az_;
-        } else {
-          return h1->az < h2->az;
-        }
+        const unsigned rotation_counter,
+        const std::string& segment) {
+      if (v->size() == 0) return;
```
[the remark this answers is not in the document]
shuzhang1989: if no this line, a test will cause segfault (config as "" content)

common/thrift_router.h
shuzhang1989: not sure how does it work?

common/thrift_router.h
```diff
@@ -363,10 +360,11 @@ class ThriftRouter {
     auto filterByRoleAndSortByPreference(
         const std::vector<std::pair<const Host*, Role>>& host_info,
         const Role role,
-        const unsigned rotation_counter) {
+        const unsigned rotation_counter,
+        const std::string& segment) {
       std::vector<const Host*> v;
       if (role == Role::ANY && !FLAGS_always_prefer_local_host) {
-        // prefer MASTER, then prefer local
+        // prefer MASTER, then prefer same group, then prefer same AZ
```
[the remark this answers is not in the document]
shuzhang1989: are you saying remove the FLAGS_always_prefer_local_host flag and combine the 2 branches into a simple common prefix length logic

common/thrift_router.cpp
newpoo: Can we get around it by inserting existing lengths into "host", and replace the current one with "host" by "erase()" and "insert()".

common/thrift_router.h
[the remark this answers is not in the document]
newpoo: I meant "prefer MASTER, then prefer groups with longer common prefix length".

common/thrift_router.h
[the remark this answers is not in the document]
newpoo: That sounds like a bug?

common/thrift_router.h
newpoo: What we need to do is to sort hosts by their common prefix length. If some hosts are of the same length, we want to sort them in a different order for each function call.

The way I suggested will sort hosts of the same length based on their address either in descending or ascending order. This will work unless we have more than 2 hosts of the same length AND the hosts' addresses are in the same order in every process in a cluster. If this does happen, we can incorporate pid info into the sorting criteria. But we don't bother to do it now.

common/thrift_router.cpp
shuzhang1989: sounds good

common/thrift_router.h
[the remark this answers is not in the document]
shuzhang1989: oh good

common/thrift_router.h
shuzhang1989: this implementation has a problem. For example we have 3 hosts both have the same length, and we call getClientsFor() to get ONE host. Because the rotation_counter only has 2 states (&0x1), we will only end up getting 2 orders: either 123 or 321. 

Then the second host will never get called, which made the load not balanced.

newpoo: Did you really read my last comment? I feel the communication is not effective.

shuzhang1989: yeah, i did. here is the comparator i wrote:

auto comparator = [this, &segment, &rotation_counter]
        (const Host* h1, const Host* h2) -> bool {
        // Descending order
        auto s1 = h1->groups_prefix_lengths.at(segment);
        auto s2 = h2->groups_prefix_lengths.at(segment);
        if (s1 == s2) {
          return (rotation_counter & 0x1) ? h1 > h2 : h2 > h1;
        } else {
          return s1 > s2;
        }
      };

common/thrift_router.cpp
```diff
@@ -105,26 +109,33 @@ std::unique_ptr<const detail::ClusterLayout> parseConfig(std::string content) {
...
-      const detail::Host* pHost = &*(cl->all_hosts.insert(host).first);
-      const auto& shard_list = segment_value[host_port_az];
+      auto host_iter = cl->all_hosts.find(host);
+      if (host_iter != cl->all_hosts.end()) {
+        host.groups_prefix_lengths.insert(
+                host_iter->groups_prefix_lengths.begin(),
+                host_iter->groups_prefix_lengths.end());
+        cl->all_hosts.erase(host_iter);
+      }
+        cl->all_hosts.insert(std::move(host));
```
newpoo: remove two spaces.

common/thrift_router.h
```diff
@@ -385,34 +384,31 @@ class ThriftRouter {
...
+    void RankHostsByGroupPrefixLength(
         std::vector<const Host*>* v,
-        const unsigned rotation_counter) {
-      auto comparator = [this] (const Host* h1, const Host* h2) -> bool {
-        if (h1->az != h2->az &&
-            (h1->az == this->local_az_ || h2->az == this->local_az_)) {
-          return h1->az == this->local_az_;
+        const unsigned rotation_counter,
+        const std::string& segment) {
+      auto comparator = [this, &segment, &rotation_counter]
```
newpoo: remove "&" before rotation_counter. it's more efficient to pass an unsigned variable by copy.

common/thrift_router.h
```diff
@@ -385,34 +384,31 @@ class ThriftRouter {
...
+        (const Host* h1, const Host* h2) -> bool {
+        // Descending order
+        auto s1 = h1->groups_prefix_lengths.at(segment);
+        auto s2 = h2->groups_prefix_lengths.at(segment);
```
newpoo: nit: const auto s1 and const auto s2.

and const auto hash1 and const auto hash2 below.

PR #174, opened 2017-11-14
add queryparam to hosts endpoint

ywang282:
as title


controller/controller-http/src/main/java/com/pinterest/rocksplicator/controller/resource/Clusters.java
```diff
@@ -510,8 +510,27 @@ public Response registerHost(@PathParam("namespace") String namespace,
...
+      final ClusterBean clusterBeanInConfig;
+      try {
+        clusterBeanInConfig = checkExistenceAndGetClusterBean(namespace, clusterName);
+      } catch (Exception e) {
+        String message = String.format("Failed to read from zookeeper: %s", e);
+        LOG.error(message);
+        return Utils.buildResponse(HttpStatus.INTERNAL_SERVER_ERROR_500,
+                ImmutableMap.of("message", message));
+      }
+      if (!hosts.removeAll(clusterBeanInConfig.getHosts())) {
```
shuzhang1989: if cannot, we should not make it an error, it just mean there is no idle hosts. return an empty set.

shuzhang1989:
great!

PR #181, opened 2017-12-05
Task order

ywang282:
This will fix the task order bug

### DIFF
--- a/controller/controller-board/src/app/controllers/TaskController.js
+++ b/controller/controller-board/src/app/controllers/TaskController.js
@@ -23,7 +23,6 @@
         vm.showDetail = [];
         vm.tasks = [];
         var BAD_REQUEST = 400;
-
         vm.getDate = function(data){
             return new Date(data).toString().split(' ').splice(1,4).join(' ');
         };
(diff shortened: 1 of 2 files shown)


controller/controller-board/src/app/controllers/TaskController.js
```diff
@@ -86,7 +86,7 @@
             taskService.getTasks(vm.namespace, vm.clustername, vm.state)
                 .then(function(result) {
                     vm.taskStatusCode = result.status;
-                    vm.tasks = result.data;
+                    vm.tasks = result.data.reverse();
```
shuzhang1989: a simple reverse is good enough?

ywang282: yes, I noticed that the task ids are indeed incremented when there is a new task coming in and the bug is actually introduced by the $index from angular. So I removed them. 

However I don't know what will happen when we exhaust all the id numbers, so we should still use the actual date to sort, however, this makes the rendering time longer.

shuzhang1989: good point using dates. 
the id probably won't overflow anytime soon, but let's use dates.

PR #185, opened 2017-12-13
small fix

ywang282:
change includes:

1. connect task in cluster to task tab
2. replace ID with "created_at" timestamp
3. show cluster in default page


shuzhang1989:
great job!

PR #212, opened 2018-02-24
Add a counter in S3Util to track number of instances. 

weipihui:
Add a counter in S3Util to track number of instances. Aws::ShutdownAPI() will be called only when there is no other instences left.

### DIFF
--- a/common/s3util.h
+++ b/common/s3util.h
@@ -28,6 +28,7 @@
 #include <aws/core/utils/Outcome.h>
 #include <boost/iostreams/categories.hpp>
 
+#include <atomic>
 #include <iosfwd>
 #include <map>
 #include <string>
@@ -175,6 +176,7 @@ class S3Util {
                   const uint32_t read_ratelimit_mb) :
     bucket_(std::move(bucket)), s3Client(client_config), options_(options),
     read_ratelimit_mb_(read_ratelimit_mb) {
+    ++counter_;
     Aws::StringStream ss;
     ss << Aws::Http::SchemeMapper::ToString(client_config.scheme) << "://";
 
@@ -187,7 +189,10 @@ class S3Util {
   }
 
   ~S3Util() {
-    Aws::ShutdownAPI(options_);
+    --counter_;
+    if (counter_.load() == 0) {
+      Aws::ShutdownAPI(options_);
+    }
   }
 
   // Download an S3 Object to a local file
@@ -245,6 +250,11 @@ class S3Util {
     return read_ratelimit_mb_;
   }
 
+  // For test only.
+  static uint32_t getInstanceCounter() {
+    return counter_.load();
+  }
+
  private:
   void listObjectsHelper(const string& prefix, const string& delimiter,
                          const string& marker, vector<string>* objects,
@@ -257,5 +267,9 @@ class S3Util {
   SDKOptions options_;
   std::string uri_;
   const uint32_t read_ratelimit_mb_;
+  // To track the number of S3Util instances. Only call Aws::ShutdownAPI() when
+  // there is no other instance exists.
+  static std::atomic<std::uint32_t> counter_;
 };
+
 }  // namespace common
--- a/common/s3util.h
+++ b/common/s3util.h
@@ -31,6 +31,7 @@
 #include <atomic>
 #include <iosfwd>
 #include <map>
+#include <mutex>
 #include <string>
 #include <tuple>
 #include <vector>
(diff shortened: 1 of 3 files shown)


newpoo:
Can you describe your use case and the exact problem you want to fix?

The current PR could introduce even more problems then it currently has.

weipihui:
This change tries to fix race condition problem between different S3Util instances which may cause job crash. 
Long story: 
As we know, aws-cpp-sdk-core maintains a global pointer `s_HttpClientFactory` to create Http request. When S3Util instance destructs, it calls Aws::ShutdownAPI(options_) which set s_HttpClientFactory to nullptr. 
When we have multiple S3Util instances, there will be race condition when some instances are destructed and some are not. The job will crash when a S3Util instance trys to read s3 file while other S3Util isntance is destructed. 

See: https://github.com/aws/aws-sdk-cpp/blob/f02083e72b1d976a34765a64da4a564d307b2afa/aws-cpp-sdk-core/source/http/HttpClientFactory.cpp#L157

common/s3util.h
```diff
@@ -187,7 +189,10 @@ class S3Util {
   }
 
   ~S3Util() {
-    Aws::ShutdownAPI(options_);
+    --counter_;
+    if (counter_.load() == 0) {
```
newpoo: This is a known issue with the current interface design. Adding an atomic reference count couldn't solve the problem. Imagine thread A finds load() returns 0, then thread B creates a new S3Util object, and then thread A calls ShutdownAPI().

A quick and dirty fix it is to make S3Util::S3Util() private, and merge the code https://github.com/pinterest/rocksplicator/blob/master/rocksdb_admin/admin_handler.cpp#L679-L712 into the static function BuildS3Util().

weipihui: You are right, the atomic counter does not solve the problem completely. The code you linked does not solve our problem either. Let's sync on Monday for a better plan. Thanks for the review.

newpoo: That's why "Can you describe your use case and the exact problem you want to fix?". As S3util can be used in various ways to trigger the bug, the code linked is how we managed to get around the problem.

newpoo:
BTW: wondering if you really need to use C++ AWS S3 client? Clients in other higher level languages are easier to use and more battle tested.

weipihui:
Yes, we do need C++ S3 client.

By the way, I have changed the code again to:
(1) make S3Util constructor effectively private. By a brief code search, I do not see code that use S3Util::S3Util(...) directly. So I guess it should be safe.
(2) Make the counter to better control when to initialize/cleanup underlying aws core status.

Please take a look and let me know if this implementation makes more sense when you have time. Thanks!

common/s3util.h
```diff
@@ -245,17 +250,51 @@ class S3Util {
...
   }
 
+  // For test only.
+  static uint32_t getInstanceCounter() {
+    std::lock_guard<std::mutex> guard(counter_mutex_);
+    return instance_counter_;
+  }
+
  private:
+  struct ThisIsPrivate {
```
newpoo: Why not simply move S3Util() into the private section?

common/tests/s3_util_test.cpp
```diff
@@ -140,6 +140,24 @@ TEST(S3UtilTest, DirectIOTest) {
   fs::remove(file_path);
 }
 
+TEST(S3UtilTest, CallDestructorOnlyOnce) {
+  SDKOptions options;
+  Aws::InitAPI(options);
```
newpoo: Why call InitAPI()?
We can just remove the whole test, as it doesn't really test much but adds unneeded complexity to the S3Util interface.

common/s3util.h
weipihui: The S3Util::BuildS3Util() uses std::make_shared<S3Util>() to create a std::shared_ptr<S3Util>. 
std::make_shared requires the S3Util constructor to be public. So it won't compile if we move constructor to private section.

common/tests/s3_util_test.cpp
weipihui: Good cache. Forgot to delete this. :)

weipihui: Changed and committed the code. 
I will leave the test here for now. Will delete when we all agree that it is not necessary. 

I feel that with this PR, the code https://github.com/pinterest/rocksplicator/blob/master/rocksdb_admin/admin_handler.cpp#L679-L712  will not be necessary anymore to avoid Aws::InitAPI() and Aps::ShutdownApi() not match problem.

weipihui:
> It looks good to me as a quick fix.
>
> Ideally, we'd remove bucket_ from the S3Util object and always pass it as a parameter to its  member functions. Because S3Client has an internal thread pool, we don't want to have multiple ones within a process.

acknowledged. 
Sure, at this moment, we have multiple S3Util instances for different buckets. Let's make a TODO for now and will make it better in later PRs.

PR #215, opened 2018-03-02
Fix an issue when addDB after DB is removed when rocksdb_allow_overlapping_keys is disabled

jinruh:


rocksdb_admin/admin_handler.cpp
```diff
@@ -786,6 +786,8 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   clearMetaData(request->db_name);
   if (!FLAGS_rocksdb_allow_overlapping_keys) {
     // clear DB if overlapping keys are not allowed
+    auto db_role = db->IsSlave() ?
```
cpd85: when we allow overlapping keys, the seqno will increment, but we won't be able to call GetUpdatesSince(), whats the reason for this change?

cpd85: sorry I actually missed the ! in the condition

PR #216, opened 2018-03-02
Fix an issue when addDB after DB is removed when rocksdb_allow_overlapping_keys is disabled

jinruh:
updated, also added uptream address


newpoo:
lgtm

PR #224, opened 2018-03-20
Log once every 1000 times to avoid disk usage issue.

weipihui:
This is to avoid disk usage issue for service. We observed that this log printing 20+GB logs which caused service crash.


weipihui:
I checked the failed checks. It seems not related to this change. there are some permission issues.

weipihui:
ok. it is passed this time.

PR #226, opened 2018-03-21
[cluster_management] don't use compareUnsigned() which is 1.8 only

newpoo:


cluster_management/src/main/java/com/pinterest/rocksplicator/MasterSlaveStateModelFactory.java
```diff
@@ -165,7 +165,8 @@ public void onBecomeMasterFromSlave(Message message, NotificationContext context
             LOG.error("Failed to get latest sequence number from " + hostName + " for " + dbName);
             continue;
           }
-          if (Long.compareUnsigned(highestSeq, seq) < 0) {
+          // if (Long.compareUnsigned(highestSeq, seq) < 0) {
```
jinruh: just remove this? or add a comment?

PR #238, opened 2018-04-16
[rocksdb_admin] switched overlapping keys to be per segment specified

cpd85:


rocksdb_admin/admin_handler.cpp
```diff
@@ -57,6 +57,9 @@ DEFINE_string(shard_config_path, "",
 DEFINE_bool(rocksdb_allow_overlapping_keys, false,
             "Allow overlapping keys in sst bulk load");
 
+DEFINE_string(allow_overlapping_keys_segments, "",
+              "comma separated list of segments supporting overlapping keys");
```
newpoo: Can we add "it takes priority over rocksdb_allow_overlapping_keys"

rocksdb_admin/admin_handler.cpp
```diff
@@ -238,6 +241,8 @@ AdminHandler::AdminHandler(
   if (db_manager_ == nullptr) {
     db_manager_ = CreateDBBasedOnConfig(rocksdb_options_);
   }
+  folly::split(",", FLAGS_allow_overlapping_keys_segments,
```
newpoo: Can we make allow_overlapping_keys_segments_  a set/unordered_set and use folly::splitTo()?
Though we don't really care about performance here, the code for looking up will be more elegant.

rocksdb_admin/admin_handler.cpp
cpd85: I'm not sure thats accurate. If FLAGS_rocksdb_allow_overlapping_keys = true and a segment doesn't exist in allow_overlapping_keys_segments, we still allow overlapping keys for the segment

rocksdb_admin/admin_handler.cpp
```diff
@@ -806,7 +817,14 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   }
 
   clearMetaData(request->db_name);
-  if (!FLAGS_rocksdb_allow_overlapping_keys) {
+
+  bool allow_overlapping_keys = allow_overlapping_keys_segments_.find(
+                                    admin::DbNameToSegment(request->db_name)) !=
+                                allow_overlapping_keys_segments_.end();
```
newpoo: fix indention.

PR #241, opened 2018-05-07
[common/s3util] add a flag to disable stream buffer

newpoo:
to reduce iowait for counter ads cluster


common/s3util.cpp
```diff
@@ -50,7 +50,8 @@ using Aws::S3::Model::HeadObjectRequest;
 
 DEFINE_int32(direct_io_buffer_n_pages, 1,
              "Number of pages we need to set to direct io buffer");
-
+DEFINE_bool(disable_s3_download_stream_buffer, false,
+            "disalbe the stream buffer used by s3 downloading");
```
jinruh: typo: disable

PR #247, opened 2018-05-18
[controller] remove controller since we adopted helix

newpoo:


shuzhang1989:
Should we move it to some deprecated folder in case we want it back in the future? controller can still be easily set up.

newpoo:
If you want to keep it, you may reset to the position before this commit, and copy it to another place.

newpoo:
Just don't want to keep two cluster management code in the repo to confuse people.

PR #265, opened 2018-10-11
Add a BulkIngestionStateModel

cpd85:


cluster_management/src/main/java/com/pinterest/rocksplicator/BulkIngestionStateModelFactory.java
```diff
@@ -0,0 +1,150 @@
...
+          // find live replicas
+
+          for (Map.Entry<String, String> instanceNameAndRole : stateMap.entrySet()) {
+            String role = instanceNameAndRole.getValue();
+            if (!role.equalsIgnoreCase("ONLINE")) {
+              continue;
+            }
+
+            String hostPort = instanceNameAndRole.getKey();
+            String host = hostPort.split("_")[0];
```
sjoshi6: perform the split and check the size of the array before using the values.
Also can we check for a null value from .getKey() using Preconditions.checkNotEmpty() or something similiar

cluster_management/src/main/java/com/pinterest/rocksplicator/BulkIngestionStateModelFactory.java
```diff
@@ -0,0 +1,150 @@
...
+
+            // We've got an online host that has the partition. Copy the data
+            String hdfsPath = "/rocksplicator/" + cluster + "/" + dbName + "/" + hostPort + "/"
+                + String.valueOf(System.currentTimeMillis());
+
+            // backup a snapshot from the upstream host, and restore it locally
+            LOG.error("Backup " + dbName + " from " + hostPort);
+            Utils.backupDB(host, port, dbName, hdfsPath);
+            LOG.error("Restore " + dbName + " from " + hdfsPath);
+            Utils.closeDB(dbName, adminPort);
```
sjoshi6: if either backup or restore throws an exception do we want to try using other hosts to copy the data ? in the current code we will go to the catch block and exit.

rocksdb_admin/admin_handler.cpp
```diff
@@ -556,6 +556,9 @@ void AdminHandler::async_tm_checkDB(
   response.set_seq_num(db->rocksdb()->GetLatestSequenceNumber());
   response.set_wal_ttl_seconds(db->rocksdb()->GetOptions().WAL_ttl_seconds);
   response.set_is_master(!db->IsSlave());
+  DBMetaData db_meta_data = getMetaData(request->db_name);
+  response.db_meta_data = db_meta_data;
+  response.__isset.db_meta_data = true;
```
newpoo: nit: replace this 3 lines with response.set_db_meta_data(getMetaData(request->db_name));

cluster_management/src/main/java/com/pinterest/rocksplicator/OnlineOfflineStateModelFactory.java
```diff
@@ -161,24 +161,24 @@ public void onBecomeOfflineFromError(Message message, NotificationContext contex
...
+    public void checkSanity(String fromState, String toState, Message message) {
       if (fromState.equalsIgnoreCase(message.getFromState())
           && toState.equalsIgnoreCase(message.getToState())
           && resourceName.equalsIgnoreCase(message.getResourceName())
           && partitionName.equalsIgnoreCase(message.getPartitionName())) {
         return;
       }
 
-      LOG.error("Invalid meesage: " + message.toString());
+      LOG.error("Invalid message: " + message.toString());
```
newpoo: could you move getLocalAdminClient() back to it original place to minimize the number of changed lines? That will make it easier for people to review.

It took me a while to figure out that the real changes are this typo and private->public.

Can you change checkSanity to protected.

cluster_management/src/main/java/com/pinterest/rocksplicator/BulkIngestionStateModelFactory.java
```diff
@@ -0,0 +1,150 @@
...
+    zkClient.start();
+  }
+
+  @Override
+  public StateModel createNewStateModel(String resourceName, String partitionName) {
+    LOG.error("Create a new state for " + partitionName);
+    return new BulkIngestionStateModel(resourceName, partitionName, adminPort, cluster, zkClient, host);
+  }
+
+  public static class BulkIngestionStateModel extends OnlineOfflineStateModelFactory.OnlineOfflineStateModel {
```
newpoo: How do you plan to setup the replication graph if it extends from OnlineOfflineStateModel.

It would be nice if you could summarize the idea about how do you plan to support polaris use case.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -170,6 +170,9 @@ public Participant(String zkConnectString, String clusterName, String instanceNa
     } else if (stateModelType.equals("MasterSlave")) {
       stateModelFactory = new MasterSlaveStateModelFactory(instanceName.split("_")[0],
           port, zkConnectString, clusterName);
+    } else if stateModelType.equals("BulkIngestion") {
```
newpoo: use a better name than BulkIngestion? which doesn't indicate data could be online updated.

cluster_management/src/main/java/com/pinterest/rocksplicator/OnlineOfflineStateModelFactory.java
cpd85: Is it good java practice to have private functions above public functions?

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
cpd85: I think we should probably meet in person to discuss this diff, it is actually intended for bulkingestion and not for online updates

PR #266, opened 2018-10-23
Block signal handling in java to stop overriding of GracefulShutdownHandler

cpd85:


rocksdb_admin/helix_client.cpp
```diff
@@ -34,7 +34,7 @@ JNIEnv* createVM(const std::string& class_path) {
   JavaVM* jvm;
   JNIEnv* env;
   JavaVMInitArgs args;
-  JavaVMOption options[2];
+  JavaVMOption options[3];
 
   args.version = JNI_VERSION_1_6;
   args.nOptions = 2;
```
newpoo: this needs to be changed?

rocksdb_admin/helix_client.cpp
```diff
@@ -45,6 +45,8 @@ JNIEnv* createVM(const std::string& class_path) {
   LOG(INFO) << "classpath=" << class_opt;
   options[0].optionString = const_cast<char*>(class_opt.c_str());
   options[1].optionString = "-verbose:jni";
+  // use -Xrs to block JVM signal handling, so graceful shutdown still works
```
newpoo: remove ", so graceful ..."?

rocksdb_admin/helix_client.cpp
cpd85: good catch... I tested this by editing the submodule in cosmos and forgot to copy this part over

PR #268, opened 2018-11-03
[application db] add db size stats

newpoo:


newpoo:
manually checked the output is as expected

PR #276, opened 2019-01-19
add gflags endpoint

shgchg:
Added a gflags endpoint so we can check effective gflag values.


shgchg:
@newpoo Thanks for the quick review!

PR #278, opened 2019-02-01
[common] make tls enabling on a per client basis

cpd85:
We don't want to always connect to everything when the tls files are set, so we prefer to do it on a per client basis

### DIFF
--- a/common/thrift_client_pool.h
+++ b/common/thrift_client_pool.h
@@ -50,12 +50,6 @@ DECLARE_int32(tcp_user_timeout_ms);
 
 DECLARE_bool(channel_enable_snappy);
 
-DECLARE_string(tls_certfile);
-
-DECLARE_string(tls_trusted_certfile);
-
-DECLARE_string(tls_keyfile);
-
 namespace common {
 
 /*
@@ -338,7 +332,7 @@ class ThriftClientPool {
   explicit ThriftClientPool(
       const std::vector<folly::EventBase*>& evbs,
       const std::shared_ptr<folly::SSLContext>* ssl_ctx = nullptr)
-      : ssl_ctx_(ssl_ctx) {
+      : event_loops_(), ssl_ctx_(ssl_ctx) {
     CHECK(!evbs.empty());
     event_loops_.reserve(evbs.size());
     for (const auto& evb : evbs) {
@@ -400,9 +394,9 @@ class ThriftClientPool {
         ssl_ctx_ == nullptr ? nullptr : std::atomic_load_explicit(ssl_ctx_, std::memory_order_acquire);
     event_loops_[idx].evb_->runInEventBaseThreadAndWait(
         [&client, &event_loop = event_loops_[idx], &addr, &is_good,
-         connect_timeout_ms, aggressively, ssl_ctx] () mutable {
+         connect_timeout_ms, aggressively, ssl_ctx = std::move(ssl_ctx)] () mutable {
           auto channel = event_loop.getChannelFor(addr, connect_timeout_ms,
-                                                  is_good, aggressively, ssl_ctx);
+                                                  is_good, aggressively, std::move(ssl_ctx));
 
           event_loop.cleanupStaleChannels(addr);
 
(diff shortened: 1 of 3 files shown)


newpoo:
We need to check the thread safety of SSLContext. If not, we can pass in a struct of the 3 file paths.

common/thrift_client_pool.h
```diff
@@ -446,11 +439,11 @@ class ThriftClientPool {
 
  private:
   std::vector<EventLoop> event_loops_;
+  std::shared_ptr<folly::SSLContext> ssl_ctx_;
```
newpoo: As mentioned in the previous comment, we need to make it std::shared_ptr<folly::SSLContext>&, otherwise it's can't be modified externally.

common/thrift_client_pool.h
```diff
@@ -404,11 +395,13 @@ class ThriftClientPool {
 
     Deleter deleter(event_loops_[idx].evb_);
     std::unique_ptr<T, Deleter> client(nullptr, deleter);
+    auto ssl_ctx =
+        std::atomic_load_explicit(&ssl_ctx_, std::memory_order_acquire);
     event_loops_[idx].evb_->runInEventBaseThreadAndWait(
         [&client, &event_loop = event_loops_[idx], &addr, &is_good,
-         connect_timeout_ms, aggressively] () mutable {
+         connect_timeout_ms, aggressively, ssl_ctx] () mutable {
```
newpoo: move capture it.

common/thrift_client_pool.h
```diff
@@ -404,11 +395,13 @@ class ThriftClientPool {
...
           auto channel = event_loop.getChannelFor(addr, connect_timeout_ms,
-                                                  is_good, aggressively);
+                                                  is_good, aggressively, ssl_ctx);
```
newpoo: std::move(ssl_ctx)

common/thrift_client_pool.h
```diff
@@ -339,15 +326,19 @@ class ThriftClientPool {
...
     CHECK_GT(n_io_threads, 0);
   }
 
   // Create a new pool by using the evbs, which are supposed to be driven by
   // some other threads. It's users' responsibility to ensure that evbs and
   // the threads driving them outlive the pool object.
-  explicit ThriftClientPool(const std::vector<folly::EventBase*>& evbs) {
+  explicit ThriftClientPool(
+      const std::vector<folly::EventBase*>& evbs,
+      const std::shared_ptr<folly::SSLContext> ssl_ctx = nullptr)
```
newpoo: we need to use std::shared_ptr<folly::SSLContext>&

common/thrift_client_pool.h
```diff
@@ -339,15 +326,19 @@ class ThriftClientPool {
   // pool.
   explicit ThriftClientPool(
       const uint16_t n_io_threads =
-      static_cast<uint16_t>(FLAGS_default_thrift_client_pool_threads))
-        : event_loops_(n_io_threads) {
+          static_cast<uint16_t>(FLAGS_default_thrift_client_pool_threads),
+      const std::shared_ptr<folly::SSLContext> ssl_ctx = nullptr)
```
newpoo: same here

common/thrift_client_pool.h
```diff
@@ -339,15 +326,19 @@ class ThriftClientPool {
...
   }
 
   // Create a new pool by using the evbs, which are supposed to be driven by
   // some other threads. It's users' responsibility to ensure that evbs and
   // the threads driving them outlive the pool object.
-  explicit ThriftClientPool(const std::vector<folly::EventBase*>& evbs) {
+  explicit ThriftClientPool(
+      const std::vector<folly::EventBase*>& evbs,
+      const std::shared_ptr<folly::SSLContext>* ssl_ctx = nullptr)
+      : ssl_ctx_(ssl_ctx) {
```
newpoo: nit: can you also explicitly default construct event_loops_ before sst_ctx_.

PR #279, opened 2019-02-07
Add putObject and putObjectCallable function in S3Util class to support uploading files to S3

hbxie:


hbxie:
@shuzhang1989 Could you help to take a look? Thank you very much.

shuzhang1989:
is it the same change in the phabricator?

hbxie:
> is it the same change in the phabricator?

Yes. Exactly the same. Thank you

PR #282, opened 2019-05-07
Add gflags s3_download_limit_mb

sihanwang-pinterest:
Provide a way to control the download limit on the rocksplicator side
instead of client side. If gflags set, we will ignore the request
level s3 download limit.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -74,6 +74,8 @@ DEFINE_bool(s3_direct_io, false, "Whether to enable direct I/O for s3 client");
 DEFINE_int32(max_s3_sst_loading_concurrency, 999,
              "Max S3 SST loading concurrency");
 
+DEFINE_int32(s3_download_limit_mb, 0, "S3 download sst bandwidth");
+
 namespace {
 
 const int kMB = 1024 * 1024;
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -789,7 +791,7 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   std::shared_ptr<common::S3Util> local_s3_util;
```
newpoo: can we just do the following here.

`if (FLAGS_s3_download_limit_mb > 0) { request-> s3_download_limit_mb = FLAGS_s3_download_limit_mb; }`

newpoo: I meant to put it immediately after line 791. Then we don't need other changes any more.

sihanwang-pinterest: good point, updated

sjoshi6: Do a max check on the FLAG value before setting it to the s3_request ?

rocksdb_admin/admin_handler.cpp
```diff
@@ -787,6 +789,9 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   // request, which is not even on any critical code path. Otherwise, we will
   // see latency spike when uploading data to production clusters.
   std::shared_ptr<common::S3Util> local_s3_util;
+  if (FLAGS_s3_download_limit_mb > 0) {
```
sjoshi6: add a max check on the FLAG value before setting it to the s3 request ?

rocksdb_admin/admin_handler.cpp
sihanwang-pinterest: it depends on do we want to follow the rule "gflag as first priority" or not.

sjoshi6: Regardless of whether we use GFLAGS as the first priority or not, it might be better to add a check against a max download limit size (2G / 5G) to make sure that if we accidentally set the value too high in GFLAG we will never cross a max limit

sihanwang-pinterest: Yes, may have another pull request for this check, besides glflag, our request level also needs this kind of check

PR #284, opened 2019-05-13
Expose helix participant log config

sihanwang-pinterest:


rocksdb_admin/helix_client.cpp
```diff
@@ -27,17 +27,18 @@
 #include "common/network_util.h"
 
 DECLARE_int32(port);
+DEFINE_string(helix_log_config, "", "Participant log config");
```
newpoo: essentially this can be any flags passed to JVM? If yes, shall we rename it?

sihanwang-pinterest: updated

rocksdb_admin/helix_client.cpp
```diff
@@ -47,6 +50,7 @@ JNIEnv* createVM(const std::string& class_path) {
   options[1].optionString = "-verbose:jni";
   // use -Xrs to block JVM signal handling
   options[2].optionString = "-Xrs";
+  options[3].optionString = const_cast<char*>(FLAGS_helix_log_config_file_path.c_str());
```
newpoo: could you please verify that it will work fine even the gflag is set to empty string?

sihanwang-pinterest: Updated the pull request, and tested with empty string and arbitrary string, jvm will ignore the wrong config.

PR #289, opened 2019-05-17
Add spectator role to rocksplicator

terry-f-shen:


cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -80,6 +81,18 @@ public void onCallback(NotificationContext notificationContext) {
...
 
+  @Override
+  public void onConfigChange(List<InstanceConfig> configs, NotificationContext changeContext) {
+    if (updateDisabledHosts()) {
+      generateShardConfig();
+    }
+  }
+
+  @Override
+  public void onExternalViewChange(List<ExternalView> externalViewList, NotificationContext changeContext) {
```
newpoo: ideally we should use the externalViewList passed in to the callback to reduce the load on zk.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -140,10 +147,11 @@ public static void main(String[] args) throws Exception {
     final String stateModelType = cmd.getOptionValue(stateModel);
     final String postUrl = cmd.getOptionValue(configPostUrl);
     final String instanceName = host + "_" + port;
+    final Boolean runAsSpectator = !cmd.hasOption(disableSpectator);
```
newpoo: rename it to runAsSpectatorOnly? 
nit: use boolean

PR #292, opened 2019-05-20
Add timeutil files and some metrics

rajathprasad:
Adding util files for time functions and some metrics.

### DIFF
--- a/common/kafka/kafka_watcher.cpp
+++ b/common/kafka/kafka_watcher.cpp
@@ -171,8 +171,8 @@ uint32_t KafkaWatcher::ConsumeUpToNow(const kafka::KafkaConsumer& consumer) {
   topic_partition_to_prev_offset.reserve(topic_names.size());
   TopicPartitionToValueMap<uint32_t> topic_partition_to_message_num;
   topic_partition_to_message_num.reserve(topic_names.size());
-  auto timeout_timestamp_ms = common::GetCurrentTimestamp(
-      common::TimeUnit::kMillisecond) +
+  auto timeout_timestamp_ms = common::timeutil::GetCurrentTimestamp(
+      common::timeutil::TimeUnit::kMillisecond) +
       kafka_init_blocking_consume_timeout_ms_;
 
   // End the loop when finding fresher message or hitting partition end for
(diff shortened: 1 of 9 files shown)


common/kafka/kafka_watcher.cpp
```diff
@@ -24,7 +24,7 @@
 
 #include "boost/functional/hash.hpp"
 #include "common/stats/stats.h"
-#include "common/timeutil/timeutil.h"
+#include "common/timeutil.h"
```
newpoo: any idea why the compiling didn't fail previously? we need to include the code into the build.

common/kafka/stats_enum.h
```diff
@@ -24,6 +24,7 @@
 enum CounterIdx {
```
newpoo: I am not sure how this would be used?

For our stats library, it only allows to register one metric vector and one counter vector in one process. If a main service already has the stats vectors, we will have to use the "string" type stat name in libraries like common/kafka here.

newpoo:
You may add rocksplicator as the reviewer for future PRs.

common/kafka/kafka_watcher.cpp
rajathprasad: I didn't realize it's not compiled. Added all the kafka files as a library now.

common/kafka/stats_enum.h
rajathprasad: Done

common/kafka/stats_enum.h
```diff
@@ -18,47 +18,19 @@
 #include <vector>
 #include "common/stats/stats.h"
 
-#define STATS_ADD_METRIC common::Stats::get()->AddMetric
-#define STATS_INCR common::Stats::get()->Incr
+const std::string kKafkaConsumerSeek = "kafka_consumer_seek";
```
newpoo: It's not a good practice to define std::string object in header files.
The reason and a solution can be found here: https://stackoverflow.com/a/10201929/4959149

common/kafka/stats_enum.cpp
```diff
@@ -25,42 +25,7 @@ DEFINE_string(index_type, "", "The index type");
...
-        "kafka_watcher_blocking_consume_timeout")
-#undef NEW_COUNTER_STAT
-};
-
-std::string getFullCounterName(const CounterIdx idx,
-                               const std::initializer_list<std::string>& tags) {
-  return getFullMetricName(kCounterNames[idx], tags);
-}
-
-std::string getFullMetricName(const std::string& metric_name,
```
newpoo: Not sure how frequently this function will be called.
We may want to optimize it by call a std::string::reserve() on full_metric_name. We can do it in a separate diff to minimize the change to muse.

common/kafka/stats_enum.h
rajathprasad: Thanks for pointing this out, will fix it.

common/kafka/stats_enum.cpp
rajathprasad: It will be called very frequently, will do as you suggest in a separate diff.

PR #293, opened 2019-05-22
[kafka-consumer] Remove anonymized namespace and give unique names fo…

rajathprasad:
…r gflags

1. The anonymous namespace for KafkaWatcher is causing issues, so adding it to the
kafka namespace.
2. Making the gflag names unique so that they don't clash with gflags defined
   by individual services.

### DIFF
--- a/common/kafka/kafka_broker_file_watcher.cpp
+++ b/common/kafka/kafka_broker_file_watcher.cpp
@@ -22,7 +22,7 @@
 #include "glog/logging.h"
 #include "common/file_watcher.h"
 
-namespace {
+namespace kafka {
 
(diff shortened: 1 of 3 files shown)


rajathprasad:
> The code changes lgtm.
> 
> There are 2 things we need to think through.
> 
> 1. since we removed some gflags and changed some gflag names, we need to make sure all current production cluster configuration is changed accordingly when rolling out this change.
> 2. stat names are changed, we want to make sure dashboards and alerts for the current clusters are changed accordingly.

Thanks for the review. Yes, will make sure the new gflags are added to the env. For stat name changes, will try to keep them the same as before so we won't have to change alerts/dashboards. Will test all this before rolling it out.

PR #294, opened 2019-05-22
[StatusServer] Fix query params parsing

IlldianX:
microhttpd already handled uri parsing, hence the url `ServeCallback` received already only contains the endpoint string. To gather the query params, we have to use MHD_get_connection_values.

This change has been locally tested on Scorpion.

### DIFF
--- a/common/stats/status_server.cpp
+++ b/common/stats/status_server.cpp
@@ -12,7 +12,6 @@
 /// See the License for the specific language governing permissions and
 /// limitations under the License.
 
-
 // Implementation of Status Server based on microhttpd.
 
 #include "common/stats/status_server.h"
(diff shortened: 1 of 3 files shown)


newpoo:
You will need to fix the unit tests introduced in https://github.com/pinterest/rocksplicator/pull/14/files

IlldianX:
> You will need to fix the unit tests introduced in https://github.com/pinterest/rocksplicator/pull/14/files

Just submitted a fix to the tests. The old code only works when the URI is already escaped(microhttpd is not unescaping it during parsing). I can support that, although it seems unlikely that ppl will manually escape the url when curling the status server :\

IlldianX:
Made it backward-compatible and tests passed. Merging.

PR #295, opened 2019-05-28
[rocksdb_admin] Add kafka message ingestion APIs.

rajathprasad:
These API's will be used to add the functionality of kafka message ingestion in
rocksplicator.


rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -149,6 +149,27 @@ struct AddS3SstFilesToDBResponse {
   # for future use
 }
 
+struct StartMessageIngestionRequest {
+  1: required string db_name,
+  2: required string topic_name,
+  3: required string kafka_broker_serverset_path,
+  4: optional i64 partition_number,
```
cpd85: If the client doesn't provide partition_number and seek_timestamp_ms, what do we do?

PR #297, opened 2019-06-05
Add initial implementations for kafka ingestion start/stop APIs.

rajathprasad:
This diff adds the implementations for start/stop kafka API's which will be
called by helix during state transitions.

This is just in initial implementation focussing on the kafka consumer
threads synchronization.


rocksdb_admin/admin_handler.h
```diff
@@ -130,6 +128,12 @@ class AdminHandler : virtual public AdminSvIf {
   std::unique_ptr<rocksdb::DB> removeDB(const std::string& db_name,
                                         AdminException* ex);
 
+  // Kafka consumer related helper functions
+  int getPartitionIDFromDBName(const std::string& db_name);
+  std::string getHostName();
+  std::string getConsumerGroupId(const std::string& topic_name,
+      const std::string& partition_id);
```
newpoo: make these 3 functions static? or move them into the anonymous namespace in admin_handler.cpp.

I slightly perfer the second approach, which will make the admin handler class interface less complicated.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  const size_t n = db_name.find_first_of("0123456789");
+  if (n != std::string::npos){
+    return std::stoi(db_name.substr(n, std::string::npos));
+  }
+  LOG(ERROR) << "No partition id found in db name " << db_name;
+
+  return -1;
+}
+
+std::string AdminHandler::getHostName() {
```
newpoo: can we use ip instead? if yes, we can reuse function getLocalIPAddress() in network_util.h.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
   callback->result(AddS3SstFilesToDBResponse());
 }
 
+int AdminHandler::getPartitionIDFromDBName(const std::string& db_name) {
```
newpoo: this might not work when a segment name contains numbers. Such as p2p1, p2p2, etc.

We can do something similar as in sharding_util.cpp by assuming the partition id part is always of length 5.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+
+  admin::AdminException e;
+  e.errorCode = AdminErrorCode::DB_ADMIN_ERROR;
+
+  const auto db_name = request->db_name;
+  const auto topic_name = request->topic_name;
+  const auto kafka_broker_serverset_path = request->kafka_broker_serverset_path;
+  const auto replay_timestamp_ms = request->replay_timestamp_ms;
+
+  LOG(INFO) << "Called startMessageIngestion for " << db_name;
```
newpoo: also log topic, serverset path and  reply timestamp?

rocksdb_admin/admin_handler.cpp
```diff
@@ -81,6 +89,43 @@ namespace {
...
+  const auto ts = message.timestamp();
+  if (ts.type == RdKafka::MessageTimestamp::MSG_TIMESTAMP_CREATE_TIME) {
+    return ts.timestamp / kMillisPerSec;
+  }
+
+  // We only expect the timestamp to be create time.
+  return -1;
+}
+
+std::string ToUTC(const int64_t time_secs) {
```
terry-f-shen: I am seeing that this is only used once when printing log, can we use std::put_time directly there? https://en.cppreference.com/w/cpp/io/manip/put_time

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+
+  const auto db_name = request->db_name;
+  const auto topic_name = request->topic_name;
+  const auto kafka_broker_serverset_path = request->kafka_broker_serverset_path;
+  const auto replay_timestamp_ms = request->replay_timestamp_ms;
+
+  LOG(INFO) << "Called startMessageIngestion for " << db_name;
+
+  // Check if there's already a thread consuming the same partition.
+  if (sync_var_map.find(db_name) != sync_var_map.end()) {
```
newpoo: access to sync_var_map needs to be synchronized.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+    LOG(ERROR) << "Could not find partition in db_name" << db_name;
+    return;
+  }
+
+  const std::unordered_set<uint32_t> partition_ids_set({partition_id});
+
+  // Find the watcher for the given path. If none exists, create a new watcher
+  std::shared_ptr<::kafka::KafkaBrokerFileWatcher> kafka_broker_file_watcher;
+
+  const auto map_iter = serverset_path_to_file_watcher_map.find(
```
newpoo: same to serverset_path_to_file_watcher_map.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  LOG(INFO) << "Now consuming live messages for " << db_name;
+
+  // Live messages continue to consumed in a separate thread, but release
+  // callback so that helix can transition this partition to bootstrap stage.
+  callback.release()->result(StartMessageIngestionResponse());
+
+  // Consume live messages until 'Stop' API is called which will notify the
+  // global condition variable.
+  LOG(INFO) << "Consuming until notified for  " << db_name;
+  std::mutex mutex;
```
terry-f-shen: You would have to put this mutex into SyncVariables also, because both thread need to take this lock

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+      std::unordered_set<std::string>({topic_name}),
+      getConsumerGroupId(topic_name, std::to_string(partition_id)),
+      kKafkaConsumerType);
+
+  const auto kafka_watcher = std::make_unique<KafkaWatcher>(
+      kKafkaWatcherName,
+      kafka_consumer_pool,
+      -1); // kafka_init_blocking_consumer_timeout_ms ;
+
+  // With kafka_init_blocking_consumer_timeout_ms set to -1, messages from
```
terry-f-shen: The mutex need to be locked before we start consuming so that the stopper won't touch the SyncVariable before we start waiting

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +974,177 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  LOG(ERROR) << "Stopping kafka watcher";
+  kafka_watcher->StopAndWait();
+  LOG(ERROR) << "Kafka watcher stopped";
+}
+
+void AdminHandler::async_tm_stopMessageIngestion(
+    std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
+      StopMessageIngestionResponse>>> callback,
+    std::unique_ptr<StopMessageIngestionRequest> request) {
+
```
terry-f-shen: acquire the same mutex here

rocksdb_admin/admin_handler.cpp
terry-f-shen: Check this document: https://en.cppreference.com/w/cpp/thread/condition_variable

rocksdb_admin/admin_handler.cpp
terry-f-shen: +1

rocksdb_admin/tests/admin_handler_test.cpp
```diff
@@ -157,6 +157,29 @@ TEST(AdminHandlerTest, CheckDB) {
...
+                              admin::RocksDBOptionsGeneratorType());
+
+  std::string db_name;
+  int partition;
+
+  db_name = "test_db_00000";
+  partition = handler.getPartitionIDFromDBName(db_name);
+  EXPECT_EQ(partition, 0);
+
+  db_name = "test_db_00030";
```
cpd85: fwiw, helix will change the db name to be the format for rocksdb, so it'd send test_db00030 in the thrift request

rocksdb_admin/admin_handler.h
rajathprasad: Used the second approach to keep it consistent with existing function.

rocksdb_admin/admin_handler.cpp
rajathprasad: I'm planning to keep these logs only while we debug the system. In addition, `put_time` would also require converting epoch timestamp to time_t. So will keep this for now.

rocksdb_admin/admin_handler.cpp
rajathprasad: Thanks, I'd missed the part where even if we use an atomic variable, a mutex is required.

rocksdb_admin/admin_handler.cpp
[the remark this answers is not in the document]
rajathprasad: Good catch, thanks!

rajathprasad:
> Can you provide some context about how our kafka consumer library is structured and how it's supposed to be used? Such as thread pool, threading model etc. This can help us to see if the logic in admin handler is the best way to use it.

The kafka consumer library is structured into 3 classes:

1. KafkaConsumer: This is simple wrapper around librdkafka's KafkaConsumer class, has additional logic of retrying if there are failures, among other minor additions.
2. KafkaConsumerPool: This constructs a pool (or a queue) of KafkaConsumers, where each consumer consumes messages in a separate thread. From what I understand, a larger queue size is used while consuming high volume of traffic from multiple partitions. For our use case, since each KafkaConsumerPool is responsible only for a single partition, and for the sake of simplicity, we set the size to 1.
3. KafkaWatcher: Base class for watchers that want to consume from kafka. Manages the kafka
 consumer and the consuming thread. Provides functions to seek to a timestamp, consume messages and then do the live consumption in a separate thread. As of now, the base class is good enough for our use case. If in the future, we need more functionality, we can create a derived class.

rocksdb_admin/tests/admin_handler_test.cpp
rajathprasad: Fixed.

rocksdb_admin/admin_handler.cpp
```diff
@@ -1127,20 +1127,27 @@ void AdminHandler::async_tm_stopMessageIngestion(
...
+    if (sync_variable == sync_var_map.end()) {
+      e.message = db_name + " consumption already stopped";
+      callback.release()->exceptionInThread(std::move(e));
+      LOG(ERROR) << "Could not find condition variable for " << db_name <<
+                    ". Consumption might have never started or already stopped";
+      return;
+    }
+
+    // Set atomic flag and notify consuming thread.
+    {
```
terry-f-shen: It would be better to unlock sync_var_map_mutex first then lock the mutex in the sync variable

rocksdb_admin/admin_handler.cpp
```diff
@@ -94,18 +94,21 @@ const char kKafkaConsumerType[] = "rocksplicator_message_consumer";
...
 
+// Global map to keep track of serverset file watchers
+  std::unordered_map<std::string, std::shared_ptr<
+      ::kafka::KafkaBrokerFileWatcher>> serverset_path_to_file_watcher_map;
+
 // Global condition variables for kafka synchronization.
 struct SyncVariables
 {
   SyncVariables(): flag(false) {} // init flag to false
   std::atomic<bool> flag;
```
terry-f-shen: we have the mutex to serialize the operation so the std::atomic is not needed?

rocksdb_admin/admin_handler.cpp
```diff
@@ -94,18 +94,21 @@ const char kKafkaConsumerType[] = "rocksplicator_message_consumer";
...
   std::condition_variable cond_var;
+  std::mutex mutex;
 };
 std::unordered_map<std::string, std::shared_ptr<SyncVariables>> sync_var_map;
-
-// Global map to keep track of serverset file watchers
-std::unordered_map<std::string, std::shared_ptr<
-    ::kafka::KafkaBrokerFileWatcher>> serverset_path_to_file_watcher_map;
+std::mutex sync_var_map_mutex;
```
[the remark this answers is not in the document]
rajathprasad: Good idea, I've added a TODO, will add a new class (or add to the AdminHandler class) in a separate diff.

rocksdb_admin/admin_handler.cpp
```diff
@@ -1036,22 +1029,31 @@ void AdminHandler::async_tm_startMessageIngestion(
...
-  sync_var_map[db_name] = sync_variable;
+  {
+    std::lock_guard<std::mutex> sync_var_map_lock(sync_var_map_mutex);
+    sync_var_map[db_name] = sync_variable;
+  }
+  // Lock the mutex so that the synchronization variables are not modified
+  // until we wait on the condition variable.
+  std::unique_lock<std::mutex> lock(sync_variable->mutex);
 
   SCOPE_EXIT { sync_var_map.erase(db_name); };
```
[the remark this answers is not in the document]
rajathprasad: Done. I also realized we need to lock before removing from sync_var_map. Do you know if there are any consequences of attempting to lock in SCOPE_EXIT? I've ensured that `sync_var_map_mutex` is locked only while the map is read/modified, for very short durations.

rocksdb_admin/admin_handler.cpp
rajathprasad: Good catch, this will ensure the mutex is locked for very short durations.

rocksdb_admin/admin_handler.cpp
rajathprasad: Done.

rocksdb_admin/admin_handler.cpp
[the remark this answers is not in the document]
terry-f-shen: Acquiring locks in SCOPE_EXIT should be fine, and that is why adding a class to wrap lock and map is better here

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +982,184 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  const auto db_name = request->db_name;
+  const auto topic_name = request->topic_name;
+  const auto kafka_broker_serverset_path = request->kafka_broker_serverset_path;
+  const auto replay_timestamp_ms = request->replay_timestamp_ms;
+
+  LOG(INFO) << "Called startMessageIngestion for db: " << db_name << ", "
+            << "topic_name: " << topic_name << ", "
+            << "serverset path: " << kafka_broker_serverset_path << ", "
+            << "replay_timestamp_ms: " << replay_timestamp_ms;
+
```
newpoo: we may want to call getDB() here. So that we can fail the request early if db doesn't exist.

rocksdb_admin/admin_handler.cpp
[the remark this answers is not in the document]
rajathprasad: I ended up doing this in the same diff.

rocksdb_admin/admin_handler.cpp
```diff
@@ -81,6 +89,51 @@ namespace {
...
+    return ts.timestamp / kMillisPerSec;
+  }
+
+  // We only expect the timestamp to be create time.
+  return -1;
+}
+
+std::string ToUTC(const int64_t time_secs) {
+  time_t raw_time(time_secs);
+  char buf[24];
```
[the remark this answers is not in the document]
rajathprasad: should be, since the number of characters is fixed. I've changed it to 48 to be safe. I'll most likely be getting rid of this function before productionizing it.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +982,184 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  };
+
+  // Lock the mutex so that the synchronization variables are not modified
+  // until we wait on the condition variable.
+  std::unique_lock<std::mutex> lock(sync_variable->mutex);
+
+  const auto kafka_consumer_pool = std::make_shared<::kafka::KafkaConsumerPool>(
+      kKafkaConsumerPoolSize,
+      partition_ids_set,
+      kafka_broker_file_watcher->GetKafkaBrokerList(),
```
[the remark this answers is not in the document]
rajathprasad: Yes, I was aware of this problem and there is a TODO in the kafka_broker_file_watcher class. I've added a TODO here as well. I had spoken to the kafka folks, and I was told that as long as at least one broker remains the same, the brokers themselves will update their config and there is no loss of messages being consumed. But yes, we do need to fix it in the long term.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +982,184 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+      serverset_path_to_file_watcher_map[kafka_broker_serverset_path] =
+          kafka_broker_file_watcher;
+    } else {
+      kafka_broker_file_watcher = map_iter->second;
+    }
+  }
+
+  // Create new synchronization variables for this partition
+  auto sync_variable = std::make_shared<SyncVariables>();
+  {
```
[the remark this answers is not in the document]
rajathprasad: Thanks for the suggestion, done.

rocksdb_admin/detail/file_watcher_manager.h
```diff
@@ -0,0 +1,52 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author rajathprasad (rajathprasad@pinterest.com)
+//
+
+#include <memory>
+#include <mutex>
```
newpoo: nit: #include <string>

rocksdb_admin/detail/file_watcher_manager.h
```diff
@@ -0,0 +1,52 @@
...
+namespace kafka {
+  class KafkaBrokerFileWatcher;
+}
+
+namespace admin {
+namespace detail {
+
+  // FileWatcherManager is a singleton class which manages the kafka broker
+  // file watchers for this process.
+  class FileWatcherManager {
```
newpoo: nit: Can we rename it to something like KafkaBrokerFileWatcher?

rocksdb_admin/detail/file_watcher_manager.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "rocksdb_admin/detail/file_watcher_manager.h"
+
+#include "common/kafka/kafka_broker_file_watcher.h"
+
+namespace admin {
+namespace detail {
+
+std::shared_ptr<::kafka::KafkaBrokerFileWatcher>
+    FileWatcherManager::getFileWatcher(const std::string& file_path) {
```
newpoo: remove the leading spaces.

rocksdb_admin/detail/file_watcher_manager.h
```diff
@@ -0,0 +1,52 @@
...
+  public:
+    std::shared_ptr<::kafka::KafkaBrokerFileWatcher> getFileWatcher(
+        const std::string &file_path);
+
+    static FileWatcherManager& getInstance();
+
+  private:
+    FileWatcherManager() {};
+
+    FileWatcherManager(FileWatcherManager const &) {};
```
newpoo: it'd better to just delete the copy constructor by "= delete". 
nit: In this repo, we usually put const in front of the type.

rocksdb_admin/detail/file_watcher_manager.h
```diff
@@ -0,0 +1,52 @@
...
+
+    FileWatcherManager &operator=(FileWatcherManager const &) {};
```
newpoo: same as above.

We should also delete move constructor and move operator=.

rocksdb_admin/admin_handler.cpp
```diff
@@ -20,22 +20,31 @@
 #include "rocksdb_admin/admin_handler.h"
 
 #include <chrono>
+#include <condition_variable>
```
newpoo: can be removed?

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +966,163 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  LOG(INFO) << "Called startMessageIngestion for db: " << db_name << ", "
+            << "topic_name: " << topic_name << ", "
+            << "serverset path: " << kafka_broker_serverset_path << ", "
+            << "replay_timestamp_ms: " << replay_timestamp_ms;
+
+  db_admin_lock_.Lock(db_name);
+  SCOPE_EXIT { db_admin_lock_.Unlock(db_name); };
+  auto db = getDB(db_name, &e);
+  if (db == nullptr) {
+    callback.release()->exceptionInThread(std::move(e));
```
newpoo: nit: set e.message?

rocksdb_admin/detail/file_watcher_manager.h
rajathprasad: Renamed to KafkaBrokerFileWatcherManager. Initially had decided against it since its a mouthful.

rocksdb_admin/admin_handler.cpp
```diff
@@ -929,6 +965,167 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+    std::lock_guard<std::mutex> lock(kafka_watcher_lock_);
+    auto iter = kafka_watcher_map_.find(db_name);
+    // Verify that there is thread consuming messages for this db.
+    if (iter == kafka_watcher_map_.end()) {
+      e.message = db_name + " is not being consumed";
+      callback.release()->exceptionInThread(std::move(e));
+      LOG(ERROR) << db_name << " is not being currently consumed";
+      return;
+    }
+    kafka_watcher = iter->second;
```
newpoo: nit: we can remove iter from the map here and remove line 1120 to 1123.

rajathprasad: I would like to remove the entry from the map only after the watcher is successfully stopped. So will keep this as is for now.

PR #298, opened 2019-06-10
[kafka] Add a new flag for cluster name

rajathprasad:
It's cleaner for the cluster name to be a separate flag (than being included
in the suffix). Separating it out in this diff.

### DIFF
--- a/common/kafka/stats_enum.cpp
+++ b/common/kafka/stats_enum.cpp
@@ -22,13 +22,17 @@
 
 DEFINE_string(kafka_stats_prefix, "", "Prefix for the kafka stats");
 DEFINE_string(kafka_stats_suffix, "", "Suffix for the kafka stats");
+DEFINE_string(kafka_cluster_name, "", "Cluster consuming the kafka messages");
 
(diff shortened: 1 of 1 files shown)


common/kafka/stats_enum.cpp
```diff
@@ -22,13 +22,17 @@
...
 
 std::string getFullStatsName(const std::string& metric_name,
                              const std::initializer_list<std::string>& tags) {
   std::string full_metric_name;
   full_metric_name.append(FLAGS_kafka_stats_prefix)
       .append(metric_name)
-      .append(FLAGS_kafka_stats_suffix);
+      .append("_")
+      .append(FLAGS_kafka_stats_suffix)
+      .append(" cluster=")
```
cpd85: up to you, but do we want to still append this if FLAGS_kafka_cluster_name is blank?

rajathprasad: Good point, but in the existing implementation it's this way, and I'm not sure how the stats client handles empty tags. So will leave it as is for now.

PR #299, opened 2019-06-11
[Thrift] Add new field to keep track of last kafka message timestamp …

rajathprasad:
…consumed.

Adding a field to DbMetaData to keep track of the last message timestamp
consumed from kafka. This will be updated periodically in the boostrap model.
It will be unused in Online-Offline model.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -310,13 +310,16 @@ bool AdminHandler::clearMetaData(const std::string& db_name) {
   return s.ok();
 }
 
-bool AdminHandler::writeMetaData(const std::string& db_name,
-                                 const std::string& s3_bucket,
-                                 const std::string& s3_path) {
+bool AdminHandler::writeMetaData(
+    const std::string& db_name,
+    const std::string& s3_bucket,
+    const std::string& s3_path,
+    const int64_t last_kafka_msg_timestamp_ms) {
   DBMetaData meta;
   meta.db_name = db_name;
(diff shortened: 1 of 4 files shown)


rocksdb_admin/admin_handler.h
```diff
@@ -134,7 +134,8 @@ class AdminHandler : virtual public AdminSvIf {
   bool clearMetaData(const std::string& db_name);
   bool writeMetaData(const std::string& db_name,
                      const std::string& s3_bucket,
-                     const std::string& s3_path);
+                     const std::string& s3_path,
+                     const int64_t& last_kafka_msg_timestamp_ms);
```
newpoo: remove &, copying an int64_t is equally expensive to copying a reference, while accessing through reference introduces an unnecessary indirection.

set a default value for the new parameter.

rocksdb_admin/admin_handler.cpp
```diff
@@ -917,7 +920,8 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
     return;
   }
 
-  writeMetaData(request->db_name, request->s3_bucket, request->s3_path);
+  writeMetaData(request->db_name, request->s3_bucket, request->s3_path,
+      -1 /* default timestamp */);
```
newpoo: if we set the default value in .h file, we don't need to do this change.

rocksdb_admin/admin_handler.h
rajathprasad: Done, thanks for the explanation.

PR #300, opened 2019-06-14
Periodically write last kafka message's timestamp to meta_db

rajathprasad:
1. Implement a new KafkaSpecManager class to manage all the objects which are
   required per db (kafka_watcher, last consumed timestamp).
2. Setup a function scheduler to periodically write the last consumed timestamp
   to meta_db. If required in the future, we can have the interval of the
   scheduler to be a config so that each use case can define it's own
   frequency of update.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -1062,7 +1062,7 @@ void AdminHandler::async_tm_startMessageIngestion(
                  << " for db:  " << db_name;
     }, std::chrono::minutes(1), /* run at interval of 1 minute */
     GetSchedulerNameID(db_name),
-    std::chrono::minutes(5) /* start delay of 5 minutes*/ );
+    std::chrono::minutes(5) /* start delay of 5 minutes */ );
 
 
   // With kafka_init_blocking_consumer_timeout_ms set to -1, messages from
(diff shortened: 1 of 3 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -1101,29 +1126,21 @@ void AdminHandler::async_tm_stopMessageIngestion(
...
+  if (!kafka_spec_manager_.isSpecPresent(db_name)) {
       e.message = db_name + " is not being consumed";
       callback.release()->exceptionInThread(std::move(e));
       LOG(ERROR) << db_name << " is not being currently consumed";
       return;
-    }
-    kafka_watcher = iter->second;
   }
 
+  auto kafka_watcher = kafka_spec_manager_.getKafkaWatcher(db_name);
```
cpd85: is it possible here to write the message that we last consumed? It would eliminate the gap between when the function was last run and the last message we've actually consumed

rocksdb_admin/admin_handler.cpp
```diff
@@ -1040,25 +1047,43 @@ void AdminHandler::async_tm_startMessageIngestion(
...
 
-  {
-    std::lock_guard<std::mutex> lock(kafka_watcher_lock_);
-    kafka_watcher_map_[db_name] = kafka_watcher;
-  }
+  auto kafka_spec = kafka_spec_manager_.addSpec(db_name, kafka_watcher,
+      meta.last_kafka_msg_timestamp_ms);
+
+  kafka_ts_updater_.addFunction(
+    [db_name, kafka_spec, this] {
```
newpoo: Did we consider put this logic into the lambda that passed to StartWith() below? that will make the code simpler.
The downside is that timestamp persistence can only be triggered when a new message is received. That should be fine?

rocksdb_admin/admin_handler.cpp
rajathprasad: Yes, I've added this. Thanks.

rocksdb_admin/admin_handler.cpp
rajathprasad: Yes, I did consider it. I preferred this since we don't have to do the time management. We would still need a global timestamp per db, or read the meta_db for every message consumed. Since this is more of a background task, and it's not super critical to run at the exact interval, thought it would be simpler to do all that in a separate thread.

newpoo: We can capture an int or a timestamp in the lambda, and for each message received, we bump the int (if we want to persist the timestamp once every N message). Or we update the captured timestamp whenever we persist a timestamp to db (if we want to persist the timestamp once every X mins). Will it work?

This will make the code much simpler. we don't need KafkaSpecs related classes any more.

rajathprasad: Sounds good, thanks for the suggestion. I'll close this PR and create a new one.

PR #301, opened 2019-06-15
Periodically write last consumed kafka message's timestamp to meta_db.

rajathprasad:
1. Periodically write the consumed message timestamp to meta_db. This period
   can be configured using gflag or zk for different use cases.
2. Choose the latest timestamp to seek to.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -84,6 +84,9 @@ DEFINE_int32(max_s3_sst_loading_concurrency, 999,
 
 DEFINE_int32(s3_download_limit_mb, 0, "S3 download sst bandwidth");
 
+DEFINE_int32(kafka_ts_update_interval, 1000, "Number of kafka messages consumed"
+                                             " before updating meta_db");
+
 namespace {
 
 const int kMB = 1024 * 1024;
@@ -1054,14 +1057,13 @@ void AdminHandler::async_tm_startMessageIngestion(
   }
 
   int64_t message_count = 0;
-  const int message_interval = 1000; // use glfag or zk to configure this
   // With kafka_init_blocking_consumer_timeout_ms set to -1, messages from
   // replay_timestamp_ms to the current are synchronously consumed. The
   // calling thread then returns after spawning a new thread to consume
   // live messages.
   kafka_watcher->StartWith(
       replay_timestamp_ms,
-      [message_count, db_name, message_interval, this](
+      [message_count, db_name, this](
           std::shared_ptr<const RdKafka::Message> message,
           const bool is_replay) mutable {
     if (message == nullptr) {
@@ -1081,7 +1083,8 @@ void AdminHandler::async_tm_startMessageIngestion(
 
     // TODO: add the logic to write the key/value to rocksdb here
 
-    if (!is_replay && (message_count % message_interval == 0)) {
+    // Update meta_db with kafka message timestamp periodically.
+    if (!is_replay && (message_count % FLAGS_kafka_ts_update_interval == 0)) {
       db_admin_lock_.Lock(db_name);
       SCOPE_EXIT { db_admin_lock_.Unlock(db_name); };
       const auto timestamp_ms = message->timestamp().timestamp;
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -1084,9 +1084,7 @@ void AdminHandler::async_tm_startMessageIngestion(
     // TODO: add the logic to write the key/value to rocksdb here
 
     // Update meta_db with kafka message timestamp periodically.
-    if (!is_replay && (message_count % FLAGS_kafka_ts_update_interval == 0)) {
-      db_admin_lock_.Lock(db_name);
-      SCOPE_EXIT { db_admin_lock_.Unlock(db_name); };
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -1045,14 +1053,17 @@ void AdminHandler::async_tm_startMessageIngestion(
     kafka_watcher_map_[db_name] = kafka_watcher;
   }
 
+  int64_t message_count = 0;
+  const int message_interval = 1000; // use glfag or zk to configure this
```
newpoo: can we add a gflag in this file and we don't need to capture message_interval in the lambda below.

rocksdb_admin/admin_handler.cpp
```diff
@@ -1070,7 +1081,16 @@ void AdminHandler::async_tm_startMessageIngestion(
 
     // TODO: add the logic to write the key/value to rocksdb here
 
-    // TODO: periodically write the message timestamp to local_meta_db
+    if (!is_replay && (message_count % message_interval == 0)) {
```
newpoo: what does is_replay exactly mean and why we don't want to write timestamp if is_replay is true?

rajathprasad: is_replay flag is set to true when the messages being consumed are part of the catch up (and not live messages). Two reasons:
1. We want this to be as performant as possible since this part is blocking the transition to the online phase. With a message interval set to 1000, we would end up writing to meta_db multiple times a second (per shard). This is not necessary.
2. The periodic update is targeted towards update/restart cases. And the likelihood of that happening when a new batch is uploaded or when a shard is introduced to a new host is less.

rocksdb_admin/admin_handler.cpp
rajathprasad: Adding it as a gflag for now. In the future, we might end up needing different intervals for each use case, so I'm leaning towards having that as part of zk config.

rocksdb_admin/admin_handler.cpp
newpoo: Thanks for the information. I am just a little concerned that how reliable is this is_replay flag and how its value is determined? Will it be true if there is a little network glitch or slowness in the kafka consumer side? If is_replay happens to be set to true for the message whose message_count is % message_interval == 0, we may skip some timestamp persistence and get some non-deterministic effect.

The overhead of one extra read and one extra write to meta_db for every 1000 DB writes seems to be small to me.

newpoo: If we are sure that is_replay is reliable, I am ok with it. just let me know.

rocksdb_admin/admin_handler.cpp
```diff
@@ -1070,7 +1083,17 @@ void AdminHandler::async_tm_startMessageIngestion(
 
     // TODO: add the logic to write the key/value to rocksdb here
 
-    // TODO: periodically write the message timestamp to local_meta_db
+    // Update meta_db with kafka message timestamp periodically.
+    if (!is_replay && (message_count % FLAGS_kafka_ts_update_interval == 0)) {
```
terry-f-shen: There can be a case that while we are doing replay and the DB was down unexpectedly, in which case we will need to start over replaying when DB is up. A DB write of once every thousand to set a milestone sounds a good balance to me, even while we are in replay mode (any way we are writing DB all the time, this is just 0.1% overhead)

rocksdb_admin/admin_handler.cpp
rajathprasad: is_replay is set to true for the synchronous consumption. Once we start consuming live messages, a new thread is spawned by the watcher and all further messages will have the flag set to false (irrespective of network or kafka delays). 
But I don't have a strong opinion for/against this. I'll remove it, we can reassess it later if this does affect performance.

rocksdb_admin/admin_handler.cpp
rajathprasad: Sounds good, I feel I was over optimizing for the catch up phase performance. Will remove it. Thanks for the feedback.

rocksdb_admin/admin_handler.cpp
```diff
@@ -1045,14 +1056,16 @@ void AdminHandler::async_tm_startMessageIngestion(
...
   // replay_timestamp_ms to the current are synchronously consumed. The
   // calling thread then returns after spawning a new thread to consume
   // live messages.
   kafka_watcher->StartWith(
       replay_timestamp_ms,
-      [](std::shared_ptr<const RdKafka::Message> message,
-          const bool is_replay) {
+      [message_count, db_name, this](
+          std::shared_ptr<const RdKafka::Message> message,
+          const bool is_replay) mutable {
```
terry-f-shen: Do we still need this?

rajathprasad: Yes, that's the function signature of the lambda required by the kafka watcher:
https://github.com/pinterest/rocksplicator/blob/97eeb18cdd5479742b98121ff6c6b5d26aa6d25a/common/kafka/kafka_watcher.h#L39

terry-f-shen: Is KafkaMessageHandler used anywhere else? If not, maybe we can remove that, doing this in a separated PR is fine

rajathprasad: Yes, this kafka library is used by the Muse team as well. This flag was added by them, so don't want to remove it.

PR #302, opened 2019-06-17
Write the consumed kafka messages to rocksdb.

rajathprasad:
Consumed messages are in the format expected by the rocksplicator service. So,
directly write them to rocksdb.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -1082,8 +1082,11 @@ void AdminHandler::async_tm_startMessageIngestion(
 
     // Write the message to rocksdb
     static const rocksdb::WriteOptions write_options;
-    auto status = db->rocksdb()->Put(write_options, *message->key(),
-        static_cast<const char *>(message->payload()));
+    auto key = rocksdb::Slice(static_cast<const char *>(message->key_pointer()),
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -1081,7 +1080,16 @@ void AdminHandler::async_tm_startMessageIngestion(
     << "msg_timestamp: " << ToUTC(msg_timestamp_secs) << " or "
     << std::to_string(msg_timestamp_secs) << " secs";
 
-    // TODO: add the logic to write the key/value to rocksdb here
+    // Write the message to rocksdb
+    static const rocksdb::WriteOptions write_options;
+    auto status = db->rocksdb()->Put(write_options, *message->key(),
```
newpoo: can we use " Slice(const char* d, size_t n)" instead?
I am not sure if message->key() and payload() guarantees that the returned char* will end with '\0' and there will be no '\0' in the middle of the string.
Also  "Slice(const char* d, size_t n)" will be more efficient, as it doesn't need to go through the string to count the size by strlen().

rajathprasad: I was not aware of that, thank you for the information. Have updated it to use slice for both key and value.

PR #309, opened 2019-07-29
[kafka_consumer] Add stats for latency (lag) in consumed kafka messages.

rajathprasad:
This can be used to setup alerts around delays in messages consumed.
Setting the threshold as 1 second, can modify that to a glfag if needed later.


rajathprasad:
@cpd85 Thanks for suggestion to convert it to a 'metric'.

PR #310, opened 2019-07-31
Add tags argument to support attach tags when uploading objects to s3

richardzhangrui:
This can enable s3 tagging to support s3 objects expiration with tags

### DIFF
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -314,7 +314,10 @@ GetObjectMetadataResponse S3Util::getObjectMetadata(const string &key) {
 
 PutObjectResponse S3Util::putObject(const string& key, const string& local_path, const string& tags) {
   PutObjectRequest object_request;
-  object_request.WithBucket(bucket_).WithKey(key).WithTagging(tags);
+  object_request.WithBucket(bucket_).WithKey(key);
+  if (!tags.empty()) {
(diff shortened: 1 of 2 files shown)


common/s3util.cpp
```diff
@@ -312,9 +312,9 @@ GetObjectMetadataResponse S3Util::getObjectMetadata(const string &key) {
   return GetObjectMetadataResponse(std::move(metadata), "");
 }
 
-PutObjectResponse S3Util::putObject(const string& key, const string& local_path) {
+PutObjectResponse S3Util::putObject(const string& key, const string& local_path, const string& tags) {
   PutObjectRequest object_request;
-  object_request.WithBucket(bucket_).WithKey(key);
+  object_request.WithBucket(bucket_).WithKey(key).WithTagging(tags);
```
newpoo: what's the function signature of WithTagging(). If it is "const string&", does it store any reference to tags? That can cause segfault as the "" temp object lifecycle is bound to tags only.

richardzhangrui: just checked. it actually has two signatures.  one is const string& and another is string&&. it is copied over to `m_tagging`.
https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-s3/include/aws/s3/model/PutObjectRequest.h

PR #312, opened 2019-08-13
Enable spectator to spectate multiple clusters

cpd85:


cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
```diff
@@ -116,25 +119,34 @@ public static void main(String[] args) throws Exception {
...
-    } catch (Exception e) {
-      LOG.error("Failed to release the mutex for cluster " + clusterName, e);
+
+    int i = 0;
+    while (true) {
+      String currentClusterName = clusterNamesList[i];
+      LOG.error("Trying to get lock for cluster " + currentClusterName);
+      InterProcessMutex mutex = new InterProcessMutex(zkClient, getClusterLockPath(currentClusterName));
+      try (Locker locker = new Locker(mutex, 5, TimeUnit.SECONDS)) {
+        Spectator spectator= new Spectator(zkConnectString, currentClusterName, instanceName);
```
rajathprasad: Will there be a quick way to tell (from the logs or other means) which host is spectating which cluster? This will help in reducing the debug time when there is an issue with a cluster.

cpd85: you could just look at the logs. I think its the same as you'd have to do right now to find the spectator for rockstore-shared-readonly

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
```diff
@@ -116,25 +119,34 @@ public static void main(String[] args) throws Exception {
...
-      LOG.error("RuntimeException thrown by cluster " + clusterName, e);
-    } catch (Exception e) {
-      LOG.error("Failed to release the mutex for cluster " + clusterName, e);
+
+    int i = 0;
+    while (true) {
+      String currentClusterName = clusterNamesList[i];
+      LOG.error("Trying to get lock for cluster " + currentClusterName);
+      InterProcessMutex mutex = new InterProcessMutex(zkClient, getClusterLockPath(currentClusterName));
+      try (Locker locker = new Locker(mutex, 5, TimeUnit.SECONDS)) {
```
newpoo: so one host will get the lock for the first cluster and start generating config for it, and all other hosts will just be blocked?
Who will be responsible to generate configs for other clusters?

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
```diff
@@ -116,25 +119,34 @@ public static void main(String[] args) throws Exception {
...
+        Spectator spectator= new Spectator(zkConnectString, currentClusterName, instanceName);
+        spectator.startListener(postUrl);
```
newpoo: should all clusters use the same post url?

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
cpd85: this is the old implementation recommended by terry. if another host cannot get some clusters lock after five seconds, it will try some other cluster until it eventually obtains some lock

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
cpd85: good catch, will update the next diff so that each post url has the cluster in it

PR #318, opened 2019-09-10
[rocksdb_admin] Add thrift structure for kafka payload

rajathprasad:
Adding a thrift structure to the kafka payload to handle different rocksdb operations
required: PUT, DELETE. Will add other operations (such as MERGE) when required.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -287,10 +287,10 @@ bool SetAddressOrException(const std::string& ip,
 }
 
 template <typename T>
-bool DecodeThriftStruct(void* data, const size_t len, T* obj) {
+bool DecodeThriftStruct(const string& data, T* obj) {
   try {
     apache::thrift::BinarySerializer::deserialize(
-        folly::IOBuf::takeOwnership(data, len).release(), *obj);
+        folly::StringPiece(data.data(), data.size()), *obj);
   } catch (const std::exception& ex) {
     LOG(ERROR) << "Error when decoding message : " << ex.what();
     return false;
@@ -299,18 +299,18 @@ bool DecodeThriftStruct(void* data, const size_t len, T* obj) {
 }
 
 bool DeserializeKafkaPayload(
-    void* kafka_payload,
+    const void* kafka_payload,
     const size_t payload_len,
-    std::unique_ptr<admin::KafkaOperationCode>& op_code,
-    std::unique_ptr<rocksdb::Slice>& value) {
+    admin::KafkaOperationCode* op_code,
+    std::string* value) {
   admin::KafkaMessagePayload msg_payload;
 
-  if (DecodeThriftStruct<admin::KafkaMessagePayload>(kafka_payload,
-      payload_len, &msg_payload)) {
-    op_code = std::make_unique<admin::KafkaOperationCode>(msg_payload.op_code);
+  const std::string payload(static_cast<const char *>(kafka_payload),
+                            payload_len);
+  if (DecodeThriftStruct<admin::KafkaMessagePayload>(payload, &msg_payload)) {
+    *op_code = msg_payload.op_code;
     if (msg_payload.__isset.value) {
-      value = std::make_unique<rocksdb::Slice>(
-          rocksdb::Slice(msg_payload.value->moveToFbString().toStdString()));
+      *value = std::move(msg_payload.value);
     }
     return true;
   } else {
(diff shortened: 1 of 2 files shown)


rajathprasad:
> will review it once we have a concrete plan for how to roll it out.

We decided to have a flag in the zk_config (meta_db) to decide whether deserialization is necessary. So we can keep the current use case as is. Future use cases which need specialized functionality (like delete, merge) will be able to set that flag and produce messages in the thrift format. 

I will update this diff with the implementation part, so we can have a discussion around the entire solution.

rocksdb_admin/admin_handler.cpp
```diff
@@ -377,6 +393,24 @@ bool AdminHandler::writeMetaData(
   return s.ok();
 }
 
+bool DeserializeKafkaPayload(
```
newpoo: move this function to the anonymous namespace in this file.

rocksdb_admin/admin_handler.cpp
```diff
@@ -377,6 +393,24 @@ bool AdminHandler::writeMetaData(
...
+    const string& payload,
+    KafkaOperationCode& op_code,
+    rocksdb::Slice& value) {
```
newpoo: for output parameter, we usually use pointers instead of references. This will make it more explicit that the function can change the parameters.

rocksdb_admin/admin_handler.cpp
```diff
@@ -377,6 +393,24 @@ bool AdminHandler::writeMetaData(
...
+  KafkaMessagePayload msg_payload;
+
+  if (DecodeThriftStruct<KafkaMessagePayload>(payload, &msg_payload)) {
+    op_code = msg_payload.op_code;
+    if (msg_payload.__isset.value) {
+      value = rocksdb::Slice(msg_payload.value);
```
newpoo: msg_payload.value is on the stack of function DeserializeKafkaPayload, and value internally keeps a reference to it. After returning from DeserializeKafkaPayload(), the content referenced by value will be invalid.

rocksdb_admin/admin_handler.cpp
```diff
@@ -1063,20 +1097,22 @@ void AdminHandler::async_tm_startMessageIngestion(
   }
 
   int64_t message_count = 0;
+  const auto should_deserialize = meta.should_deserialize_kafka_payload;
```
newpoo: where do we set meta.should_deserialize_kafka_payload?

Will it be simpler if we remove should_deserialize_kafka_payload and use a gflag instead?

rocksdb_admin/admin_handler.cpp
```diff
@@ -1095,18 +1131,50 @@ void AdminHandler::async_tm_startMessageIngestion(
...
-      // TODO: if there are transient errors, add a retry
-      LOG(ERROR) << "Failure while writing to " << db_name << ": "
-                 << status.ToString();
-      common::Stats::get()->Incr(kKafkaDbWriteErrors);
+
+    // Deserialize the kafka payload
+    KafkaOperationCode op_code;
+    rocksdb::Slice value;
+    if (should_deserialize) {
+      if (!DeserializeKafkaPayload(payload, op_code, value)) {
```
newpoo: If should_deserialize, we can avoid constructing payload by changing DeserializeKafkaPayload() interface to take message->payload() and message->len().

rocksdb_admin/admin_handler.cpp
```diff
@@ -281,6 +285,18 @@ bool SetAddressOrException(const std::string& ip,
   }
 }
 
+template <typename T>
+bool DecodeThriftStruct(const string& data, T* obj) {
+  auto ex = folly::try_and_catch<std::exception>([data, obj]() {
+    apache::thrift::CompactSerializer::deserialize(
```
newpoo: We may use BinarySerializer, I remember there were some compatible issues between Java and CPP for Compact protocol.

We may call deserialize() directly inside a try() catch {}.

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -170,6 +171,16 @@ struct StopMessageIngestionResponse {
...
 }
 
+enum KafkaOperationCode {
+  PUT = 1,
+  DELETE = 2
+}
+
+struct KafkaMessagePayload {
+  1: required KafkaOperationCode op_code,
+  2: optional binary value
```
newpoo: To optimize it, we may use std::unique_ptr<IOBuf> instead of binary.

There are some examples in cosmos that demonstrate how to do it.

rocksdb_admin/admin_handler.cpp
rajathprasad: This will be set in the upload script.

There are already too many gflags and I personally prefer segment level configs to be present in the meta_db.

rocksdb_admin/admin_handler.cpp
rajathprasad: Done.
For my knowledge, why is try-catch better than the folly function?

rocksdb_admin/admin_handler.cpp
rajathprasad: I changed it to reference of unique_ptr since I'm allocating  new object in this function.

rocksdb_admin/admin_handler.cpp
rajathprasad: Good point. Not sure why it didn't cause a seg fault when I tested.

rocksdb_admin/admin_handler.cpp
```diff
@@ -281,6 +286,39 @@ bool SetAddressOrException(const std::string& ip,
...
+    return false;
+  }
+  return true;
+}
+
+bool DeserializeKafkaPayload(
+    void* kafka_payload,
+    const size_t payload_len,
+    std::unique_ptr<admin::KafkaOperationCode>& op_code,
+    std::unique_ptr<rocksdb::Slice>& value) {
```
newpoo: std::unique_ptr<IOBuf>* value

rocksdb_admin/admin_handler.cpp
```diff
@@ -281,6 +286,39 @@ bool SetAddressOrException(const std::string& ip,
...
+  admin::KafkaMessagePayload msg_payload;
+
+  if (DecodeThriftStruct<admin::KafkaMessagePayload>(kafka_payload,
+      payload_len, &msg_payload)) {
+    op_code = std::make_unique<admin::KafkaOperationCode>(msg_payload.op_code);
```
newpoo: *op_code = msg_payload.op_code

rocksdb_admin/admin_handler.cpp
```diff
@@ -1095,18 +1136,55 @@ void AdminHandler::async_tm_startMessageIngestion(
...
-      // TODO: if there are transient errors, add a retry
-      LOG(ERROR) << "Failure while writing to " << db_name << ": "
-                 << status.ToString();
-      common::Stats::get()->Incr(kKafkaDbWriteErrors);
+
+    // Deserialize the kafka payload
+    KafkaOperationCode op_code;
+    rocksdb::Slice value;
+    if (should_deserialize) {
+      std::string deser_val;
```
newpoo: move it out side if() {}, otherwise value will be reference to invalid content when outside the if block.

rocksdb_admin/admin_handler.cpp
```diff
@@ -281,6 +286,39 @@ bool SetAddressOrException(const std::string& ip,
...
+}
+
+bool DeserializeKafkaPayload(
+    const void* kafka_payload,
+    const size_t payload_len,
+    admin::KafkaOperationCode* op_code,
+    std::string* value) {
+  admin::KafkaMessagePayload msg_payload;
+
+  const std::string payload(static_cast<const char *>(kafka_payload),
```
newpoo: can we change the interface of DecodeThriftStruct() to take void* and a length to avoid make a copy of the data?

rajathprasad:
> lgtm

@newpoo thanks for the detailed review.

PR #321, opened 2019-09-24
[helix_client] added functionality for helix_client to disconnect Paticipant from helixManager

cpd85:
Tested via private build on canary, here is sample log output (will update when I change the code per Bo's comments)
16:28:40.026 [Thread-9] ERROR rest.rocksplicator.Participant - Signal to main
16:28:40.026 [main] ERROR rest.rocksplicator.Participant - Signal received, unlock the lock
16:28:40.027 [main] ERROR rest.rocksplicator.Participant - Disconnect from helixManager
E0924 16:28:40.051600 17892 helix_client.cpp:113] Participant main() exited

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
@@ -18,6 +18,9 @@
 
 package com.pinterest.rocksplicator;
 
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.HashMap;
 import java.util.Map;
 import org.apache.commons.cli.CommandLine;
@@ -59,8 +62,10 @@ public class Participant {
   private static final String stateModel = "stateModelType";
   private static final String configPostUrl = "configPostUrl";
   private static final String disableSpectator = "disableSpectator";
+  private static final Lock lock = new ReentrantLock();
+  private static final Condition condition = lock.newCondition();
 
-  private HelixManager helixManager;
+  private static HelixManager helixManager;
   private StateModelFactory<StateModel> stateModelFactory;
 
   private static Options constructCommandLineOptions() {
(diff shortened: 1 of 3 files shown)


rocksdb_admin/helix_client.cpp
```diff
@@ -117,10 +117,10 @@ void invokeClass(JNIEnv* env,
 
   // The participant main function should never exit
   env->ExceptionDescribe();
-  CHECK(false) << "Participant main() exited";
```
newpoo: Would it be simpler to call  helixManager.disconnect() in ShutdownJVM() directly?
We can keep the CHECK() here and let it fail loudly if the participant fails to join helix cluster on startup.

cpd85: Great suggestion, thanks!

cpd85:
E0925 19:10:53.160398  1642 helix_client.cpp:171] EDATCHED, trying to attach thread
19:10:53.160 [Thread-14] ERROR rest.rocksplicator.Participant - Disconnect the helixManager

logs from the latest commit

rocksdb_admin/helix_client.cpp
```diff
@@ -117,10 +117,10 @@ void invokeClass(JNIEnv* env,
...
+  LOG(ERROR) << "Participant main() exited";
```
newpoo: revert this change?

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -60,7 +60,7 @@
   private static final String configPostUrl = "configPostUrl";
   private static final String disableSpectator = "disableSpectator";
 
-  private HelixManager helixManager;
+  private static HelixManager helixManager;
```
newpoo: Now we initialize a static data member in the constructor which is not ideal. I am not sure if there is a way to not allow the constructor gets called twice. One thing we may do is to make the Participant() constructor private. So we won't initialize helixManager more than once if we don't call main() twice.

cpd85: i can also check for isConnected() in the constructor if you think that might help as well, and only initialize if isConnected() is false

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -163,10 +163,17 @@ public static void main(String[] args) throws Exception {
...
     helixAdmin.setConfig(scope, properties);
 
+    LOG.error("Participant running");
     Thread.currentThread().join();
   }
 
-  public Participant(String zkConnectString, String clusterName, String instanceName,
+  public static void disconnectHelixManager() throws Exception {
+    LOG.error("Disconnect the helixManager");
+    helixManager.disconnect();
```
newpoo: one last nit: check if (helixManager) before calling disconnect().

PR #324, opened 2019-10-09
Revert "Enable spectator to spectate multiple clusters (#312)"

cpd85:
This reverts commit 2659f5223e1ea11939bc3ae9409a80a57ea86651. For now, we prefer to launch a new cluster to spectate a corresponding rocksplicator cluster


newpoo:
Can you please test it and roll it out to the 2 spectator clusters that we have.

PR #332, opened 2019-10-22
Add segment tags to kafka consumer metrics.

rajathprasad:
This is to differentiate metrics for different segments in the same multi-tenant
cluster.
Bonus: add a metric for invalid op_code ingestion


rocksdb_admin/admin_handler.cpp
```diff
@@ -1109,7 +1111,7 @@ void AdminHandler::async_tm_startMessageIngestion(
   // live messages.
   kafka_watcher->StartWith(
       replay_timestamp_ms,
-      [message_count, db_name, db, should_deserialize, this](
+      [message_count, db_name, db, should_deserialize, segment, this](
```
cpd85: nit : can we std::move(segment)?

PR #334, opened 2019-10-30
added a NOOP replication db when replication should not be performed

cpd85:
For kafka in rocksplicator use cases, we have been noticing that there is some replication latency stats reported. Since the WAL is used when we merge the kafka messages, and by default we add a DB as a slave with itself as upstream, the DBs on a kafka in rocksplicator host will try to periodically pull updates from themselves. With OnlineOffline use cases, the global seqno is never changed, so we don't run into this issue.

The call stack that produces this self replication is

AdminHandler::async_tm_addDB() calls ApplicationDBManager::addDB()
ApplicationDBManager::addDB() calls ApplicationDB::ApplicationDB() 
ApplicationDB::ApplicationDB() constructor calls addDB if (!IsSlave() || upstream_addr_)
RocksDBReplicator::addDB() calls pullFromUpstream() in line 90 if (db::role == SLAVE)	

Thus adding db::role = NOOP will prevent thiss from happening. The other option is to not set the upstream address, but I feel that this reads more cleanly when we explicitly tell rocksplicator not to replicate

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -473,13 +473,13 @@ void AdminHandler::async_tm_addDB(
   if (request->__isset.db_role) {
     if (request->db_role == "SLAVE") {
       role = replicator::DBRole::SLAVE;
-    } else if (request->db_role == "MASTER") {
-      role = replicator::DBRole::MASTER;
     } else if (request->db_role == "NOOP") {
       role = replicator::DBRole::NOOP;
     } else {
-      // Default behavior is to use SLAVE, so keep that
-      LOG(INFO) << "Invalid role requested to addDB " << request->db_role;
+      e.errorCode = AdminErrorCode::INVALID_DB_ROLE;
+      e.message = std::move(request->db_role);
+      callback.release()->exceptionInThread(std::move(e));
+      return;
     }
   }
 
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -469,9 +469,23 @@ void AdminHandler::async_tm_addDB(
 
   // add the db to db_manager
   std::string err_msg;
+  replicator::DBRole role = replicator::DBRole::SLAVE;
+  if (request->__isset.db_role) {
+    if (request->db_role == "SLAVE") {
+      role = replicator::DBRole::SLAVE;
+    } else if (request->db_role == "MASTER") {
+      role = replicator::DBRole::MASTER;
+    } else if (request->db_role == "NOOP") {
+      role = replicator::DBRole::NOOP;
+    } else {
+      // Default behavior is to use SLAVE, so keep that
+      LOG(INFO) << "Invalid role requested to addDB " << request->db_role;
+    }
+  }
+
   if (!db_manager_->addDB(request->db_name,
                           std::unique_ptr<rocksdb::DB>(rocksdb_db),
-                          replicator::DBRole::SLAVE, std::move(upstream_addr),
+                          role, std::move(upstream_addr),
                           &err_msg)) {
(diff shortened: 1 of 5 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -469,9 +469,23 @@ void AdminHandler::async_tm_addDB(
...
+  if (request->__isset.db_role) {
+    if (request->db_role == "SLAVE") {
+      role = replicator::DBRole::SLAVE;
+    } else if (request->db_role == "MASTER") {
+      role = replicator::DBRole::MASTER;
+    } else if (request->db_role == "NOOP") {
+      role = replicator::DBRole::NOOP;
+    } else {
+      // Default behavior is to use SLAVE, so keep that
+      LOG(INFO) << "Invalid role requested to addDB " << request->db_role;
```
rajathprasad: Nit: it's cleaner IMO to use a switch case (but a personal preference)

rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -47,6 +47,8 @@ struct AddDBRequest {
   2: required string upstream_ip,
   # if overwrite is true, destroy any rocksdb instance under db_name
   3: optional bool overwrite = false,
+  # if set, add the db to the db_manager with the specified role. one of MASTER, SLAVE, NOOP
+  4: optional string db_role = "SLAVE",
```
rajathprasad: Would it be better to instead add a boolean argument `should_replicate` and translate that to a NOOP DBRole? This way we won't have to expose the concept of DBRole in the API.

cpd85: it's exposed in the api here already : https://github.com/pinterest/rocksplicator/blob/master/rocksdb_admin/rocksdb_admin.thrift#L114

Was wondering if in the future there would be a use case to add a master later, or if we should keep it consistent with further up in the thrift file. But still happy to change if you think so!

rocksdb_admin/admin_handler.cpp
cpd85: c++ doesn't support switch on a string : https://stackoverflow.com/questions/650162/why-the-switch-statement-cannot-be-applied-on-strings

I was thinking of just exposing the db role enum, but if we decide to use a bool instead we can also get rid of this

rocksdb_replicator/rocksdb_replicator.h
```diff
@@ -63,6 +63,7 @@ struct LogExtractor : public rocksdb::WriteBatch::Handler {
 enum class DBRole {
   MASTER,
   SLAVE,
+  NOOP, // Don't perform any replication with this DB
```
newpoo: Can we just use MASTER for this use case?

rocksdb_admin/admin_handler.cpp
newpoo: Can we set the role default to be SLAVE, and only change it to be NOOP if db_role in the request is NOOP. Just wanted to minimize the changes to the code paths of other use cases.

cpd85: sounds good, just removed ability to set master

rocksdb_replicator/rocksdb_replicator.h
cpd85: prefer NOOP because I personally think it is more readable/understandable to mark the db as not meant for replication, but if you prefer MASTER I think it should work. I'm not 100% confident on the code and if it will try to push updates out to slaves, but willing to just go with MASTER if you prefer

newpoo: Master won't push data to slave. It only reactively answer request from slaves. I am ok to go with NOOP, as it can help minimize changes to other use cases.

PR #337, opened 2019-12-19
[s3_env] Add s3_env head file to rocksplicator

ghanpipi:
The env_s3.h is the s3 environment for rocksdb, and the class S3Env overrides
all the methods in rocksdb/env.h, and the methods body will in another patch.
The common/sys_time.h is a copy of rocksdb/port/sys_time.h.

### DIFF
--- a/common/rocksdb_env_s3.h
+++ b/common/rocksdb_env_s3.h
@@ -1,0 +1,187 @@
+/// Copyright 2018 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author ghan (ghan@pinterest.com)
+//
+
+#pragma once
+
+#include <algorithm>
+#include <stdio.h>
+#include <time.h>
+#include <iostream>
+#include <mutex>
+#include <pthread.h>
+
+#include "common/s3util.h"
+#include "common/sys_time.h"
+#include "rocksdb/env.h"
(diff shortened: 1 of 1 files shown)


common/env_s3.h
```diff
@@ -0,0 +1,198 @@
...
+
+using rocksdb::SequentialFile;
+using rocksdb::EnvOptions;
+using rocksdb::WritableFile;
+using rocksdb::Directory;
+using rocksdb::FileLock;
+using rocksdb::Logger;
+using rocksdb::Status;
+using rocksdb::Env;
+using rocksdb::RandomAccessFile;
```
newpoo: Usually it's discouraged to do this in header file to avoid polluting namespace.

common/sys_time.h
```diff
@@ -0,0 +1,45 @@
...
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+//
+// Copyright (c) 2011 The LevelDB Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file. See the AUTHORS file for names of contributors.
+
+// This file is a portable substitute for sys/time.h which does not exist on
+// Windows
```
newpoo: Can you please share why we need this file?

common/env_s3.h
```diff
@@ -0,0 +1,198 @@
...
+#include <iostream>
+#include <mutex>
+#include <pthread.h>
+
+#include "common/s3util.h"
+#include "common/sys_time.h"
+#include "rocksdb/env.h"
+#include "rocksdb/status.h"
+
+namespace common {
```
newpoo: does it make more sense to put it in rocksdb namespace? It may help reduce some code in the implementation code later.

common/env_s3.h
```diff
@@ -0,0 +1,198 @@
+/// Copyright 2018 Pinterest Inc.
```
newpoo: can we rename it to rocksdb_env_s3.h?

common/sys_time.h
ghanpipi: it maybe unnecessary in the body implementation. will remove it.

common/env_s3.h
ghanpipi: good point

PR #338, opened 2020-01-06
[rocksdb_env_s3] rocksdb s3 enviroment methods body implementation

ghanpipi:
1. We will rely on the local posix env for file opeartions: read, write,
creating, deletion etc
2. For read operations, we will first copy the files from S3 to local then
perform the read via posix_env
3. For write operations, we will first perform the write in the local tmp files,
and then copy them to s3 before closing. Due to the this s3 env will be only
used for storing backups, so this behaviour should be safe.

### DIFF
--- a/common/rocksdb_env_s3.cpp
+++ b/common/rocksdb_env_s3.cpp
@@ -1,0 +1,457 @@
+/// Copyright 2018 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author ghan (ghan@pinterest.com)
+//
+
+#include "common/rocksdb_env_s3.h"
+
+#include <string>
+
+#include "boost/filesystem.hpp"
+#include "common/rocksdb_glogger/rocksdb_glogger.h"
+#include "gflags/gflags.h"
+#include "glog/logging.h"
+#include "rocksdb/status.h"
+
+DECLARE_bool(s3_direct_io);
+
+namespace rocksdb {
+
+// open a file for sequential reading
+Status S3Env::NewSequentialFile(const std::string& fname,
+                                std::unique_ptr<SequentialFile>* result,
+                                const EnvOptions& options) {
+  assert(s3_util_ != nullptr);
+  result->reset();
+
+  // We read first from local storage and then from S3.
+  auto local_full_path = local_directory_ + GetRelativePath(fname);
+  auto st = posix_env_->NewSequentialFile(local_full_path, result, options);
+
+  if (!st.ok()) {
+    // If file doesnt exist in local, we copy the file to the local storage from S3
+    size_t slash = local_full_path.find_last_of('/');
+    assert(slash != std::string::npos);
+    boost::system::error_code create_err;
+    boost::filesystem::create_directories(local_full_path.substr(0, slash), create_err);
+    if (create_err) {
+      LOG(ERROR) << "Cant create dir in local: " << local_full_path;
+      return Status::IOError();
+    }
+
+    auto resp = s3_util_->getObject(fname, local_full_path, FLAGS_s3_direct_io);
+    if (!resp.Error().empty()) {
+      LOG(ERROR) << "Error happened when copying the file from S3 to local: "
+                 << resp.Error();
+      return Status::IOError();
+    }
+  }
+
+  // we successfully copied the file, try opening it locally now
+  return posix_env_->NewSequentialFile(local_full_path, result, options);
+}
+
+// open a file for random reading
+Status S3Env::NewRandomAccessFile(const std::string& fname,
+                                  std::unique_ptr<RandomAccessFile>* result,
+                                  const EnvOptions& options) {
+  assert(s3_util_ != nullptr);
+  result->reset();
+
+  // We read first from local storage and then from S3.
+  auto local_full_path = local_directory_ + GetRelativePath(fname);
+  auto st = posix_env_->NewRandomAccessFile(local_full_path, result, options);
+
+  if (!st.ok()) {
+    // If file doesnt exist in local, we copy the file to the local storage from S3
+    size_t slash = local_full_path.find_last_of('/');
+    assert(slash != std::string::npos);
+    boost::system::error_code create_err;
+    boost::filesystem::create_directories(local_full_path.substr(0, slash), create_err);
+    if (create_err) {
+      LOG(ERROR) << "Cant create dir in local: " << local_full_path;
+      return Status::IOError();
+    }
+
+    auto resp = s3_util_->getObject(fname, local_full_path, FLAGS_s3_direct_io);
+    if (!resp.Error().empty()) {
+      LOG(ERROR) << "Error happened when copying the file from S3 to local: "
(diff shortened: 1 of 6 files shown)


common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+  // we successfully copied the file, try opening it locally now
+  st = posix_env_->NewRandomAccessFile(local_full_path, result, options);
+
+  return st;
+}
+
+// S3WritableFile is a wrapper for writing a file in S3. We will perform the
+// operations in a local tmp file and upload to S3 when closing the file.
+class S3WritableFile : public WritableFile {
+ private:
```
rajathprasad: Convention is to put the `public` section first, followed by `protected` and then finally `private`

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+                 std::shared_ptr<common::S3Util> s3_util) :
+      fname_(local_fname),
+      s3_fname_(s3_fname),
+      env_(env),
+      s3_util_(s3_util) {
+    auto local_env = env_->GetBaseEnv();
+    auto s = local_env->NewWritableFile(local_fname, &local_file_, options);
+    if (!s.ok()) {
+      LOG(ERROR) << "Error happened when creating writable file in local: " << s.ToString();
+      status_ = s;
```
rajathprasad: Any reason for setting a status in the constructor? It's odd for the caller to check status as soon as calling the constructor (and the caller might forget to do so). Would it be better to do this in a separate function?

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+namespace rocksdb {
+
+// open a file for sequential reading
+Status S3Env::NewSequentialFile(const std::string& fname,
+                                std::unique_ptr<SequentialFile>* result,
+                                const EnvOptions& options) {
+  assert(s3_util_ != nullptr);
+  result->reset();
+
+  // We read first from local storage and then from S3.
```
rajathprasad: Could the file have changed in s3 but not locally?

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  return Status::OK();;
+}
+
+Status S3Env::GetChildren(const std::string& path,
+                          std::vector<std::string>* result) {
+  auto s3_full_path = s3_key_prefix_ + path;
+  // first fetch all files using the given path as the key prefix in S3
+  auto resp = s3_util_->listObjects(s3_full_path);
```
rajathprasad: If `path` does not exist, how is it handled?

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+  // trim the file name
+  for (auto& v : resp.Body()) {
+    // Our path should be a prefix of the fetched value
+    if (v.find(s3_full_path) == 0) {
+      result->push_back(v.substr(s3_full_path.size()));
+    }
+  }
+
+  // then fetch all files that exist in the local posix directory
```
rajathprasad: If you have already read from s3, what is the purpose of reading from local dir? Is it for cases when changes have been made locally but the file has not been closed?

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+    }
+  }
+
+  return st;
+}
+
+// The rename is not atomic. S3 does not support renaming natively.
+// Copy file to a new object in S3 and actually only IDENTITY files could be renamed.
+Status S3Env::RenameFile(const std::string& src, const std::string& target) {
+  auto resp = s3_util_->putObject(s3_key_prefix_ + target, local_directory_ + src);
```
rajathprasad: Do we need to delete the older file?

common/s3util.cpp
```diff
@@ -312,6 +314,22 @@ GetObjectMetadataResponse S3Util::getObjectMetadata(const string &key) {
   return GetObjectMetadataResponse(std::move(metadata), "");
 }
 
+GetObjectSizeAndModTimeResponse S3Util::getObjectSizeAndModTime(const string& key) {
+  HeadObjectRequest headObjectRequest;
+  headObjectRequest.SetBucket(bucket_);
+  headObjectRequest.SetKey(key);
+  map<string, uint64_t> metadata;
+  auto headObjectOutcome = s3Client->HeadObject(headObjectRequest);
```
newpoo: how is HeadObject() different from the MakeHttpRequest() call from line 297? Can we reuse getObjectMetadata()?

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+Status S3Env::LockFile(const std::string& fname, FileLock** lock) {
+  // there isn's a very good way to atomically check and create
+  // a file via libs3
+  *lock = nullptr;
+  return Status::OK();
+}
+
+Status S3Env::UnlockFile(FileLock* lock) {
+  return Status::OK();
+}
```
newpoo: what's the semantics of these functions? can you please add some comments for it. Does rocksdb backup/restore need to use these two functions?

common/rocksdb_env_s3.h
```diff
@@ -54,7 +53,12 @@ class S3FatalException : public std::exception {
 class S3Env : public Env {
```
newpoo: Can you document the design, assumption (and why), and limitation (and why) of the implementation?

common/rocksdb_env_s3.cpp
```diff
@@ -0,0 +1,432 @@
...
+  auto local_full_path = local_directory_ + fname;
+  auto st = posix_env_->NewSequentialFile(local_full_path, result, options);
+
+  if (!st.ok()) {
+    // If file doesnt exist in local, we copy the file to the local storage from S3
+    auto resp = s3_util_->getObject(s3_key_prefix_ + fname, local_full_path, FLAGS_s3_direct_io);
+    if (!resp.Error().empty()) {
+      LOG(ERROR) << "Error happened when copying the file from S3 to local: "
+                 << resp.Error();
+      return Status::IOError();
```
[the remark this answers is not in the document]
ghanpipi: the rocksdb will first check the status then read the value under result

common/rocksdb_env_s3.cpp
ghanpipi: good catch!

common/rocksdb_env_s3.cpp
ghanpipi: all the files used in the bakcup/restore are immutable.

common/rocksdb_env_s3.cpp
ghanpipi: they the rocksdb backup engine wont take any operation under this path

common/rocksdb_env_s3.cpp
ghanpipi: yes. we wont need to read them from local again. I already fixed it

common/rocksdb_env_s3.cpp
ghanpipi: fixed in the new patch

common/s3util.cpp
ghanpipi: some infomation like the ModTime was not exposed in the MakeHttpRequest() API

common/rocksdb_env_s3.h
ghanpipi: sure. Added more details.

common/s3util.cpp
newpoo: Is "LastModified" on https://docs.aws.amazon.com/cli/latest/reference/s3api/head-object.html what we need?

common/rocksdb_env_s3.cpp
ghanpipi: I followed the convention in the rocksdb hdfs implementation

common/s3util.cpp
ghanpipi: the format of LastModified by MakeHttpRequest is not easy to convert to the uint64_t format

PR #339, opened 2020-01-06
[rocksdb_s3_env] add s3_env APIs to rocksdb_admin.thrift

ghanpipi:


rocksdb_admin/rocksdb_admin.thrift
```diff
@@ -84,6 +84,39 @@ struct RestoreDBResponse {
   # for future use
 }
 
+struct BackupDBToS3Request {
+  # the db to backup
+  1: required string db_name,
+  # the s3 bucket to backup to
+  2: required string s3_bucket,
+  # the s3 key prefix to backup to
+  3: required string s3_backup_dir,
```
rajathprasad: Would be good to keep the naming convention standard between the backup and restore structs. Either keep both as `s3_backup_dir` or both as `s3_path`. Also, with these s3 paths, I'm always confused with the convention, should the path name end with a slash? Should an example be documented here or in the c++ header file?

ghanpipi: good catch. I forgot to change it to s3_backup_dir.

PR #347, opened 2020-03-12
Use s3 env for rocksdb backup & restore

ghanpipi:
1. Refactor rocksdb_admin backup/restore methods to be used for hdfs/s3 env
backup & restore APIs
2. Add flags to indicate which env(hdfs/s3) will be used for backup & restore

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -32,17 +32,15 @@
 #include "common/kafka/kafka_consumer_pool.h"
 #include "common/kafka/kafka_watcher.h"
 #include "common/network_util.h"
+#include "common/rocksdb_env_s3.h"
 #include "common/rocksdb_glogger/rocksdb_glogger.h"
 #include "common/stats/stats.h"
 #include "common/thrift_router.h"
 #include "common/timeutil.h"
 #include "folly/FileUtil.h"
-#include "folly/MoveWrapper.h"
 #include "folly/ScopeGuard.h"
 #include "folly/String.h"
 #include "librdkafka/rdkafkacpp.h"
-#include "rocksdb/env.h"
-#include "rocksdb/status.h"
 #include "rocksdb/options.h"
 #include "rocksdb/utilities/backupable_db.h"
 #include "rocksdb_admin/detail/kafka_broker_file_watcher_manager.h"
@@ -112,6 +110,14 @@ const std::string kKafkaDbDeleteErrors = "kafka_db_delete_errors";
 const std::string kKafkaDbMergeErrors = "kafka_db_merge_errors";
 const std::string kKafkaDeserFailure = "kafka_deser_failure";
 const std::string kKafkaInvalidOpcode = "kafka_invalid_opcode";
+const std::string kHDFSBackupSuccess = "hdfs_backup_success";
+const std::string kS3BackupSuccess = "s3_backup_success";
+const std::string kHDFSBackupFailure = "hdfs_backup_failure";
+const std::string kS3BackupFailure = "s3_backup_failure";
+const std::string kHDFSRestoreSuccess = "hdfs_restore_success";
+const std::string kS3RestoreSuccess = "s3_restore_success";
+const std::string kHDFSRestoreFailure = "hdfs_restore_failure";
+const std::string kS3RestoreFailure = "s3_restore_failure";
 
 int64_t GetMessageTimestampSecs(const RdKafka::Message& message) {
   const auto ts = message.timestamp();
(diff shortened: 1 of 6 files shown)


rocksdb_admin/admin_handler.h
```diff
@@ -158,11 +170,31 @@ class AdminHandler : virtual public AdminSvIf {
...
   std::atomic<int> num_current_s3_sst_downloadings_;
+  // number of the current concurrenty s3 uploadings
+  std::atomic<int> num_current_s3_sst_uploadings_;
   // Map of db_name to kafka watcher
   std::unordered_map<std::string, std::shared_ptr<KafkaWatcher>>
     kafka_watcher_map_;
   // Lock for synchronizing access to kafka_watcher_map_
   std::mutex kafka_watcher_lock_;
+
+  bool backupDBHelper(const std::string& db_name,
```
sihanwang-pinterest: do we implement these three functions?

rocksdb_admin/helix_client.cpp
```diff
@@ -109,8 +114,12 @@ void invokeClass(JNIEnv* env,
   env->SetObjectArrayElement(args, 12, env->NewStringUTF("--configPostUrl"));
   env->SetObjectArrayElement(args, 13,
                              env->NewStringUTF(config_post_url.c_str()));
+  if (FLAGS_use_s3_backup) {
+    env->SetObjectArrayElement(args, 14, env->NewStringUTF("--s3Bucket"));
+    env->SetObjectArrayElement(args, 15, env->NewStringUTF(FLAGS_s3_bucket.c_str()));
+  }
   if (disable_spectator) {
-    env->SetObjectArrayElement(args, 14, env->NewStringUTF("--disableSpectator"));
+    env->SetObjectArrayElement(args, FLAGS_use_s3_backup ? 16 : 14, env->NewStringUTF("--disableSpectator"));
```
rajathprasad: what is the logic here? why are `disableSpectator` and `use_s3_backup` related?

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +508,314 @@ void AdminHandler::async_tm_ping(
...
+                             FLAGS_rocksdb_replicator_port,
+                             upstream_addr.get(),
+                             &callback)) {
+    common::Stats::get()->Incr(kS3RestoreFailure);
+    return;
+  }
+
+  auto local_s3_util = createLocalS3Util(request->limit_mbs, request->s3_bucket);
+  std::string formatted_s3_dir_path = rtrim(request->s3_backup_dir, '/');
+  rocksdb::Env* s3_env = new rocksdb::S3Env(formatted_s3_dir_path, local_path, std::move(local_s3_util));
```
sihanwang-pinterest: std::move(local_path)?

rocksdb_admin/admin_handler.cpp
```diff
@@ -32,17 +32,16 @@
...
 #include "common/thrift_router.h"
 #include "common/timeutil.h"
 #include "folly/FileUtil.h"
-#include "folly/MoveWrapper.h"
 #include "folly/ScopeGuard.h"
 #include "folly/String.h"
 #include "librdkafka/rdkafkacpp.h"
-#include "rocksdb/env.h"
-#include "rocksdb/status.h"
+#include "common/rocksdb_env_s3.h"
```
rajathprasad: You've included this header twice.

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +508,314 @@ void AdminHandler::async_tm_ping(
...
-  auto status = rocksdb::NewHdfsEnv(&hdfs_env, full_path);
-  if (!OKOrSetException(status,
-                        AdminErrorCode::DB_ADMIN_ERROR,
-                        &callback)) {
-    return;
+  rocksdb::BackupEngine* backup_engine;
+  auto status = rocksdb::BackupEngine::Open(
+      rocksdb::Env::Default(), options, &backup_engine);
+  if (!status.ok()) {
+    e->errorCode = AdminErrorCode::DB_ADMIN_ERROR;
```
rajathprasad: Log error.

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +508,314 @@ void AdminHandler::async_tm_ping(
...
-  if (!OKOrSetException(status,
-                        AdminErrorCode::DB_ADMIN_ERROR,
-                        &callback)) {
-    return;
+  auto status = rocksdb::BackupEngine::Open(
+      rocksdb::Env::Default(), options, &backup_engine);
+  if (!status.ok()) {
+    e->errorCode = AdminErrorCode::DB_ADMIN_ERROR;
+    e->message = status.ToString();
+    return false;
```
rajathprasad: Log error.

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +508,314 @@ void AdminHandler::async_tm_ping(
...
+
+  return str;
+}
+
+void AdminHandler::async_tm_backupDBToS3(
+    std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
+      BackupDBToS3Response>>> callback,
+    std::unique_ptr<BackupDBToS3Request> request) {
+  AdminException e;
+  auto n = num_current_s3_sst_uploadings_.fetch_add(1);
```
rajathprasad: can we make this `const auto`?

rocksdb_admin/admin_handler.cpp
```diff
@@ -888,33 +1092,11 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
-      // Invoke destructor explicitly to make sure Aws::InitAPI()
-      // and Aps::ShutdownApi() appear in pairs.
-      s3_util_ = nullptr;
-      s3_util_ = common::S3Util::BuildS3Util(request->s3_download_limit_mb,
-                                             request->s3_bucket);
-    }
-    local_s3_util = s3_util_;
-  }
-
+  auto local_s3_util = createLocalS3Util(request->s3_download_limit_mb, request->s3_bucket);
```
rajathprasad: Could this cause a bottleneck for backup or restoring? If there are quite a few downloads/uploads happening for a particular segment (with saw 'x' as the download limit), another request with 'y' as the download limit could wait for a long time to create the S3 client. Could this cause issues when helix is trying to promote a master?

rocksdb_admin/admin_handler.h
```diff
@@ -158,11 +170,31 @@ class AdminHandler : virtual public AdminSvIf {
...
+  bool restoreDBHelper(const std::string& db_name,
+                       const std::string& backup_dir,
+                       std::unique_ptr<rocksdb::Env> env_holder,
+                       std::unique_ptr<folly::SocketAddress> upstream_addr,
+                       const bool enable_restore_rate_limit,
+                       const uint32_t restore_rate_limit,
+                       AdminException* e);
+
+  std::shared_ptr<common::S3Util> createLocalS3Util(const uint32_t read_ratelimit_mb = 50,
+                                                    const std::string& bucket = "");
```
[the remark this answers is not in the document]
rajathprasad: My bad

rocksdb_admin/admin_handler.cpp
ghanpipi: good catch

rocksdb_admin/admin_handler.cpp
ghanpipi: I don't think it would cause a bottleneck. Actually we are using the same logic in the readonly cluster which has more downloading requests/bandwidth than the backup/restore use-case. It wont happen in the master promotion due to this process only happen in offline to slave state transition.

rocksdb_admin/helix_client.cpp
ghanpipi: they are not related. but we need to calculate the number of arguments passed to the jni

rocksdb_admin/helix_client.cpp
```diff
@@ -30,6 +30,10 @@ DECLARE_int32(port);
 DEFINE_string(helix_log_config_file_path, "",
     "Participant log config file path "
     "e.g: /etc/config/helix/log4j.properties");
+DEFINE_bool(use_s3_backup, false,
+            "whether we should use S3 env for rocksdb backup&rockstore");
+DEFINE_string(s3_bucket, "pinterest-jackson",
```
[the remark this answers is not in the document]
ghanpipi: we use the similar thing for hdfs env...

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +507,317 @@ void AdminHandler::async_tm_ping(
...
-                        AdminErrorCode::DB_ADMIN_ERROR,
-                        &callback)) {
+  LOG(INFO) << "HDFS Restore is done.";
+  common::Stats::get()->Incr(kHDFSRestoreSuccess);
+  callback->result(RestoreDBResponse());
+}
+
+inline std::string rtrim(std::string str, char c) {
+  if (str.length() > 0 && str.back() == c) {
+    str.substr(0, str.size() - 1) ;
```
newpoo: substr() won't change the content of str.

can we do the following:

`if (!str.empty() && str.back() == c) {
    str.pop_back();
}`

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +507,317 @@ void AdminHandler::async_tm_ping(
...
-    std::unique_ptr<BackupDBRequest> request) {
-  db_admin_lock_.Lock(request->db_name);
-  SCOPE_EXIT { db_admin_lock_.Unlock(request->db_name); };
+bool AdminHandler::backupDBHelper(const std::string& db_name,
+                                  const std::string& backup_dir,
+                                  std::unique_ptr<rocksdb::Env> env_holder,
+                                  const bool enable_backup_rate_limit,
+                                  const uint32_t backup_rate_limit,
+                                  AdminException* e) {
+  assert(env_holder != nullptr);
```
newpoo: can you use CHECK() instead of assert()?

rocksdb_admin/admin_handler.cpp
```diff
@@ -501,147 +507,317 @@ void AdminHandler::async_tm_ping(
...
 
 void AdminHandler::async_tm_restoreDB(
     std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
       RestoreDBResponse>>> callback,
     std::unique_ptr<RestoreDBRequest> request) {
-  db_admin_lock_.Lock(request->db_name);
-  SCOPE_EXIT { db_admin_lock_.Unlock(request->db_name); };
-
-  AdminException e;
+  auto full_path = FLAGS_hdfs_name_node + request->hdfs_backup_dir;
```
newpoo: can you move it to right before where it is used?

rocksdb_admin/admin_handler.cpp
newpoo: This is an issue that we may need to improve in the future. Though slave->master transition doesn't need s3 client, it may be blocked by offline->slave transitions for other DBs. Because they need to compete for helix concurrency, admin threads, etc.

ghanpipi: yes. we will optimize it in the future when it becomes the performance bottleneck.

rocksdb_admin/admin_handler.cpp
```diff
@@ -230,7 +236,7 @@ std::unique_ptr<::admin::ApplicationDBManager> CreateDBBasedOnConfig(
          upstream_addr = folly::makeMoveWrapper(std::move(upstream_addr)),
          my_role, &db_manager] () mutable {
           std::string err_msg;
-          auto db = std::move(*db_future).get();
+          auto db = (*db_future).get();
```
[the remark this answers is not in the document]
ghanpipi: maybe i forgot to rebase.

rocksdb_admin/admin_handler.cpp
ghanpipi: good catch!

PR #351, opened 2020-03-19
Setup settings.xml to handle maven https

fishcakez:
So the java CI can pass


fishcakez:
@ghanpipi now it works 🎉

newpoo:
Thanks for helping fix it!

PR #354, opened 2020-03-27
[cluster management] enable participant to register to multiple state models

kangnanli:
Currently, Participant can only register to 1 state model (e.g. MasterSlave or OnlineOffline). Change the logic to support register to multiple state models. 

Example: pass the "stateModelType" as "MasterSlave;Task", the cluster will register to "MasterSlave" and "Task".  This prepare Rocksplicator support both db management (by MasterSlave state model) and Backup/Restore (by Task state model). 
Other usage, register to "MasterSlave;OnlineOffline" is also allowed, but not recommended since better practice to separate realtime mode (ie. MasterSlave) from batch mode (ie OnlineOffline).

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
@@ -41,6 +41,8 @@ import org.apache.helix.participant.HelixCustomCodeRunner;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.participant.statemachine.StateModel;
 import org.apache.helix.participant.statemachine.StateModelFactory;
+import org.apache.helix.task.TaskFactory;
+import org.apache.helix.task.TaskStateModelFactory;
 import org.apache.log4j.BasicConfigurator;
 import org.apache.log4j.ConsoleAppender;
 import org.apache.log4j.Level;
@@ -187,12 +189,13 @@ public class Participant {
     }
   }
 
-
   private Participant(String zkConnectString, String clusterName, String instanceName,
-                     String stateModelType, int port, String postUrl,
-                     boolean useS3Backup, String s3BucketName, boolean runSpectator) throws Exception {
+                      String stateModelType, int port, String postUrl, boolean useS3Backup,
+                      String s3BucketName, boolean runSpectator) throws Exception {
     helixManager = HelixManagerFactory.getZKHelixManager(clusterName, instanceName,
         InstanceType.PARTICIPANT, zkConnectString);
+
+    Map<String, TaskFactory> taskFactoryRegistry = new HashMap<>();
 
     if (stateModelType.equals("OnlineOffline")) {
       stateModelFactory = new OnlineOfflineStateModelFactory(port, zkConnectString, clusterName);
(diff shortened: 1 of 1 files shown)


buildinglogic:
> Can we just add a new "else if" section for "MasterSlave;Task"? This can avoid misuse and minimize changes to other models.

Two new combinations will be used for now: 
- MasterSlave;Task.   // to support db management and backup/restore
- Task.      // support export (dedup). This cluster will only use "Task" framework from helix. 

I can change to 2 new elseif to support these?

newpoo:
Sounds good to me to have 2 new elseif.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -204,12 +215,43 @@ private Participant(String zkConnectString, String clusterName, String instanceN
...
+          port, zkConnectString, clusterName, useS3Backup, s3BucketName);
+
+      Map<String, TaskFactory> taskFactoryRegistry = new HashMap<>();
+      // TODO: register backup/restore factories
+      // taskFactoryRegistry.put("Backup", new BackupTaskFactory());
+      // taskFactoryRegistry.put("Restore", new RestoreTaskFactory());
+
+      TaskStateModelFactory helixTaskFactory =
+          new TaskStateModelFactory(helixManager, taskFactoryRegistry);
+      stateMach.registerStateModelFactory("Task", helixTaskFactory);
```
newpoo: Can we just create and fill taskFactoryRegistry here and move registerStateModelFactory "Task" to be after? This will help remove code duplication.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -187,13 +191,20 @@ public static void disconnectHelixManager() throws Exception {
...
-                     String stateModelType, int port, String postUrl,
-                     boolean useS3Backup, String s3BucketName, boolean runSpectator) throws Exception {
+  public Participant(String zkConnectString, String clusterName, String instanceName,
+                      String stateModelType, int port, String postUrl,
+                      boolean useS3Backup, String s3BucketName, boolean runSpectator)
+      throws Exception {
     helixManager = HelixManagerFactory.getZKHelixManager(clusterName, instanceName,
         InstanceType.PARTICIPANT, zkConnectString);
 
+    StateMachineEngine stateMach = helixManager.getStateMachineEngine();
```
newpoo: If we do the suggested changes below, we should move this to be after the following if else.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -187,13 +191,20 @@ public static void disconnectHelixManager() throws Exception {
...
+  }
+
 
-  private Participant(String zkConnectString, String clusterName, String instanceName,
-                     String stateModelType, int port, String postUrl,
-                     boolean useS3Backup, String s3BucketName, boolean runSpectator) throws Exception {
+  public Participant(String zkConnectString, String clusterName, String instanceName,
+                      String stateModelType, int port, String postUrl,
+                      boolean useS3Backup, String s3BucketName, boolean runSpectator)
+      throws Exception {
```
newpoo: why these changes?

kangnanli: I made the constructor to be public for the convenience of my unit testing of verifying registering to multiple modes. 
I removed the unit testing, seem to be unnecessary; thus, change back to "private".

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -50,6 +53,7 @@
 
 
 public class Participant {
+
```
newpoo: nit: remove this line.

PR #355, opened 2020-03-27
[cluster management] refactor configGenerator.java to Generator.java in or…

kangnanli:
currently: ConfigGenerator is triggered at "external_view" and "config" change in {@link HelixCustomCodeRunner} on Spectator. 
Here: we abstract the logic into Generator.java, so, ConfigGenerator can extend Generator to generate shard map. Later on, JobStateGenerator will also extend Generator to generate job state map.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
@@ -19,54 +19,26 @@
 package com.pinterest.rocksplicator;
 
 import org.apache.helix.api.listeners.PreFetch;
-import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixManager;
 import org.apache.helix.NotificationContext;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.InstanceConfig;
-import org.apache.helix.participant.CustomCodeCallbackHandler;
-import org.apache.helix.spectator.RoutingTableProvider;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.entity.StringEntity;
-import org.apache.http.impl.client.DefaultHttpClient;
-import org.json.simple.JSONArray;
-import org.json.simple.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.TimeUnit;
 
-public class ConfigGenerator extends RoutingTableProvider implements CustomCodeCallbackHandler {
+public class ConfigGenerator extends Generator {
   private static final Logger LOG = LoggerFactory.getLogger(ConfigGenerator.class);
 
-  private final String clusterName;
-  private HelixManager helixManager;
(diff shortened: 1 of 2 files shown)


newpoo:
Can you please provide more context about how we are gonna use "job state map"?

kangnanli:
> Can you please provide more context about how we are gonna use "job state map"?

Thanks for reviewing. I will add more details tmr. 

**How:**
Similar to ConfigGenerator listening to "externalView" and "Config". 
JobStateGenerator will listen to "externalView" (add others is later needed). 
**Why:**
We need keep track of status of backup or restore: 1. we will have UI to show backup status of each version (the time db is backed up); 2. our downstream operation on these backup dbs will depends on the status of these backup. 

**Implementation details**
job is treated as helix resource (similar to our segment). each job is like, "backup segmentA". 
 - When job is created/beforeRunning, externalView will change, jobStateGenerator will generate a map: 
resourceName -> {resourceVersion -> {jobVersion -> jobState}}
// resourceName is segment; resourceVersion is the time job is created, which equals to jobVersion; jobState will be NOT_STARTED
// we need jobVersion, bcz, for "restore", jobVersion != resourceVersion; jobVersion is timestamp restoreJob created, resourceVersion is the timestamp resource backed up. 
-  when job is running/beforeCompletedorFailed, externalView will change, jobStateGenerator will update the job status to RUNNING
- when job is failed due to some errors, externalView change, jobStateGenerator will update the job status to FAILED
- when job successfully back up a whole db, then, status will update to COMPLETED
by checking these status: we can prevent backuping the same segment in 2 different jobs; we can schedule downstream operation based on these status.

newpoo:
Will it be simpler if we rely on Helix Rest to poll the status?

Should we have a marker file on S3 to indicate backup finish to avoid potential S3 consistency issue caused by S3 replication.

kangnanli:
> Will it be simpler if we rely on Helix Rest to poll the status?
> 
> Should we have a marker file on S3 to indicate backup finish to avoid potential S3 consistency issue caused by S3 replication.

I would prefer to use helix task framework to get status compared to helix rest based on following analysis. 

There are few benefits to use Task Java code (ie. helix TaskDriver.java) to retrieve workflow/job status info compared to helix rest:
- from design point: by listening to externalView change, we can generate real-time job status; If pull with helix rest, we will probably have some other job periodically pull the status, which add more complexity.
- use helix rest add 1 more internet hoop. 
With helix rest, the call will be: server->restCluster->Zk and return info; use server directly, will be: server->Zk and return info. Plus, helix rest internally use ServerContext.java (helix/rest), to retrieve TaskDriver.java, and then use TaskDriver to get info; Instead, we can directly use TaskDriver in our JobStateGenerator.
- Add TaskDriver into Generator.java enable us to skip workflow/job from ConfigGenerator.java 's "resources". Both job/db_resource are listed in idealState, which will all be retrieved by current ConfigGenerator.java. We will have to skip workflow/job to make the ConfigGenerator.java continue to work once we have Task state model enabled on these db clusters. 

Potential drawbacks if not use helix rest:
- code change into Generator.java to support job status report: add a TaskDriver. But, since taskDrive is constructed from ZkClient or HelixManager, we can easily construct it. 
- every ExternalView change will trigger status report which will read/write to Zk, which might increase traffic to Zk. This will at most double the *Generator traffic to Zk, by monitoring zk traffic, and evenly redistribute cluster among zk, we could support this new feature.

kangnanli:
Current design to report job status similar to ConfigGenerator might add too much overhead to the Participant. 
Will consider add job status report into Spectator or some other process, which can periodically pull the status directly from zk. 
Close this pull request.

PR #356, opened 2020-03-29
[cluster management] add backup taskFactory and task impl

kangnanli:
Add backup support. This will enable schedule ad-hoc or recurrent backup of segments. With backup, the cluster is more resilient to cluster failure; and provide source data for downstream offline data analysis. 

Details: after backup job is configured and submit to helix task framework (ie. through helix rest); helix will pick up the job and create backup tasks, and each Task is responsible to call backupDB from rocksdb_admin to upload db from local to cloud.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -281,6 +281,22 @@ public class Utils {
     }
   }
 
+  public static void backupDBWithLimit(String host, int adminPort, String dbName, String hdfsPath,
+                                       int limitMbs)
+      throws RuntimeException {
+    LOG.error("Backup " + dbName + " from " + host + " to " + hdfsPath);
+    try {
+      Admin.Client client = getAdminClient(host, adminPort);
+
+      BackupDBRequest req = new BackupDBRequest(dbName, hdfsPath);
+      req.setLimit_mbs(limitMbs);
+      client.backupDB(req);
+    } catch (TException e) {
+      LOG.error("Failed to backup DB: ", e.toString());
+      throw new RuntimeException(e);
+    }
+  }
+
   /**
    * Restore the local DB from HDFS
    * @param adminPort
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
@@ -1,0 +1,121 @@
+/// Copyright 2017 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author kangnanli (kangnanli@pinterest.com)
+//
+
+package com.pinterest.rocksplicator.task;
+
+import com.pinterest.rocksplicator.Utils;
+
+import org.apache.helix.task.Task;
+import org.apache.helix.task.TaskResult;
+import org.apache.helix.task.UserContentStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * backup a single local db shard to cloud
+ */
+public class BackupTask extends UserContentStore implements Task {
+
+  private static final Logger LOG = LoggerFactory.getLogger(BackupTask.class);
+
+  private final String taskCluster;
+  private final String partitionName;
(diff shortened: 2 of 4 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/task/TaskUtils.java
```diff
@@ -0,0 +1,38 @@
...
+package com.pinterest.rocksplicator.task;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TaskUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(TaskUtils.class);
+
+  /**
+   * lock segment by segment and job creation time
```
newpoo, line 28: can you change the comment to reflect what the function really does?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/TaskUtils.java
```diff
@@ -0,0 +1,38 @@
...
+/// limitations under the License.
+
+//
+// @author kangnanli (kangnanli@pinterest.com)
+//
+
+package com.pinterest.rocksplicator.task;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
newpoo, line 22: are they needed?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/TaskUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class TaskUtils {
+  private static final Logger LOG = LoggerFactory.getLogger(TaskUtils.class);
```
newpoo, line 25: is it needed?

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -281,6 +281,22 @@ public static void backupDB(String host, int adminPort, String dbName, String hd
...
   }
 
+  public static void backupDBWithLimit(String host, int adminPort, String dbName, String hdfsPath,
+                                       int limitMbs)
+      throws RuntimeException {
+    LOG.error("Backup " + dbName + " from " + host + " to " + hdfsPath);
+    try {
+      Admin.Client client = getAdminClient(host, adminPort);
+
+      BackupDBRequest req = new BackupDBRequest(dbName, hdfsPath);
```
newpoo, line 291: are we backing up to hdfs?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTaskFactory.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private final String host;
+  private final int adminPort;
+  private final String cluster;
+  private CuratorFramework zkClient;
+
+  private static final int DEFAULT_BACKUP_LIMIT_MBS = 64;
+
+  public BackupTaskFactory(String host, int adminPort, String zkConnectString, String cluster) {
+    this.host = host;
```
newpoo, line 61: Do we really need it?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
```diff
@@ -0,0 +1,160 @@
...
+   */
+  @Override
+  public TaskResult run() {
+
+    String dbName = Utils.getDbName(partitionName);
+    String segmentName = partitionName.substring(0, partitionName.lastIndexOf("_"));
+
+    // backup does not need inter-process lock, since jobVersion equal resourceVersion and
+    // jobVersion is already unified at jobConfig level. put into userStore for state map generation
+    try (Locker locker = new Locker(jobVersionMutext)) {
```
newpoo, line 75: The comment says we don't need to lock.

Can you explain if we need it or not?

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
kangnanli: As current impl, I will still keep it as hdfs. Once we have task framework ready, we can easily switch to S3 or, add both hdfs and s3 support.

cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
kangnanli: removed it. Since task is only responsible execution now, it does not need be responsible to unifying resource version among tasks.

kangnanli:
Thanks for your quick review, sorry, I have to refactor the code quite a few to be consistent with the new design. 
**changelists**
1. Task is only responsible to execute the backup;
2. TaskFactory is only responsible to create Task with provided configs. 
3. Neither Task or TaskFactory has other logic, e.g. communicate with zk to unifying resource version among tasks. These logic will be moved to scheduling service (ie. Jenkins job, etc)
4. I split changes into 3 commits, especially separate out the unit tests.

cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTaskFactory.java
newpoo: We may want to remove host from Task and TaskFactory as it's only used in a logging. Local host name is not that useful in a log.

PR #357, opened 2020-03-31
[cluster-management] filter out workflows/jobs from cluster's resources

kangnanli:
ConfigGenerator use helixAdmin to get resources from the cluster, which will include both db resources, and workflows/jobs; this will break the configGenerator. Thus, we need exclude the wf/jobs from the resources retrieved from cluster's idealStates, and only pass db_resources for config Generator.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
@@ -24,6 +24,7 @@ import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixManager;
 import org.apache.helix.NotificationContext;
 import org.apache.helix.model.ExternalView;
+import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.participant.CustomCodeCallbackHandler;
 import org.apache.helix.spectator.RoutingTableProvider;
@@ -39,6 +40,7 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -100,6 +102,7 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
     HelixAdmin admin = helixManager.getClusterManagmentTool();
 
     List<String> resources = admin.getResourcesInCluster(clusterName);
+    filterOutTaskResources(resources);
 
     Set<String> existingHosts = new HashSet<String>();
 
@@ -237,4 +240,28 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
     disabledHosts = latestDisabledInstances;
     return true;
   }
+
+  /**
+   * filter out resources with "Task" state model (ie. workflows and jobs);
+   * only keep db resources from ideal states
+   */
+  public void filterOutTaskResources(List<String> resources) {
+    HelixAdmin admin = helixManager.getClusterManagmentTool();
+    Iterator<String> iter = resources.iterator();
+    while (iter.hasNext()) {
+      String res = iter.next();
+      IdealState ideal = admin.getResourceIdealState(clusterName, res);
+      if (ideal != null) {
+        String stateMode = ideal.getStateModelDefRef();
+        if (stateMode != null && stateMode.equals("Task")) {
+          iter.remove();
+        }
+      } else {
+        LOG.error(
+            "Did not remove resource from shard map generation, due to can't get ideal state for "
+                + res);
+      }
+    }
+  }
+
 }
--- a/cluster_management/pom.xml
+++ b/cluster_management/pom.xml
@@ -103,6 +103,18 @@
       <artifactId>slf4j-log4j12</artifactId>
       <version>1.7.25</version>
     </dependency>
+    <dependency>
+      <groupId>org.testng</groupId>
+      <artifactId>testng</artifactId>
+      <version>6.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.helix</groupId>
+      <artifactId>helix-core</artifactId>
+      <version>0.8.4</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <repositories>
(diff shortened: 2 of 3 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -237,4 +242,37 @@ private boolean updateDisabledHosts() {
...
+    List<String> db_resources = new ArrayList<>();
+
+    Map<String, WorkflowConfig> workflowConfigMap = (driver == null ? null : driver.getWorkflows());
+
+    if (driver != null && workflowConfigMap != null && !workflowConfigMap.isEmpty()) {
+      Set<String> workflows = workflowConfigMap.keySet();
+      for (String resource : resources) {
+        boolean isWfOrJob = false;
+        for (String workflow : workflows) {
+          if (!workflow.isEmpty() && resource.startsWith(workflow)) {
```
newpoo, line 262: How do we guarantee that no db resource will start with any workflow name?

Is there any function/tag/something in helix zk that can help us tell if a resource is a regular db resource or workflow resource directly?

kangnanli: We have planned to enforce duplication check (specifically, prefix duplication check) when new db resource or workflow addition. This can prevent no db resource and workflow/jobs will have same name. 

I can't find any existing method to retrieve db only resources from zk. I will ask in helix community to confirm this.

newpoo: Will replacing startswith() with equals() work?

The implementation of this function can be refactored/simplified a little bit.

kangnanli: I got some feedback from helix team, will re-implement the filtering.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -237,4 +239,27 @@ private boolean updateDisabledHosts() {
     disabledHosts = latestDisabledInstances;
     return true;
   }
+
+  /**
+   * filter out resources with "Task" state model (ie. workflows and jobs);
+   * only return db resources from ideal states
+   *
+   * db_resource and wf/job have to be unique, bcz wf/job is treated as "resource" in helix.
+   * uniqueness will be enforced when workflow/job/db_resource is added
```
newpoo, line 248: remove this two lines, as they are not relevant any more.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -237,4 +239,27 @@ private boolean updateDisabledHosts() {
...
+   */
+  public List<String> filterOutNoneDbResources(List<String> resources) {
```
newpoo, line 250: can we rename it to filterOutTaskResources() to make it more accurate to what the function does?

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -237,4 +239,27 @@ private boolean updateDisabledHosts() {
...
+    List<String> db_resources = new ArrayList<>();
+    for (String res : resources) {
+      IdealState ideal = admin.getResourceIdealState(clusterName, res);
+      if (ideal != null) {
+          String stateMode = ideal.getStateModelDefRef();
+          if (stateMode != null && !stateMode.equals("Task")) {
+            db_resources.add(res);
+          }
+      }
+    }
```
newpoo, line 261: Can we do it the other way around. i.e., filter out resources that we are sure they are Tasks? This seems to be safer to existing clusters.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -237,4 +242,27 @@ private boolean updateDisabledHosts() {
...
+      String res = iter.next();
+      IdealState ideal = admin.getResourceIdealState(clusterName, res);
+      if (ideal != null) {
+        String stateMode = ideal.getStateModelDefRef();
+        if (stateMode != null && stateMode.equals("Task")) {
+          iter.remove();
+        }
+      } else {
+        LOG.error(
+            "Did not remove resource from sharp map generation, due to can't get ideal state for "
```
newpoo, line 262: shard map

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -92,14 +95,16 @@ public void onConfigChange(List<InstanceConfig> configs, NotificationContext cha
 
   @Override
   @PreFetch(enabled = false)
-  public void onExternalViewChange(List<ExternalView> externalViewList, NotificationContext changeContext) {
+  public void onExternalViewChange(List<ExternalView> externalViewList,
```
newpoo, line 98: undo this change?

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -39,12 +40,14 @@
...
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 public class ConfigGenerator extends RoutingTableProvider implements CustomCodeCallbackHandler {
+
```
newpoo, line 50: undo this change?

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -237,4 +242,27 @@ private boolean updateDisabledHosts() {
...
     return true;
   }
+
+  /**
+   * filter out resources with "Task" state model (ie. workflows and jobs);
+   * only keep db resources from ideal states
+   */
+  public void filterOutTaskResources(List<String> resources) {
+    HelixAdmin admin = helixManager.getClusterManagmentTool();
+    for (Iterator<String> iter = resources.iterator(); iter.hasNext(); ) {
```
newpoo, line 252: nit: move Iterator<String> iter = resources.iterator() outside the loop and use while() instead of for()?

PR #358, opened 2020-04-03
add dedup task and taskfactory

kangnanli:
dedup task will download db from given source path, and run full compaction locally based on rocksdb options from the downloaded db, and upload the compacted db back to cloud. 

This full compaction can dedup any lazy-merging keys in rocksdb sst files, and prepare for sst dump into other formats, e.g. SequenceFile,  for offline consumption.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
@@ -1,0 +1,113 @@
+package com.pinterest.rocksplicator.task;
+
+import com.pinterest.rocksplicator.Utils;
+
+import org.apache.helix.task.Task;
+import org.apache.helix.task.TaskResult;
+import org.apache.helix.task.UserContentStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DedupTask extends UserContentStore implements Task {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DedupTask.class);
+
+  private final String srcStorePathPrefix;
+  private final long resourceVersion;
+  private final String partitionName;
+  private final String taskCluster;
+  private final String job;
+  private final int adminPort;
+  private final String destStorePathPrefix;
+
+  public DedupTask(String srcStorePathPrefix, long resourceVersion, String partitionName,
+                   String taskCluster, String job, int adminPort,
+                   String destStorePathPrefix) {
+    this.srcStorePathPrefix = srcStorePathPrefix;
+    this.resourceVersion = resourceVersion;
+    this.partitionName = partitionName;
+    this.taskCluster = taskCluster;
+    this.job = job;
+    this.adminPort = adminPort;
+    this.destStorePathPrefix = destStorePathPrefix;
+  }
+
+  /**
+   * Execute the task.
+   * @return A {@link TaskResult} object indicating the status of the task and any additional
+   *         context information that can be interpreted by the specific {@link Task}
+   *         implementation.
+   */
+  @Override
+  public TaskResult run() {
+
+    if (srcStorePathPrefix.isEmpty() || resourceVersion == -1 || destStorePathPrefix.isEmpty()) {
+      String errMsg =
+          "Cancel the task, due to job command config map failed to provide all three of "
+              + "srcStorePathPrefix, resourceVersion, and destStorePathPrefix";
+      LOG.error(errMsg);
+      return new TaskResult(TaskResult.Status.CANCELED, errMsg);
+    }
+
+    // check task created with resource name and match prefix of target partition
+    String dbName = Utils.getDbName(partitionName);
+
+    try {
+
+      String srcStorePath =
+          String.format("%s/%s/%s", srcStorePathPrefix, String.valueOf(resourceVersion), dbName);
+
+      String destStorePath =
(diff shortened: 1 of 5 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
```diff
@@ -0,0 +1,113 @@
...
+   *         context information that can be interpreted by the specific {@link Task}
+   *         implementation.
+   */
+  @Override
+  public TaskResult run() {
+
+    if (srcStorePathPrefix.isEmpty() || resourceVersion == -1 || destStorePathPrefix.isEmpty()) {
+      String errMsg =
+          "Cancel the task, due to job command config map failed to provide all three of "
+              + "srcStorePathPrefix, resourceVersion, and destStorePathPrefix";
```
rajathprasad, line 47: What does `resourceVersion` correspond to?
Correct me if I missed something, but looks like resourceVersion is used only to construct the source and dest paths. Can we remove this logic from the task and let the client handle it? The client needs to pass in the path considering the version. This is a small detail but will make this task more generic and flexible.

cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
```diff
@@ -0,0 +1,113 @@
...
+    try {
+      Utils.addDB(dbName, adminPort);
+      Utils.closeDB(dbName, adminPort);
+      Utils.restoreLocalDB(adminPort, dbName, srcStorePath, "127.0.0.1", adminPort);
+      LOG.error("restoreDB is done, begin compactDB");
+
+      Utils.compactDB(adminPort, dbName);
+      LOG.error("compactDB is done");
+
+      Utils.backupDB("127.0.0.1", adminPort, dbName, destStorePath);
```
rajathprasad, line 91: Have you considered splitting these into separate tasks (jobs)? My concern is that if compaction or backup fails, on retries you will unnecessarily repeat all the prior steps.

cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTaskFactory.java
```diff
@@ -0,0 +1,89 @@
...
+    String destStorePathPrefix = "";
+
+    try {
+      Map<String, String> jobCmdMap = jobConfig.getJobCommandConfigMap();
+      if (jobCmdMap != null && !jobCmdMap.isEmpty()) {
+        if (jobCmdMap.containsKey("SRC_STORE_PATH_PREFIX")) {
+          srcStorePathPrefix = jobCmdMap.get("SRC_STORE_PATH_PREFIX");
+        }
+        if (jobCmdMap.containsKey("RESOURCE_VERSION")) {
+          resourceVersion = Long.parseLong(jobCmdMap.get("RESOURCE_VERSION"));
```
rajathprasad, line 61: Same here, can we avoid taking in the version as an argument?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
kangnanli: I would like to keep it for 2 reasons: 
1. keep consistent as backupTask. In backupTask, we made it more inclusive to parse version: client can pass a resourceVersion (i.e jobSubmission timestamp), or if not, we use the helixJobCreation time as the default version. Thus, the storage path in backup is: store_prefix/version/dbName. 
Therefore, to be consistent, we keep dedup has same format of : store_prefix/version/dbName. 
2.pass jobSubmission timestamp into Task will enable us to extract useful delay stats: taskExecutionTime - jobCreationTime/jobSubmissionTime. I think it might be useful to keep this optional open.

cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
kangnanli: good point. 
We can keep current implementation as a 1st version, since it won't repeat all the work even though it might repeat all function calls. For example, if 1st try failed at "upload", the task will repeat previous function calls, but, since local db still existed, download are incremental already with rocksdb api; so it will issue a restore call, but won't download anything. It might repeat a whole "compactDB", which could be an optimization we can do when we implement @cancel function in the Task.

kangnanli: This does raise the concern, if any of these steps failed without cleaning local db (ie. current @cancel is empty), and task got re-distributed to other host to retry. Then, hosts are left with some zoombie dbs.. 
I can optimize this when I implement @cancel in Task

ghanpipi:
lgtm

cluster_management/src/main/java/com/pinterest/rocksplicator/task/DedupTask.java
kangnanli: This is a good point. 
Current implementation, with default TaskStateMode, interruption of a `running` task to `task_aborted', 'stopped', or 'task_error' , etc. will call the @cancel to clear local dbs if existed. Then, it will repeat all add/restore/compact/upload steps. 
Current implementation can prevent un-cleaned local dbs with potential repeated works, so, I will open another ticket to optimize this step.

PR #363, opened 2020-04-28
Rocksplicator/use s3 as cloud store

kangnanli:
Previously, we use HDFS as cloud storage to store backed up dbs, and deduped files. Now, we add support use s3 as the storage when flag: s3Bucket is provided.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
@@ -40,9 +40,12 @@ public class BackupTask extends UserContentStore implements Task {
   private final long resourceVersion;
   private final String job;
   private final int adminPort;
+  private final boolean useS3Store;
+  private final String s3Bucket;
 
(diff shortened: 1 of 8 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTask.java
```diff
@@ -79,7 +84,7 @@ public TaskResult run() {
                   + "%s, job: %s, version: %d}", partitionName, storePath, taskCluster, job,
               resourceVersion));
 
-      executeBackup("127.0.0.1", adminPort, dbName, storePath, backupLimitMbs);
+      executeBackup("127.0.0.1", adminPort, dbName, storePath, backupLimitMbs, useS3Store);
```
rajathprasad, line 87: Did you miss passing in `s3Bucket`?

kangnanli: yes. fixed it

PR #364, opened 2020-04-30
Rocksplicator shouldn't hard code allowed az which is application dep…

gopalrajpurohit:
…endent

### DIFF
--- a/common/availability_zone.cpp
+++ b/common/availability_zone.cpp
@@ -27,17 +27,6 @@
 
 namespace {
 
-bool isAllowedAz(const std::string& az) {
-  static const std::set<std::string> allowed_azs = {
-    "us-east-1a",
-    "us-east-1c",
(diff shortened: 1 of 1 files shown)


gopalrajpurohit:
> Can you add the new AZs that you want to support to the allowed_azs as a whit list?
> 
> This is more defensive just in case when the format of the value returned by "http://169.254.169.254/latest/meta-data/placement/availability-zone" changes. This is out of our control. In that case, we'd crash a few processes to fail loudly.

As a public library, I feel this is not right place for such check. If applications wants to restrict deployment to specific availability zones, it can be better done by applications themselves....

Adding a whitelist in this library restricts it's usability. and unnecessary constraint. We can instead
provide for an application/user specified filter to be applied which is white list... This way, users can customize the whitelist of AZ's in which they want their application instances to be deployed.

PR #365, opened 2020-05-05
[common] Catch TApplicationException from recv_func

richardzhangrui:
I found sometimes in our application logs i saw std::exception and it turns out to be TApplicationException thrown by recv_func.

Change safe_future_client to catch TApplicationException to provide us more visibility

### DIFF
--- a/common/safe_future_client.h
+++ b/common/safe_future_client.h
@@ -186,7 +186,7 @@ class SafeFutureClient {
 
     auto ex = folly::try_and_catch<std::exception,
                                    apache::thrift::TApplicationException,
-                                   typename RequestTraits<T>::exception_type>([&p, &recvState]() {
+                                   typename RequestTraits<T>::exception_type>([p, recvState]() {
        typename RequestTraits<T>::response_type response;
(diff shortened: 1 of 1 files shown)


common/safe_future_client.h
```diff
@@ -183,16 +184,16 @@ class SafeFutureClient {
...
-      p->setValue(std::move(response));
-    } catch (typename RequestTraits<T>::exception_type& e) {
-      p->setException(
-        folly::make_exception_wrapper<
-          typename RequestTraits<T>::exception_type>(std::move(e)));
-    } catch (std::exception& e) {
-      p->setException(folly::exception_wrapper(std::move(e)));
+    auto ex = folly::try_and_catch<std::exception,
+                                   apache::thrift::TApplicationException,
+                                   typename RequestTraits<T>::exception_type>([&p, &recvState]() {
```
newpoo, line 189: nit: can we capture by value instead of capture by reference? which can reduce one level of indirect access.

richardzhangrui: sure. just found p is a pointer..

PR #367, opened 2020-05-09
[rocksplicator][statemodel]: Add entry-exit log messages for each of StateModel's transition methods

gopalrajpurohit:
…on methods

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -177,6 +177,11 @@ public class Utils {
         + " for " + message.getPartitionName());
   }
 
+  public static void logTransitionCompletionMessage(Message message) {
+    LOG.error("Switched from " + message.getFromState() + " to " + message.getToState()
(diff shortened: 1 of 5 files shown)


gopalrajpurohit:
> Might be a good idea to add it to all the StateModels?

Done.

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -177,6 +177,11 @@ public static void logTransitionMessage(Message message) {
         + " for " + message.getPartitionName());
   }
 
+  public static void logTransitionCompletionMessage(Message message) {
+    LOG.error("Switched from " + message.getFromState() + " to " + message.getToState()
```
missyjcat, line 181: I feel like it would be difficult to read the difference between "Switching" and "Switched." Maybe change this to `"Completed switch from "`?

gopalrajpurohit: Agreed:

Replaced with 

Transition Started
Transition Completed

PR #368, opened 2020-05-09
[rocksplicator][kafka config]: Add method to read in kafka java client properties file

gopalrajpurohit:
Simple configuration/properties file providing k=v in simple format can be read in for allowing application to set kafka parameters ...

# comment
# empty line is ok as below

key=value
key1=value1

# empty line with white spaces is not accepted.

### DIFF
--- a/common/kafka/kafka_config.cpp
+++ b/common/kafka/kafka_config.cpp
@@ -1,0 +1,54 @@
+//
+// Created by Gopal Rajpurohit on 5/8/20.
+//
+
+#include <fstream>
+
+#include "glog/logging.h"
+#include "common/kafka/kafka_config.h"
+
+namespace kafka {
+
+/**
+ * @brief Read (Java client) configuration file
+ */
+bool read_conf_file(
+  const std::string &conf_file,
+  const std::shared_ptr<ConfigMap> configMap,
+  bool compatibility_flag) {
+  std::ifstream inf(conf_file.c_str());
+
+  if (!inf) {
+    LOG(ERROR) << ": " << conf_file << ": could not open file" << std::endl;
+    return false;
+  }
+
+  LOG(INFO) << ": " << conf_file << ": read config file" << std::endl;
+
+  std::string line;
+  int linenr = 0;
+
+  while (std::getline(inf, line)) {
+    linenr++;
+
+    // Ignore comments and empty lines
+    if (line[0] == '#' || line.length() == 0)
+      continue;
+
+    // Match on key=value..
+    size_t d = line.find("=");
+    if (d == 0 || d == std::string::npos) {
+      LOG(INFO) << ": " << conf_file << ":" << linenr << ": " << line
+                << ": invalid line (expect key=value): " << ::std::endl;
+      return false;
+    }
+
+    std::string key = line.substr(0, d);
+    std::string val = line.substr(d + 1);
+
+    configMap->insert(make_pair(key, make_pair(val, compatibility_flag)));
+  }
+  inf.close();
+  return true;
+}
+} // namespace kafka
--- a/common/kafka/kafka_config.cpp
+++ b/common/kafka/kafka_config.cpp
@@ -13,6 +13,7 @@
 /// limitations under the License.
 
 #include <fstream>
+#include <folly/ScopeGuard.h>
 
 #include "glog/logging.h"
 #include "common/kafka/kafka_config.h"
@@ -37,6 +38,10 @@ bool KafkaConfig::read_conf_file(
   std::string line;
   int linenr = 0;
 
+  SCOPE_EXIT {
+    inf.close();
+  };
+
   while (std::getline(inf, line)) {
     linenr++;
 
@@ -57,7 +62,6 @@ bool KafkaConfig::read_conf_file(
 
     (*configMap)[key] = std::move(val);
   }
-  inf.close();
   return true;
 }
 } // namespace kafka
--- a/common/kafka/kafka_config.cpp
+++ b/common/kafka/kafka_config.cpp
@@ -14,8 +14,8 @@
 
 #include <fstream>
 #include <folly/ScopeGuard.h>
+#include <glog/logging.h>
 
-#include "glog/logging.h"
 #include "common/kafka/kafka_config.h"
 
 namespace kafka {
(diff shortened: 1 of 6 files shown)


common/kafka/tests/client_config.properties
```diff
@@ -0,0 +1,10 @@
+# Comment Java kafka client config file.
+# Following is empty line without any white space.
+
+enable.sparse.connections=true
+
+timeout_millis=1200
+socket_timeout_ms=300
+
```
rajathprasad, line 8: Can you add a commented line to the test?

common/kafka/kafka_config.h
```diff
@@ -0,0 +1,32 @@
...
+#include <memory>
+#include <string>
+
+namespace kafka {
+
+typedef std::map<std::string, std::pair<std::string, bool>> ConfigMap;
+
+class KafkaConfig {
+public:
+  static bool read_conf_file(
```
rajathprasad, line 27: Can you add some documentation for this function (what function does, and what each argument is)

common/kafka/kafka_config.h
```diff
@@ -0,0 +1,32 @@
...
+    const std::string &conf_file,
+    const std::shared_ptr<ConfigMap> configMap,
+    bool compatibility_flag);
```
rajathprasad, line 30: Since this is an input argument, can you move it to before the output argument?
http://google.github.io/styleguide/cppguide.html#Output_Parameters

common/kafka/kafka_config.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (d == 0 || d == std::string::npos) {
+      LOG(INFO) << ": " << conf_file << ":" << linenr << ": " << line
+                << ": invalid line (expect key=value): " << ::std::endl;
+      return false;
+    }
+
+    std::string key = line.substr(0, d);
+    std::string val = line.substr(d + 1);
+
+    (*configMap)[key] = make_pair(val, compatibility_flag);
```
rajathprasad, line 59: What is the purpose of the compatibility_flag?

common/kafka/kafka_config.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace kafka {
+
+typedef std::map<std::string, std::pair<std::string, bool>> ConfigMap;
+
+class KafkaConfig {
+public:
+  static bool read_conf_file(
+    const std::string &conf_file,
+    const std::shared_ptr<ConfigMap> configMap,
```
rajathprasad, line 29: Any reason this is a shared_ptr instead of passing by reference?

common/kafka/kafka_config.cpp
```diff
@@ -0,0 +1,64 @@
...
+  LOG(INFO) << ": " << conf_file << ": read config file" << std::endl;
+
+  std::string line;
+  int linenr = 0;
+
+  while (std::getline(inf, line)) {
+    linenr++;
+
+    // Ignore comments and empty lines
+    if (line[0] == '#' || line.length() == 0)
```
rajathprasad, line 45: Should the line.length check be first? line[0] will fail if line is empty.

common/kafka/kafka_config.cpp
gopalrajpurohit: I think this doesn't belong here... Will move to application code using this configurations.

common/kafka/kafka_config.cpp
gopalrajpurohit: 👍

common/kafka/kafka_config.h
gopalrajpurohit: 👍

common/kafka/kafka_config.cpp
gopalrajpurohit: 👍

common/kafka/kafka_config.h
gopalrajpurohit: 👍 Since I removed last parameter, as I feel it is better for application using this function to impose any restriction on compatibility at point of use.

common/kafka/kafka_config.h
gopalrajpurohit: 👍

common/kafka/tests/client_config.properties
gopalrajpurohit: 👍

common/kafka/kafka_config.h
```diff
@@ -0,0 +1,40 @@
...
+  // The format of properties file is similar to java properties file
+  // except it doesn't dereference any escape characters or special characters.
+  // Empty lines must not contain any whitespace.
+  // parameters:
+  // conf_file: path of configuration file to read.
+  // configMap: Output-only parameter, map containing all key-value pairs read
+  //            from the provided config file.
+  static bool read_conf_file(
+    const std::string &conf_file,
+    ConfigMap &configMap);
```
rajathprasad, line 38: Sorry, I realized I typed it incorrectly in my last comment. You need to pass it as a pointer (not reference), so that the calling sight is aware that this is going to be changed.

gopalrajpurohit: so shared_ptr is fine?

gopalrajpurohit: After offline conversation... and rajath pointing to https://google.github.io/styleguide/cppguide.html#Output_Parameters guidelines... changed this to raw pointer for output paramter.

common/kafka/kafka_config.cpp
```diff
@@ -0,0 +1,63 @@
...
+                << ": invalid line (expect key=value): " << ::std::endl;
+      return false;
+    }
+
+    std::string key = line.substr(0, d);
+    std::string val = line.substr(d + 1);
+
+    (*configMap)[key] = std::move(val);
+  }
+  inf.close();
```
rajathprasad, line 60: Can you close the file under a SCOPE_EXIT so that it closes whenever the function scope ends? In the current case, it doesn't close the file when returning on line 52

You can follow this example: https://github.com/pinterest/rocksplicator/blob/ec91db9af275648ea7905eea33409be6af61f1c3/common/availability_zone.cpp#L55

gopalrajpurohit: Done.

PR #369, opened 2020-05-09
[rocksplicator][kafkaconsumer]: Allow kafka client parameters to be provided in configuration file

gopalrajpurohit:
Depends on : https://github.com/pinterest/rocksplicator/pull/368 pull request

PR #370, opened 2020-05-09
[rocksplicator][librdkafka]:upgrade librdkafka to version 1.4.0

gopalrajpurohit:
Depends on https://github.com/pinterest/rocksplicator/pull/369

The major change in this PR is library upgrade for librdkafka from 0.11.4 to 1.4.0. There are several improvements from last version used in rocksplicator. However the one that motivated this upgrade is availability of sparse connections, as we have dealt with number of connections to brokers causing concern with big rocksplicator bootstrap clusters.

Note:  We found one functional regression w.r.t. it's usage in rocksplicator regarding enable.partition.eof configuration property which was fixed in https://github.com/pinterest/rocksplicator/pull/369, as rocksplicator's BootstrapStateModel depends on this feature.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
       cache:
         - ccache
       before_install:
-        - docker build -f $TRAVIS_BUILD_DIR/docker/Dockerfile -t local/rocksplicator-build:latest
+        - docker pull rajathprasad/rocksplicator-build:latest
       script:
         - cd $HOME/.ccache
         - |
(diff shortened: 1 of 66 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ matrix:
       cache:
         - ccache
       before_install:
-        - docker pull rajathprasad/rocksplicator-build:latest
+        - docker pull gopalrajpurohit/rocksplicator-build:librdkafka_1_4_0
```
rajathprasad, line 10: My personal preference is to tag is `latest` instead of `build:librdkafka_1_4_0`. That's the convention I've seen.

gopalrajpurohit: latest tag is moving target which in normal case is fine. Here since we are upgrading the librdkafka significantly... and is the major change, I feel tagging it likewise would be more appropriate....

PR #372, opened 2020-06-04
add restore task to Participant

kangnanli:
add Restore task into participant, which will enable participant cluster to restore from a static backup from cloud.
Usage: this will use to restore/rollback a segment to a previous version (which need be backup ahead of time); or, it could be used to bootstrap a copy of the segment into another cluster.  
Limitation: during restore, it will close the DB first, which means real time write to the DB during the Restore will be loss.


kangnanli:
Since we have decided to enable backup with checksum to resolve the filename conflicts during multiple rocksdbs instances backup to the same directory, instead of append "hostName" to storage path. The restore will work as before, we no further changes needed. 
Thus,  I will merge it now.

PR #373, opened 2020-06-08
[FileWatcherHub]: Enable multiple observers for notifiations on file …

gopalrajpurohit:
…watchers to same file

### DIFF
--- a/common/file_watcher_hub.cpp
+++ b/common/file_watcher_hub.cpp
@@ -1,0 +1,88 @@
+/// Copyright 2019 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#include "file_watcher.h"
+#include "file_watcher_hub.h"
+#include "glog/logging.h"
+
+namespace common {
+
+FileWatcherHub * FileWatcherHub::Instance() {
+  static FileWatcherHub instance;
+  return &instance;
+}
+
+bool FileWatcherHub::subscribe(
+  std::string fileName,
+  std::string observerId,
+  EventObserverPtr observerPtr) {
+  folly::RWSpinLock::WriteHolder write_gaurd(subscription_lock_);
+  auto watcherIter = subscriptions_.find(fileName);
+  if (watcherIter == subscriptions_.end()) {
+    // Add to the subscription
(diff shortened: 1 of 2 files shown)


common/file_watcher_hub.cpp
```diff
@@ -0,0 +1,88 @@
...
+  folly::RWSpinLock::WriteHolder write_gaurd(subscription_lock_);
+  auto watcherIter = subscriptions_.find(fileName);
+  if (watcherIter != subscriptions_.end()) {
+    auto obsPtr = watcherIter->second->find(observerId);
+    if (obsPtr != watcherIter->second->end()) {
+      watcherIter->second->erase(obsPtr);
+      if (watcherIter->second->empty()) {
+        // Remove watch on the given fileName
+        subscriptions_.erase(watcherIter);
+        common::FileWatcher::Instance()->RemoveFile(fileName);
```
newpoo, line 78: This can cause deadlock?

The thread calling unsubscribe() holding subscription_lock_  can block in removeFile() waiting the file watcher thread, while file watcher thread could execute line 46 above trying to grab subscription_lock_.

It might be easier to add this feature to FileWatcher itself.

gopalrajpurohit: Good point. Changing FileWatcher is bit more involved. Instead I could move the RemoveFile line down outside the subscription_lock_ guard. That seems easier approach?

common/file_watcher_hub.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    auto obsPtr = watcherIter->second->find(observerId);
+    if (obsPtr == watcherIter->second->end()) {
+      watcherIter->second->insert(ObserverEventCallbackPair(observerId, observerPtr));
+    }
+
+    CHECK(common::FileWatcher::Instance()->AddFile(
+      fileName,
+      [fileName = fileName, this] (std::string content) {
+        folly::RWSpinLock::ReadHolder read_gaurd(subscription_lock_);
```
gopalrajpurohit, line 49: This is a deadlock here...

FileWatcher AddFile will immediately callback ones a successful watch is established... which means... here we will have FileWatcher thread calling this callback trying to obtain subscription_lock_ ... which is already held by thread calling subscribe ...

common/file_watcher_hub.cpp
gopalrajpurohit: deadlock won't happen between thread trying to call AddFile and RemoveFile, as those methods are called exclusively.

However, the callback requires a lock.. so scenario, where unusubscribe is called... while executing RemoveFile, will have held the subscription_lock_.... Before RemoveFile is executed... a callback is called... which will block trying to acquire subscription_lock_ and will block. So now the callback cann't be executed... and at the same time RemoveFile now trying to remove File cann't be executed if FileWatcher tries to remove a File on which callback is being executed (in best case)... meaning a deadlock.

Similar scenario happens when AddFile is called in subscribe method... holding a write lock.... Internally AddFile will wait on a thread that is supposed to immediately callBack... and the callback gets blocked... and hence AddFile itself gets blocked causing deadlock again.

Will need to think more about better way to do this.

As per existing FileWatcher... it doesn't have mechanism for identifying which callback to unsubscribe. As different subscribers may unsubscribe at different times, we need a mechanism for identifying who wan't to unsubscribe watching a specific file.

Thanks for pointing out this issue, will re-think the approach.

PR #375, opened 2020-06-11
add timestamp to local bakcup dir

ghanpipi:
if two shards send the backup reqeust to the same at the same time window, it
will fail due to we use the db segment name as the local backup dir

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -731,7 +731,8 @@ void AdminHandler::async_tm_backupDBToS3(
     return;
   }
 
-  auto local_path = FLAGS_rocksdb_dir + "s3_tmp/" + request->db_name + "/";
+  auto ts = common::timeutil::GetCurrentTimestamp();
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -731,7 +731,8 @@ void AdminHandler::async_tm_backupDBToS3(
     return;
   }
 
-  auto local_path = FLAGS_rocksdb_dir + "s3_tmp/" + request->db_name + "/";
+  auto ts = common::timeutil::GetCurrentTimestamp();
+  auto local_path = folly::stringPrintf("%ss3_tmp/%s%d/", FLAGS_rocksdb_dir.c_str(), request->db_name, ts);
```
newpoo, line 735: add c_str() to db_name

rocksdb_admin/admin_handler.cpp
```diff
@@ -789,7 +790,8 @@ void AdminHandler::async_tm_restoreDBFromS3(
...
```
newpoo, line 794: same here.

PR #376, opened 2020-06-11
Pulling filepoller and multifilepoller sources from latest wangle library

gopalrajpurohit:
Adds FilePoller and MultiFilePoller sources from wangle library to rocksplicator, to use for file rotations providing SSL configurations.


gopalrajpurohit:
> Can you add a TODO to each file added in this PR to remove them once we upgrade to a later version of wangle?

Added one in docker/Dockerfile, as change will be required in docker/Dockerfile when we upgrade wangle.

PR #377, opened 2020-06-16
add error count of a Kafka_watcher

hjiangh:
Screen Shot 2020-06-15 at 9 23 45 PM
Screen Shot 2020-06-15 at 9 24 06 PM
 
michaelmi and rajathprasad accepted this revision on 6/12 in company phabricator.

### DIFF
--- a/common/kafka/kafka_watcher.cpp
+++ b/common/kafka/kafka_watcher.cpp
@@ -223,7 +223,7 @@ uint32_t KafkaWatcher::ConsumeUpToNow(const kafka::KafkaConsumer& consumer) {
       // happens when consumer hasn't got any message from the broker within the
       // timeout. We should retry in this case.
     } else {
-      ++err_count_;
+      err_count_.fetch_add(1, std::memory_order_seq_cst);
       common::Stats::get()->Incr(
           getFullStatsName(kKafkaWatcherNeedReEstablish,
               {kafka_watcher_metric_tag_}));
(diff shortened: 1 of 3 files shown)


common/kafka/kafka_watcher.cpp
```diff
@@ -223,6 +223,10 @@ uint32_t KafkaWatcher::ConsumeUpToNow(const kafka::KafkaConsumer& consumer) {
       // happens when consumer hasn't got any message from the broker within the
       // timeout. We should retry in this case.
     } else {
+      ++err_count_;
```
newpoo, line 226: can we use fetch_add() instead of ++ to make it more explicit that err_count_ is a std::atomic?

common/kafka/kafka_watcher.cpp
```diff
@@ -423,6 +427,10 @@ void KafkaWatcher::StartWatchLoop() {
             // receiving ERR__PARTITION_EOF and there was still no messages to
             // be consumed after timeout.
           } else {
+            ++err_count_;
```
newpoo, line 430: same here.

common/kafka/kafka_watcher.h
```diff
@@ -93,6 +93,8 @@ class KafkaWatcher {
   // stop in parallel
   void Stop() { is_stopped_.store(true); }
 
+  uint64_t ErrorCount() { return err_count_.load(); }
```
newpoo, line 96: do you plan to use it in a separate PR?

common/kafka/kafka_watcher.cpp
hjiangh: thanks for suggestion. done

common/kafka/kafka_watcher.cpp
hjiangh: done

common/kafka/kafka_watcher.h
hjiangh: It is up to the caller how to use it.

common/kafka/kafka_watcher.cpp
hjiangh: in https://github.com/pinterest/rocksplicator/pull/377/commits/8524c2732077825e6063701eb9411ec9d8522728

PR #378, opened 2020-06-17
Enable Re-creating kafka-consumer (rdkafka consumer) when watched files have changed

gopalrajpurohit:
In case of SSL context (cert,key and root cert chains files) are rotated, the librdkafka doesn't automatically re-load the SSL context from changed files. This change is to provide a simple mechanism of re-loading SSL context by re-creating rdkafka-consumer instance if the watched files have changed.


common/kafka/kafka_consumer.h
```diff
@@ -15,34 +15,32 @@
...
 namespace kafka {
 
 class KafkaConsumer {
 public:
-  KafkaConsumer(const std::unordered_set<uint32_t>& partition_ids,
-                const std::string& broker_list,
-                const std::unordered_set<std::string>& topic_names,
-                const std::string& group_id,
-                const std::string& kafka_consumer_type);
+  KafkaConsumer(const std::unordered_set<uint32_t> &partition_ids,
```
newpoo, line 29: In rocksplicator repo, we have been using following space style, i.e., const T& t;

Can you undo these changes.

gopalrajpurohit: sure. np.

common/kafka/kafka_consumer_holder.cpp
```diff
@@ -0,0 +1,448 @@
...
+    LOG(INFO) << "Assigned kafka consumer for topic: "
+              << folly::join(",", topic_names_)
+              << ", partitions: "
+              << folly::join(",", partition_ids_)
+              << std::endl << std::flush;
+
+    return true;
+  }
+
+  bool seekToRelevantOffsets(const TopicPartitionToValueMap<ConsumedOffset>* consumedOffsets) {
```
rajathprasad, line 282: Is `consumedOffsets` an in-parameter? Can we pass it as a const reference?

common/kafka/kafka_consumer_holder.cpp
```diff
@@ -0,0 +1,448 @@
...
+  }
+};
+
+RdKafkaConsumerHolder* RdKafkaConsumerHolderFactory::createInstance(
+  const std::unordered_set<uint32_t>& partition_ids,
+  const std::string& broker_list,
+  const std::unordered_set<std::string>& topic_names,
+  const std::string& group_id,
+  const std::string& kafka_consumer_type) {
+  return new RdKafkaConsumerHolderRenewable(partition_ids, broker_list, topic_names, group_id, kafka_consumer_type);
```
rajathprasad, line 441: So will we be using a `RdKafkaConsumerHolderRenewable` consumer for PII use cases and `RdKafkaConsumerHolderCached` for non-PII?

common/kafka/kafka_consumer.h
```diff
@@ -72,36 +70,33 @@ class KafkaConsumer {
...
-  std::unordered_set<std::string> topic_names_;
+  const std::unordered_set<std::string> topic_names_;
   std::string topic_names_string_;
   const std::unordered_set<uint32_t> partition_ids_;
-  std::shared_ptr<RdKafka::KafkaConsumer> consumer_;
+  const std::shared_ptr<RdKafkaConsumerHolder> consumer_;
   // Tag used when logging metrics, so we can differentiate between kafka
   // consumers for different use cases.
   const std::string kafka_consumer_type_metric_tag_;
+  std::shared_ptr<common::MultiFilePoller::CallbackId> cbIdPtr_;
```
rajathprasad, line 96: Can we use the snake_case naming style to keep it consistent? 
Also, how about renaming to `callback_id_`?

common/kafka/kafka_consumer.cpp
```diff
@@ -20,293 +20,248 @@
...
+        continue;
+      }
+
+      return;
+    }
+    LOG(INFO) << "Successfully created kafka consumer for partitions: "
+              << partition_ids_str_ << ", topics: "
+              << folly::join(",", topic_names_);
+    is_healthy_.store(true);
+  } while (std::atomic_load(&reset_));
```
rajathprasad, line 167: Is this while loop for the case where the callback is called while the partitions are being assigned?

common/kafka/kafka_consumer.cpp
```diff
@@ -20,293 +20,248 @@
...
-    return false;
-  }
+                         {kafka_consumer_type_metric_tag_}));
+
+      if (std::atomic_load(&reset_)) {
+        continue;
+      }
+      return false;
+    }
+  } while (std::atomic_load(&reset_));
```
rajathprasad, line 203: Do we need a while loop here? If the flag gets reset while seeking (line 191), isn't it okay to return false? The caller can retry.

common/kafka/kafka_consumer.cpp
```diff
@@ -367,46 +332,82 @@ bool KafkaConsumer::SeekInternal(
...
+RdKafka::Message* KafkaConsumer::Consume(int32_t timeout_ms) {
+  do {
+    if (consumer_ == nullptr || consumer_->getInstance() == nullptr) {
+      return nullptr;
+    }
 
-  auto* message = consumer_->consume(timeout_ms);
+    while (std::atomic_exchange(&reset_, false)) {
+      consumer_->resetInstance(&consumedOffsets_);
+    }
```
rajathprasad, line 364: This function is in the critical path for consuming kafka messages. Is it necessary to do a read-modify-write every time? Can we do just a read, instead? Or rely on the logic in lines 368->372 to reset the consumer?
I'm concerned about the performance impact of this (especially for non-PII cases where reset will never be set to True)
How do you feel about having another variable to indicate if this is a PII use case or not (or of the file watcher has been setup)? This variable will not change once initialized, and we could check that first before checking _reset.

common/kafka/kafka_consumer_holder.cpp
gopalrajpurohit: No. RdKafkaConsumerHolderCached is cache of already constructed RdKafkaConsumer instance... It doesn't know how to re-construct it.... So, it doesn't reset anything... This is used in test.

In production scenario, the RdKafkaConsumer is constructed from broker file providing broker instances.. RdKafkaConsumerHolderRenewable holds on to information required to re-construct RdKafkaConsumer. It is used both when no file is watched (where resetInstance is never called, and when a file is watched for changes and triggers resetInstance from KafkaConsumer class.)

common/kafka/kafka_consumer_holder.cpp
gopalrajpurohit: Sure.

common/kafka/kafka_consumer.h
gopalrajpurohit: Sure

common/kafka/kafka_consumer.cpp
gopalrajpurohit: while loop is to ensure the operations within single call of a method happens without any updates on file changes as notified...

common/kafka/kafka_consumer.cpp
gopalrajpurohit: call sites can be at different places... Isn't it better to retry here to ensure all callsites will get affected and it is transparent to call sites?

common/kafka/kafka_consumer.cpp
gopalrajpurohit: I modified this logic.. with a condition which check's if atomic operations are required at all.

PR #380, opened 2020-06-19
remove the slash prefix for hdfs_backup_dir to be consistent with s3 …

kangnanli:
Remove the the slash prefix from hdfs_backup_dir, so it could be consistent with s3_path conversion. 
During backup/dedup, client provide storePathPrefix which has no visibility of which cloud was used (hdfs or s3). Thus, rocksplicator code should be consistent about how to concat namenode/bucket with storePathPrefix. Therefore, in admin_handler, we need to concat hdfs_namenode + "/" + hdfs_backup_dir, which is consistent with s3, which will eventually use path: s3_bucket + "/" + s3_backup_dir.


kangnanli:
Instead of modifying rocksplicator code, we can also mark the cloud source when "store_prefix" is populated. I will seek that option for now and close the current pull request. 

Reopen when needed again.

PR #382, opened 2020-06-24
Manualcommit

JianGuoPinterest:
in librdkafka, it has manual commit feature and APIs, but we don't have those in the Kafka library inside rocksplicator. Added them so we can use the manual commit feature.


common/kafka/kafka_consumer.cpp
```diff
@@ -412,6 +412,26 @@ RdKafka::Message* KafkaConsumer::Consume(int32_t timeout_ms) {
...
 
+RdKafka::ErrorCode KafkaConsumer::Commit(RdKafka::Message* message, bool is_async) {
+  if (is_async) {
+    return rd_kafka_consumer_provider_->getInstance()->commitAsync(message);
+  } else {
+    return rd_kafka_consumer_provider_->getInstance()->commitSync(message);
+  }
+}
+
+RdKafka::ErrorCode KafkaConsumer::Commit(std::vector<RdKafka::TopicPartition*>& offsets, bool is_async) {
```
gopalrajpurohit, line 423: Shouldn't this be const reference? similarly in other changes.

JianGuoPinterest: I checked the interface librdkafka provided,  for commitSync, one accepts non-const reference when the other one accepts const reference, for committed, it only accepts non-const reference. I think in this case, we should use non-const reference, correct me if I'm wrong.
```
virtual ErrorCode RdKafka::KafkaConsumer::commitSync(std::vector<TopicPartition*>& offsets)	
virtual ErrorCode RdKafka::KafkaConsumer::commitAsync(const std::vector<TopicPartition*>& offsets)
virtual ErrorCode RdKafka::KafkaConsumer::committed(std::vector<TopicPartition*>& partitions, int timeout_ms)	
```

PR #383, opened 2020-07-07
enable share_files_with_checksum during rocksdb instance backup

kangnanli:
Rocksdb backupEngine support backup with adding "timestamp" and "checksum" to sst filename by setting flag: ```share_files_with_checksum``` in BackupOptions, so that multiple rocksdb instance could backup to the same directory without introducing conflicts. 
- We add this as an optional field in the BackupRequest so that whoever request the backup, could control whether enable the flag. 
- We did not introduce another global Gflag to control this due to our existing statemodels use the default BackupOptions which has this flag disabled, by passing from BackupRequest, we could keep existing statemodels still use old options with flag disabled; while our new backup transitions with this flag enabled. 
- this is only enabled when hdfs is used as cloud storage, since it will conflict with current implementation of checkpoint backup with S3Env.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTaskFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTaskFactory.java
@@ -77,6 +77,7 @@ public class BackupTaskFactory implements TaskFactory {
     String storePathPrefix = "";
     long resourceVersion = jobCreationTime;
     int backupLimitMbs = DEFAULT_BACKUP_LIMIT_MBS;
+    boolean shareFilesWithChecksum = true;
 
     try {
       Map<String, String> jobCmdMap = jobConfig.getJobCommandConfigMap();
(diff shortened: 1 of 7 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -914,12 +920,12 @@ void AdminHandler::async_tm_backupDBToS3(
     auto local_s3_util = createLocalS3Util(request->limit_mbs, request->s3_bucket);
     std::string formatted_s3_dir_path = rtrim(request->s3_backup_dir, '/');
     rocksdb::Env* s3_env = new rocksdb::S3Env(formatted_s3_dir_path, local_path, std::move(local_s3_util));
-
     if (!backupDBHelper(request->db_name,
                         formatted_s3_dir_path,
                         std::unique_ptr<rocksdb::Env>(s3_env),
                         request->__isset.limit_mbs,
                         request->limit_mbs,
+                        false,
```
rajathprasad, line 928: Can you add a comment explaining why it's set to false for s3?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/BackupTaskFactory.java
```diff
@@ -77,6 +77,7 @@ public Task createNewTask(TaskCallbackContext context) {
     String storePathPrefix = "";
     long resourceVersion = jobCreationTime;
     int backupLimitMbs = DEFAULT_BACKUP_LIMIT_MBS;
+    boolean shareFilesWithChecksum = true;
```
rajathprasad, line 80: By default (if not present in job command), you want to keep this as true?

kangnanli: Yes. by default, always enable checksum for backup/export. 
Compared to statemodel transitions, we want to be to backward consistent, so, by default, disable the checksum during backup.

PR #384, opened 2020-07-15
[common/future_util] optimize memory use time

newpoo:
This can reduce memory usage when speculative failover timeout is set to be too large.


newpoo:
Ran existing unittests.

PR #385, opened 2020-07-15
MultiFilePollerTest seems flaky for some, increasing timeout

gopalrajpurohit:


gopalrajpurohit:
Increasing timeout for flaky test.

PR #388, opened 2020-08-19
Add stat for when the follower requests to sync with the leader but it is ahead of the leader already

iprentic:


rocksdb_replicator/replicated_db.cpp
```diff
@@ -236,6 +239,12 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
   auto db = shared_from_this();
   std::weak_ptr<ReplicatedDB> weak_db = db;
   auto seq_no = static_cast<rocksdb::SequenceNumber>(request->seq_no);
+
+  // Inverse of predicate below: if requested sequence number is HIGHER than latest sequence number on leader, emit a stat)
+  if (FLAGS_emit_stat_for_leader_behind && db->db_->GetLatestSequenceNumber() < seq_no) {
+    incCounter(kReplicatorLeaderBehindCount, 1, db->db_name_);
```
newpoo, line 245: nit: Do you think it would be helpful if we use logMetric() and log (seq_no - local_latest_sequence_no)?

iprentic: Done!

rocksdb_replicator/replicator_stats.cpp
```diff
@@ -39,6 +39,8 @@ const std::string kReplicatorGetUpdatesSinceErrors =
 const std::string kReplicatorGetUpdatesSinceMs =
   "replicator_get_update_since_ms";
 const std::string kReplicatorWriteMs = "replicator_write_ms";
+const std::string kReplicatorLeaderBehindCount = "replicator_leader_behind_count";
```
newpoo, line 42: this can be removed now.

rocksdb_replicator/replicator_stats.h
```diff
@@ -31,6 +31,8 @@ extern const std::string kReplicatorRemoteApplicationExceptions;
 extern const std::string kReplicatorGetUpdatesSinceErrors;
 extern const std::string kReplicatorGetUpdatesSinceMs;
 extern const std::string kReplicatorWriteMs;
+extern const std::string kReplicatorLeaderBehindCount;
```
newpoo, line 34: this can be removed now.

PR #391, opened 2020-08-27
add monitor stats to rocksplicator

kangnanli:
Add MBeans to track rocksplicator stats

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
@@ -106,8 +106,7 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
 
   @Override
   @PreFetch(enabled = false)
-  public void onExternalViewChange(List<ExternalView> externalViewList,
-                                   NotificationContext changeContext) {
+  public void onExternalViewChange(List<ExternalView> externalViewList, NotificationContext changeContext) {
     generateShardConfig();
   }
 
@@ -134,7 +133,6 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
       if (externalView == null) {
         monitor.incrementConfigGeneratorNullExternalView();
         LOG.error("Failed to get externalView for resource: " + resource);
-        continue;
       }
       Set<String> partitions = externalView.getPartitionSet();
 
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
@@ -18,6 +18,8 @@
 
 package com.pinterest.rocksplicator;
 
+import com.pinterest.rocksplicator.monitoring.mbeans.RocksplicatorMonitor;
+
 import org.apache.helix.api.listeners.PreFetch;
 import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixConstants;
@@ -36,6 +38,7 @@ import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import com.google.common.base.Stopwatch;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -47,6 +50,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 public class ConfigGenerator extends RoutingTableProvider implements CustomCodeCallbackHandler {
+
   private static final Logger LOG = LoggerFactory.getLogger(ConfigGenerator.class);
 
   private final String clusterName;
@@ -56,8 +60,14 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
   private JSONObject dataParameters;
   private String lastPostedContent;
   private Set<String> disabledHosts;
+  private RocksplicatorMonitor monitor;
 
   public ConfigGenerator(String clusterName, HelixManager helixManager, String configPostUrl) {
+    this(clusterName, helixManager, configPostUrl, new RocksplicatorMonitor(helixManager.getInstanceName()));
+  }
+
+  public ConfigGenerator(String clusterName, HelixManager helixManager, String configPostUrl,
+                         RocksplicatorMonitor monitor) {
     this.clusterName = clusterName;
     this.helixManager = helixManager;
     this.hostToHostWithDomain = new HashMap<String, String>();
@@ -69,6 +79,7 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
     this.dataParameters.put("content", "{}");
     this.lastPostedContent = null;
     this.disabledHosts = new HashSet<>();
+    this.monitor = monitor;
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


kangnanli:
Add some test stats of get the MBeans attr after deploying Rocksplicator to clusters. 
note: replace the real hostname with <hostName> at below: 
```
<host>:~$ java -jar jmxterm-1.0.1-uber.jar --url localhost:10102
Welcome to JMX terminal. Type "help" for available commands.

$>domains
#following domains are available
...
RocksplicatorReport
...
$>domain RocksplicatorReport
#domain is set to RocksplicatorReport

$>beans
#domain = RocksplicatorReport:
RocksplicatorReport:InstanceName=<hostName>

$>bean RocksplicatorReport:InstanceName=<hostName>
#bean is set to RocksplicatorReport:InstanceName=<hostName>

$>info
#mbean = RocksplicatorReport:InstanceName=<hostName>
#class name = com.pinterest.rocksplicator.monitoring.mbeans.RocksplicatorSpectatorMonitor
# attributes
  %0   - ConfigGeneratorCalledCount (long, r)
  %1   - ConfigGeneratorNullExternalView (long, r)
  %2   - MaxConfigGeneratorLatency (double, r)
  %3   - MeanConfigGeneratorLatency (double, r)
  %4   - MinConfigGeneratorLatency (double, r)
#there's no operations
#there's no notifications

$>get ConfigGeneratorCalledCount
#mbean = RocksplicatorReport:InstanceName=<hostName>:
ConfigGeneratorCalledCount = 58;

$>get ConfigGeneratorNullExternalView
#mbean = RocksplicatorReport:InstanceName=<hostName>:
ConfigGeneratorNullExternalView = 0;

$>get MeanConfigGeneratorLatency
#mbean = RocksplicatorReport:InstanceName=<hostName>:
MeanConfigGeneratorLatency = 2179649.3157894737;
```

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -204,6 +221,10 @@ private void generateShardConfig() {
     } catch (Exception e) {
       LOG.error("Failed to post the new config", e);
     }
+
+    final long endNanos = System.nanoTime();
+    final int elapsedMicros = (int) ((endNanos - startNanos) / 1000);
+    monitor.reportConfigGeneratorLatency(elapsedMicros);
   }
```
gopalrajpurohit, line 228: Can you encapsulate this in separate StopWatch class...

StopWatch stopWatch = new StopWatch(true) <- start the timer


monitor.reportConfigGeneratorLatency(stopWatch.getElapsedMicros())

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -115,6 +129,9 @@ private void generateShardConfig() {
       }
 
       ExternalView externalView = admin.getResourceExternalView(clusterName, resource);
+      if (externalView == null) {
```
gopalrajpurohit, line 132: if externalView is null.... there is no need to continue for that resource.

This is a possibility that externalView is null in some cases, where there is a race condition between getting all resources and iterating to get externalView for a resource when one resource get's deleted in-between

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -94,11 +104,15 @@ public void onConfigChange(List<InstanceConfig> configs, NotificationContext cha
 
   @Override
   @PreFetch(enabled = false)
-  public void onExternalViewChange(List<ExternalView> externalViewList, NotificationContext changeContext) {
+  public void onExternalViewChange(List<ExternalView> externalViewList,
+                                   NotificationContext changeContext) {
     generateShardConfig();
   }
 
   private void generateShardConfig() {
```
gopalrajpurohit, line 112: This method should be synchronized.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
kangnanli: Nice. This is very likely. I will add a continue to skip when externalView is null then and report the stats as well.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
kangnanli: Are you worrying multiple threads will be working on `generateShardConfig` at the same time?

cluster_management/src/main/java/com/pinterest/rocksplicator/monitoring/mbeans/RocksplicatorSpectatorMonitor.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public void incrementConfigGeneratorCalledCount(int count) {
+    configGeneratorCalledCount += count;
+  }
+
+  void addLatency(LATENCY_TYPE type, double latency) {
+    assert (latencyMonitorMap.containsKey(type));
+    latencyMonitorMap.get(type).addData(latency);
+  }
```
gopalrajpurohit, line 38: Why assert here? if the latency type has not previously been registered... isn't it better to create it ? rather then assert?

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: Also it will be better to log the resource name for which externalView is found to be null, in case investigation is required to be done later.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: yes.

cluster_management/src/main/java/com/pinterest/rocksplicator/monitoring/mbeans/RocksplicatorSpectatorMonitor.java
kangnanli: We initiate/reset all the stats targeted at the Constructor of RocksplicatorSpectatorMonitor, so, later addLatency need to make sure only add stats to those stats targeted.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
kangnanli: Great. Thank, StopWatch is better.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
kangnanli: I will leave this method as it is and limit current pull request to be stats only. 
If future stats shows that we need synchronize this method, we can add synchronization then.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
kangnanli: @rajathprasad @gopalrajpurohit  is this too risky to continue generate ShardMap when externalView for certain resource are empty?
Pros: it could skip those situation mentioned by Gopal above, ie. resource got deleted between "getResources" and "resource iteration"
Cons: if some transient getExternalView error returned Null, we would still generate a new shardMap and excluding this resource and update the new shardMap to routers. Then, all requests to the resource will fail.

gopalrajpurohit (approved):
After offline conversation, we agree to limit this change for reporting stats. Next follow up change should be made to handle the null externalView correctly.

PR #392, opened 2020-09-04
[rocksplicator] delete db in async way

ghanpipi:
If the db/segment's size is huge, there will be latency spikes caused by high
disk IO usage when delete a helix resource which will close and delete all the
dbs in the host at the same time. In order to mitigate these spikes, we provide
an async way to handle the dbs deletion: 1. mv the dbs to db_tmp/ in
async_tm_clearDB which has cheap disk operation overhead. 2. delete the dbs
under db_tmp/ periodically.

Tested it in test & stage cluster, the spikes were gone.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -55,6 +55,7 @@
 #include "thrift/lib/cpp2/protocol/Serializer.h"
 #if __GNUC__ >= 8
 #include "folly/executors/CPUThreadPoolExecutor.h"
+#include "folly/system/ThreadName.h"
 #else
 #include "wangle/concurrent/CPUThreadPoolExecutor.h"
 #endif
@@ -120,6 +121,15 @@ DEFINE_int32(num_s3_upload_download_threads, 8,
 
 DEFINE_int64(max_s3_upload_download_task_queue_size, 1000, "The queue size in the executor");
 
+DEFINE_bool(enable_async_delete_dbs, false, "Enable delete db files in async way");
+
+DEFINE_int32(async_delete_dbs_frequency_sec,
+             20,
+             "How frequently in sec to check the dbs need deleting in async way");
+DEFINE_int32(async_delete_dbs_wait_sec,
+             60,
+             "How long in sec to wait between the dbs deletion");
+
 #if __GNUC__ >= 8
 using folly::CPUThreadPoolExecutor;
 using folly::LifoSemMPMCQueue;
@@ -160,6 +170,7 @@ const std::string kHDFSBackupMs = "hdfs_backup_ms";
 const std::string kHDFSRestoreMs = "hdfs_restore_ms";
 const std::string kS3BackupMs = "s3_backup_ms";
 const std::string kS3RestoreMs = "s3_restore_ms";
+const std::string kDeleteDBFailure = "delete_db_failure";
 
 int64_t GetMessageTimestampSecs(const RdKafka::Message& message) {
   const auto ts = message.timestamp();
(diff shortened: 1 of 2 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -389,6 +399,50 @@ CPUThreadPoolExecutor* S3UploadAndDownloadExecutor() {
...
+      common::Stats::get()->Incr(kDeleteDBFailure);
+      LOG(ERROR) << "Cant delete db: " << itr->path() << remove_err.message();
+    } else {
+      LOG(INFO) << "Done deleting db: " << itr->path();
+    }
+    std::this_thread::sleep_for(std::chrono::seconds(kDeleteDBWaitSec));
+  }
+}
+
+// schedule the periodically health check
```
rajathprasad, line 428: Fix the comment.

rocksdb_admin/admin_handler.cpp
```diff
@@ -389,6 +399,50 @@ CPUThreadPoolExecutor* S3UploadAndDownloadExecutor() {
...
+void scheduleDeleteDBs() {
+  auto delayed_future =
+      common::GenerateDelayedFuture(std::chrono::minutes(FLAGS_async_delete_dbs_frequency_minutes));
+
+#if __GNUC__ >= 8
+  std::move(delayed_future).then([](auto&&) {
```
rajathprasad, line 434: I think this is backward compatible and works with U14 too. Can you check?

rajathprasad (approved):
LGTM

rocksdb_admin/admin_handler.cpp
```diff
@@ -415,8 +469,25 @@ AdminHandler::AdminHandler(
...
     << "Invalid FLAGS_max_s3_sst_loading_concurrency: "
     << FLAGS_max_s3_sst_loading_concurrency;
+
+  if (FLAGS_enable_async_delete_dbs) {
+    static const std::string db_tmp_path = FLAGS_rocksdb_dir + "db_tmp/";
+    if (!boost::filesystem::exists(db_tmp_path)) {
+      boost::system::error_code create_err;
+      boost::filesystem::create_directories(db_tmp_path, create_err);
+      if (create_err) {
+        LOG(ERROR) << "Failed to create dir: " << db_tmp_path << create_err.message();
```
newpoo, line 479: Should we fail loudly in this case?

rocksdb_admin/admin_handler.cpp
```diff
@@ -389,6 +399,50 @@ CPUThreadPoolExecutor* S3UploadAndDownloadExecutor() {
...
+    }
+    boost::system::error_code remove_err;
+    boost::filesystem::remove_all(itr->path(), remove_err);
+    if (remove_err) {
+      common::Stats::get()->Incr(kDeleteDBFailure);
+      LOG(ERROR) << "Cant delete db: " << itr->path() << remove_err.message();
+    } else {
+      LOG(INFO) << "Done deleting db: " << itr->path();
+    }
+    std::this_thread::sleep_for(std::chrono::seconds(kDeleteDBWaitSec));
```
newpoo, line 424: so we are deleting a DB directory at a time. Should we consider using a finer control of throttling? For example, delete at a speed of X MB/s? Is there any existing library that we could use?

rocksdb_admin/admin_handler.cpp
```diff
@@ -389,6 +399,50 @@ CPUThreadPoolExecutor* S3UploadAndDownloadExecutor() {
...
+  }
+}
+
+// schedule the periodically health check
+void scheduleDeleteDBs() {
```
newpoo, line 429: This seems to be a little over complicated to me.
Can we just add a unique_ptr<thread> to AdminHandler and initialize it in AdminHandler:: AdminHandler if the gflag is set?

Using timer thread (delayed future) to delete data and sleep could cause trouble.

rocksdb_admin/admin_handler.cpp
ghanpipi: there are some better solutions: like use truncate() to gradually remove the db or use ionice to perform the deletion when the disk io is idle and no other processes compete for the disk io. but I dont find a good lib we could easily employ. so i prefer to adopting the current solution as a temp fix to unblock our use cases and then keep working on figuring out a a better solution for it

rocksdb_admin/admin_handler.cpp
```diff
@@ -389,6 +399,32 @@ CPUThreadPoolExecutor* S3UploadAndDownloadExecutor() {
   return &executor;
 }
 
+// The dbs moved to db_tmp/ shouldnt be re-used or re-opened, so we can
+// delete them via boost filesystem operations rather than rocksdb::DestroyDB()
+void deleteDBs() {
```
newpoo, line 404: deleteTmpDBs

rocksdb_admin/admin_handler.cpp
```diff
@@ -415,8 +451,48 @@ AdminHandler::AdminHandler(
...
+        LOG(ERROR) << "Failed to set thread name for DB deletion thread";
+      }
+
+      LOG(INFO) << "Starting DB deletion thread ...";
+      evb->loopForever();
+      LOG(INFO) << "Stopping DB deletioin thread ...";
+    });
+
+    scheduleDeleteDBs();
+  }
```
newpoo, line 478: why do we need an event base? can we just let the thread do the deletion work synchronously?

ghanpipi: the point is when we just use a thread with x min sleep, then for deployment, we may have to wait for x mins to stop and join the thread. use evb->terminateLoopSoon() could help us to avoid it

rocksdb_admin/admin_handler.cpp
```diff
@@ -160,6 +168,8 @@ const std::string kHDFSBackupMs = "hdfs_backup_ms";
 const std::string kHDFSRestoreMs = "hdfs_restore_ms";
 const std::string kS3BackupMs = "s3_backup_ms";
 const std::string kS3RestoreMs = "s3_restore_ms";
+const std::string kDeleteDBFailure = "delete_db_failure";
+const int kDeleteDBWaitSec = 80;
```
newpoo, line 172: it's better to make it a gflag so that we can tune without building a new package.

newpoo (approved):

PR #402, opened 2020-11-18
Make sure NoNode exceptions are retriable with BoundedExponentialBackoff Retry Policy for fetching metadata

gopalrajpurohit:
…offRetry

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/OnlineOfflineStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/OnlineOfflineStateModelFactory.java
@@ -26,7 +26,7 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
-import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.apache.curator.retry.BoundedExponentialBackoffRetry;
 import org.apache.helix.NotificationContext;
 import org.apache.helix.model.Message;
 import org.apache.helix.participant.statemachine.StateModel;
@@ -38,6 +38,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class OnlineOfflineStateModelFactory extends StateModelFactory<StateModel> {
+  private static final int MAX_ZK_RETRIES = 7;
   private final int adminPort;
   private final String cluster;
   private CuratorFramework zkClient;
@@ -48,7 +49,7 @@ public class OnlineOfflineStateModelFactory extends StateModelFactory<StateModel
     this.adminPort = adminPort;
     this.cluster = cluster;
     this.zkClient = CuratorFrameworkFactory.newClient(zkConnectString,
-        new ExponentialBackoffRetry(1000, 3));
+        new BoundedExponentialBackoffRetry(1000,  10000, MAX_ZK_RETRIES));
     zkClient.start();
   }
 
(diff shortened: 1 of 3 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -472,10 +477,55 @@ public static void checkStateTransitions(String fromState, String toState, Messa
...
+        try {
+          zkClient.sync().forPath(clusterMetaPath);
+          zkClient.sync().forPath(resourceMetaPath);
+          zkClient.sync().forPath(metadataPath);
+          return new String(zkClient.getData().forPath(metadataPath));
+        } catch (KeeperException.NoNodeException exp) {
+          LOG.error(String.format(
+              "MetaDataPath (%s) not visible or doesn't exist yet: ", metadataPath), exp);
+          // By throwing the OperationTimeoutException, we force the retry, as NoNode exception is
+          // not retryable.
```
kangnanli, line 515: Is there any wiki pointing to what types of exe are retryable, and what types are not?

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -472,10 +477,55 @@ public static void checkStateTransitions(String fromState, String toState, Messa
...
+    final String metadataPath = Utils.getMetaLocation(cluster, resourceName);
+
+    final AtomicInteger retryCount = new AtomicInteger(0);
+    return RetryLoop.callWithRetry(zkClient.getZookeeperClient(), new Callable<String>() {
+      @Override
+      public String call() throws Exception {
+        retryCount.incrementAndGet();
+        try {
+          zkClient.sync().forPath(clusterMetaPath);
+          zkClient.sync().forPath(resourceMetaPath);
```
kangnanli, line 508: I have two Qs: 1) can u share what does sync() do? why do we need it?
2) If we only access the overall path, why do we need to sync the parent path?

cluster_management/src/main/java/com/pinterest/rocksplicator/OnlineOfflineStateModelFactory.java
```diff
@@ -48,7 +49,7 @@
     this.adminPort = adminPort;
     this.cluster = cluster;
     this.zkClient = CuratorFrameworkFactory.newClient(zkConnectString,
-        new ExponentialBackoffRetry(1000, 3));
+        new BoundedExponentialBackoffRetry(1000,  10000, MAX_ZK_RETRIES));
```
kangnanli, line 52: ExponentialBackoffRetry is already bounded to retry up to 3 times. Any reason why we need to bound by time as well with BoundedExponentialBackoffRetry?

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
gopalrajpurohit: Check RetryLoop.shouldRetry and ReteryLoop.isRetryException to understand what Exceptions are automatically retried by RetryLoop in the CuratorFramework.

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
gopalrajpurohit: The CuratorFramework doesn't provide a documentation on it, but it believe it performs a synchronous sync from leader to the follower this client is connected to, hence it forces the latest view on given zknode.

Reason for recursively syncing path from top to bottom, is if the zk follower is lagging behind in syncing the parent path, it will not see a newly created children node. This is what is causing the regular exceptions when resources are created a fresh per version (internal concept)

cluster_management/src/main/java/com/pinterest/rocksplicator/OnlineOfflineStateModelFactory.java
gopalrajpurohit: In order to guarantee the max time. The intermediate intervals are randomized... so we can't guarantee on the upperbound on time. This will help keep that latency low, in worst case.

kangnanli (approved):
lgtm. 
thanks for the explanation

PR #404, opened 2020-12-01
[admin_handler] Do not raise exceptions if there are duplicate messag…

rajathprasad:
…es from helix.

We've seen instances where shards go into ERROR state when there are
duplicate state transition messages from helix. These ERROR states cause
degradation in the service. This diff ignores these duplicate messages
and returns an empty response immediately instead of raising an
exception.


gopalrajpurohit (commented):
These changes are specific to BootStrapModel. 

Instead of changes in here, wouldn't it be easier to prevent this scenario in all of the Models?

State Transitions are initiated in helix (java sub-component) and passed on to rocksdb_admin through RPC.

gopalrajpurohit (approved):

ghanpipi (approved):

PR #405, opened 2020-12-01
Upgrade helix version from 0.8.4 to 0.9.1

gopalrajpurohit:


kangnanli:
Have we already tested 0.9.1 on helix participant clusters to be compatible with helixUI/Rest/controller?

gopalrajpurohit:
> Have we already tested 0.9.1 on helix participant clusters to be compatible with helixUI/Rest/controller?

I tested 0.9.1 participant and it works with both 0.8.4 and 0.9.1 controller. So this should be good to go.

kangnanli (approved):
This rollout will impact all production clusters as helix participants. Can we double check the test before merge it?

PR #409, opened 2020-12-05
only enable dumping shard config to local if the dir exists

ghanpipi:


cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -83,6 +85,7 @@ public ConfigGenerator(String clusterName, HelixManager helixManager, String con
     this.lastPostedContent = null;
     this.disabledHosts = new HashSet<>();
     this.monitor = monitor;
+    this.enableDumpToLocal = new File("/var/log/helixspectator").exists();
```
kangnanli, line 88: why only enable write to local when this dir exist?

ghanpipi: can use it to only enable local dump on demand and also to avoid any potential issues caused by the dir is not created or no access right

PR #410, opened 2020-12-07
Add the rest of the state models for completeness.

rajathprasad:
Adding the following default helix state models to rocksplicator repo
for completeness:
1. LeaderStandby
2. OnlineOffline
3. MasterSlave

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/state_models/LeaderStandby
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/state_models/LeaderStandby
@@ -1,0 +1,39 @@
+{
+  "id" : "LeaderStandby",
+  "mapFields" : {
+    "DROPPED.meta" : {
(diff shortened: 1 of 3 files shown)


newpoo:
They are helix built-in state models. What's the motivation to add them to our repo?

rajathprasad:
> They are helix built-in state models. What's the motivation to add them to our repo?
For reference and completeness.

newpoo:
How do we keep them consistent with the ones in helix repo? Does it make more sense to use some kind of pointers for reference purpose?

rajathprasad:
> How do we keep them consistent with the ones in helix repo? Does it make more sense to use some kind of pointers for reference purpose?

I couldn't even find the json files in the helix repo. Let me look into this more. Yes, keeping them in sync will be an issue. I'm okay with abandoning this PR in it's current state.

PR #411, opened 2021-01-06
Handle NPE gracefully in Spectator and address race conditions for updating shard_map

gopalrajpurohit:


newpoo (commented):
Can you add some contexts about how NPE can happen and how this PR fixes the NPE issue? This will help reviewers review the change.

Another question: are all the changes in this PR required to fix NPE issue? if not, can we split it into multiple PRs? Basically I'd prefer to make 1 PR do 1 thing. This will make the changes less risky and easier to review.

gopalrajpurohit:
> Can you add some contexts about how NPE can happen and how this PR fixes the NPE issue? This will help reviewers review the change.
> 
> Another question: are all the changes in this PR required to fix NPE issue? if not, can we split it into multiple PRs? Basically I'd prefer to make 1 PR do 1 thing. This will make the changes less risky and easier to review.

No... all of these changes are not required for NPE issue to be addressed. I will update this PR soon with only minimal changes to address immediate PR and move other improvements over to another PR.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -267,19 +304,24 @@ private String getHostWithDomain(String host) {
 
   // update disabledHosts, return true if there is any changes
   private boolean updateDisabledHosts() {
-    HelixAdmin admin = helixManager.getClusterManagmentTool();
+    this.updateLock.lock();
```
newpoo: Which fields does this lock prorect? I was under the impression that the event callbacks won't be called concurrently by helix framework, thus no application side synchronization is required.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -138,6 +150,22 @@ private void generateShardConfig() {
...
+         * and potential deletion of a resource in between.
+         *
+         * Another situation where we may receive null externalView for a resource is in case where
+         * a resource is newly created but it's externalView has not yet been generated by helix
+         * controller.
+         *
+         * There can be quite a few race conditions which are difficult to enumerate and under such
+         * scenarios, we can't guarantee that externalView for a given resource exists at a specific
+         * moment. In such cases, it is safe to ignore such resource until it's externalView is
+         * accessible and not null.
```
newpoo, line 167: would it be possible that we get a null external view for an existing active resource? In that case, should we use the old shard map for this resource?

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -225,6 +258,10 @@ private void generateShardConfig() {
     // Write the config to ZK
     LOG.error("Generating a new shard config...");
 
+    /**
+     * TODO:
```
newpoo, line 263: Can you add your username to this and the above TODOs.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: That shouldn't be a case.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: Sure.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: I am not so sure about what helix framework will do when there are different event type notifications. In any case adding application side lock should be a no-op if helix framework does guarantee delivery of all event notifications in singe thread. If it doesn't, then above help prevent race condition.

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
```diff
@@ -130,15 +130,30 @@ public static void main(String[] args) throws Exception {
...
+
+    /**
+     * TODO: Improve this logic.
+     *
+     * This is an issue if the lock is released without the process going down. This can happen,
+     * if the client looses connection to zookeeper for a period of time. In such a scenario,
+     * the client doesn't have a way to stop spectator from processing shard_map generation.
+     * The lock may subsequently be obtained by another instance of spectator and hence breaking
+     * the semantics of single updator.
+     * In that case, the spectator will continue to listen and generate shard_map.
```
newpoo: Guodong is working on a fix for this via a connection state listener interface.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: Done.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
newpoo: I remembered I verified that Helix use single thread to run these callbacks. I'd prefer to not add it for code simplicity.

But if you still want to add this lock, can you add it to the top level of the callback functions? and make the lock name explicit that this lock is to synchronize callback functions. Otherwise, people may need to guess which fields this lock will protect.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
newpoo: good to know that.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: Done.

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
gopalrajpurohit: As. a long term solution, I would prefer a distributed controller approach and have Spectator instance per cluster be managed through LeaderStandby state model.

ghanpipi: we can treat the connection state listener as a short term fix

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
ghanpipi: can we add these discussion and explanation as part of the comment to help people understand it. ty

gopalrajpurohit: Added.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -51,20 +51,23 @@
...
-  private Set<String> disabledHosts;
-  private RocksplicatorMonitor monitor;
   private final boolean enableDumpToLocal;
+  private final Map<String, String> hostToHostWithDomain;
+  private final JSONObject dataParameters;
+  private final RocksplicatorMonitor monitor;
+  private final ReentrantLock synchronizedCallbackLock;
+  private HelixManager helixManager;
+  private volatile String lastPostedContent;
+  private volatile Set<String> disabledHosts;
```
newpoo, line 70: why volatile?
Previously the code is under the assumption that it is single threaded.
Now on top of that, we have a global lock for all the public interfaces of the class.

newpoo (approved):
lgtm, thanks for addressing my comments.

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
gopalrajpurohit: Not needed.

cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
gopalrajpurohit: As a long term solution: https://github.com/pinterest/rocksplicator/pull/412/ is a first step towards distributed spectator at par. with distributed controller.

PR #413, opened 2021-01-06
Add state transition handlers for LeaderFollower state model.

rajathprasad:
This diff adds the state transtion handlers for the LeaderFollower
state model. It's an exact copy of the MasterSlaveStateModelFactory
except that the state names have been replaced to adhere to inclusive
terminology guidelines.

This diff does not change the names of the "Role"s used in
rocksplicator. These will be done in the future. This diff also retains
the "M" and "S" indicators used in the shard map.

Testing: I deployed these changes in a binary and verified that I was
able to successfully initialize a new resource in the LeaderFollower
state model.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
@@ -164,8 +164,10 @@ public class ConfigGenerator extends RoutingTableProvider implements CustomCodeC
           String state = entry.getValue();
           if (!state.equalsIgnoreCase("ONLINE") &&
               !state.equalsIgnoreCase("MASTER") &&
+              !state.equalsIgnoreCase("LEADER") &&
+              !state.equalsIgnoreCase("FOLLOWER") &&
               !state.equalsIgnoreCase("SLAVE")) {
-            // Only ONLINE, MASTER and SLAVE states are ready for serving traffic
+            // Only ONLINE, MASTER, LEADER, FOLLOWER and SLAVE states are ready for serving traffic
             continue;
           }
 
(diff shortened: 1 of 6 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -464,6 +464,27 @@ public static boolean isMasterReplica(String host, int adminPort, String dbName)
...
+   * @param dbName
+   * @return
+   */
+  public static boolean isLeaderReplica(String host, int adminPort, String dbName) {
+    try {
+      Admin.Client client = getAdminClient(host, adminPort);
+
+      CheckDBRequest req = new CheckDBRequest(dbName);
+      CheckDBResponse res = client.checkDB(req);
+      // TODO: add a new field "is_leader" and use that instead.
```
gopalrajpurohit, line 480: Add usename against TODO

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -164,6 +164,8 @@ private void generateShardConfig() {
           String state = entry.getValue();
           if (!state.equalsIgnoreCase("ONLINE") &&
               !state.equalsIgnoreCase("MASTER") &&
+              !state.equalsIgnoreCase("LEADER") &&
+              !state.equalsIgnoreCase("FOLLOWER") &&
```
gopalrajpurohit, line 168: Update comment below...

rocksdb_admin/admin_handler.cpp
```diff
@@ -609,7 +609,7 @@ void AdminHandler::async_tm_addDB(
   std::string err_msg;
   replicator::DBRole role = replicator::DBRole::SLAVE;
   if (request->__isset.db_role) {
-    if (request->db_role == "SLAVE") {
+    if (request->db_role == "SLAVE" || request->db_role == "FOLLOWER") {
```
gopalrajpurohit, line 612: Can we Add DBRole::LEADER / FOLLOWER as well... so that we don't have to carryon the burden of using M/S terminology later on and doesn't becomes a mixed baggage... What that means is that it is either M/S or L/F ... Role...

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
rajathprasad: Done

rocksdb_admin/admin_handler.cpp
rajathprasad: I want to keep this PR limited to adding the new state transitions. Adding the new role would lead to changes in multiple places which I want to avoid in the PR.

gopalrajpurohit (approved):

PR #414, opened 2021-01-06
Use singleton HelixAdmin instance obtained from singleton HelixManager

gopalrajpurohit:


kangnanli:
why do we need this change? what's the benefit? can u add some description?

gopalrajpurohit:
> why do we need this change? what's the benefit? can u add some description?

helixManaget is effectively a singleton object. helixAdmin is derived from it. closing helixAdmin will also close any zkclient held by helixManager. Hence for a given helixManager, we should only contain one helixAdmin. Basically clean code.

ghanpipi (approved):

PR #415, opened 2021-01-06
Log all unchecked exceptions from shard generation

gopalrajpurohit:


ghanpipi (approved):

cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java
```diff
@@ -150,7 +169,14 @@ public void onConfigChange(List<InstanceConfig> configs, NotificationContext cha
   public void onExternalViewChange(List<ExternalView> externalViewList,
                                    NotificationContext changeContext) {
     try (AutoCloseableLock autoLock = AutoCloseableLock.lock(this.synchronizedCallbackLock)) {
-      generateShardConfig();
+      logUncheckedException(new Runnable() {
+        @Override
+        public void run() {
+          if (updateDisabledHosts()) {
```
rajathprasad, line 175: this also adds the updateDisabledHosts() check. Is that intentional?

PR #417, opened 2021-01-08
fix spectator hosts continue publish config maps even lost lock

ghanpipi:
In some scenarios where the leader spectator lost the zk connection, the
interprocess lock is released but the it continues to publish config map and the
new leader will also start to generate config map concurrently. We made the fix
by terminating the original leader process when a zk LOST event happened.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
@@ -33,6 +33,8 @@ import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.recipes.locks.InterProcessMutex;
 import org.apache.curator.framework.recipes.locks.Locker;
+import org.apache.curator.framework.state.ConnectionState;
+import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.retry.ExponentialBackoffRetry;
 import org.apache.helix.HelixConstants;
 import org.apache.helix.HelixManager;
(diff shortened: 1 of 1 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Spectator.java
```diff
@@ -129,6 +131,20 @@ public static void main(String[] args) throws Exception {
...
 
     CuratorFramework zkClient = CuratorFrameworkFactory.newClient(zkConnectString, new ExponentialBackoffRetry(1000, 3));
+
+    zkClient.getConnectionStateListenable().addListener(new ConnectionStateListener() {
+      @Override
+      public void stateChanged(CuratorFramework client, ConnectionState newState) {
+      // If the connection state changes to LOST, then it means it already lost the lock and
+      // the current leader role is not legal any more. Here we simply terminate it and rely
+      // on restarting to re-acquire the inter-process lock.
+        if (newState == ConnectionState.LOST) {
```
newpoo, line 141: according to the errorhandling section on http://curator.apache.org/curator-recipes/shared-lock.html We should commit suicide for SUSPENDED event as well.

Can we remove the TODO from line 153 to line 164 below? As it's not the case with this PR.

ghanpipi: I was thinking the SUSPENDED status will turn into RECONNECTED or LOST eventually, and only need to handle the SUSPENDED->LOST case where the lock is released. And you are right, to make it safe and clean, we should also commit suicide in SUSPENDED evnet.

newpoo (approved):
Can you please verify that it works as expected before rolling it out to production?

PR #419, opened 2021-01-14
Update shard map before notifying followers of new leader.

rajathprasad:
We have seen issues where a follower has its upstream set to a previous
leader. This causes the follower to not receive any updates from the
current leader, leading to inconsistent data being read by clients.

One of the causes for the above could be that the shard map is
potentially not updated for 10 minutes before notifying the followers.
Although there is a distributed lock for each partition and it is
unlikely that a new follower might have come up while a follower->leader
transition is taking place, with zookeeper involved it could be possible.
This is an easy fix to mitigate the issue to a certain extent.

I was unable to reproduce the issue on a test cluster. So will need to
test it once I land it.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -153,10 +153,11 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
       try (Locker locker = new Locker(partitionMutex)) {
         HelixAdmin admin = context.getManager().getClusterManagmentTool();
         Map<String, String> stateMap = null;
+        ExternalView view = null;
 
         // sanity check no existing Leader for up to 59 seconds
         for (int i = 0; i < 60; ++i) {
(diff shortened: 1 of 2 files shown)


ghanpipi (approved):

newpoo (commented):
The PR looks to be low risk to me. As you said, just not sure how to easily verify it if we can't reproduce the issue.

Just a quick idea. I remember there was a linkedin blog post (https://engineering.linkedin.com/blog/2020/improving-espresso-availability) mentioning that using current state for routing is faster than external view. I am wondering if relying on current state can help with setting up replication graph?

rajathprasad:
> The PR looks to be low risk to me. As you said, just not sure how to easily verify it if we can't reproduce the issue.
> 
> Just a quick idea. I remember there was a linkedin blog post (https://engineering.linkedin.com/blog/2020/improving-espresso-availability) mentioning that using current state for routing is faster than external view. I am wondering if relying on current state can help with setting up replication graph?

We are considering using external view for the spectator to improve availability. It's an interesting idea to use it for setting up the replication graph too. I'll look into it. In the meanwhile, since this is an easy fix and low risk, I'm going to land this and test it out.

PR #422, opened 2021-01-25
Upgrade java to version java 8

gopalrajpurohit:
We are already using helix-0.9.1 version, the library version that is using java 8. Hence we can't compile cluster_management with java 7 anyway.

### DIFF
--- a/cluster_management/pom.xml
+++ b/cluster_management/pom.xml
@@ -73,8 +73,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
   </properties>
 
   <dependencies>


kangnanli:
> We are already using helix-0.9.1 version, the library version that is using java 8. Hence we can't compile cluster_management with java 7 anyway.

If helix is already in 0.9.1 which won't compile with java 7, then, how does the cluster_management compilation works fine since the helix upgrade?

gopalrajpurohit:
> > We are already using helix-0.9.1 version, the library version that is using java 8. Hence we can't compile cluster_management with java 7 anyway.
> 
> If helix is already in 0.9.1 which won't compile with java 7, then, how does the cluster_management compilation works fine since the helix upgrade?

I should have been bit more careful with the wording. Since cluster_management depends on helix version 0.9.1 for which the source and target is set to java8, it needs java8 compiler to compile. Which means the library helix-*.jar we depend on in cluster_management is java8 compiled.

The source and target for cluster_management is set to version 7. Which means cluster_management's source code is to be compiled assuming it complies with java 7 spec. However if we try to compile cluster_management with java7 version, it may / may not compile the cluster_management. However the runtime for cluster_management MUST be java8 anyway, since it includes a helix classes that were compiled with java8.

In .travis, we use java8 for compiling cluster_management. Hence it will have no issues.. We also must have java runtime that complies with java 8 version compiled classes. So there is no point in assuming that cluster_management code is java 7. Moreover There is no further need to compile cluster_management code to meet java 7 spec.

kangnanli (approved):

PR #423, opened 2021-01-25
[Profiling Leader Handoff]: Implementation of zkstore and async read-modiy-write of LeaderEvents.

gopalrajpurohit:
This PR implements

1. LeaderEventsHistoryStore: This encapsulates a zkClient and creates a facade that provides async update through use of executor service. It internally manages instances of zkStore corresponding to each of the resource/partition. It also provides mechanism for pruning out any LeaderEvents that are deemed / guaranteed to be duplicate using a cache of localhistory that is up-to-date w.r.t. changes committed by the local process. It makes sure that only state changes are logged and stored in zk.


cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
```diff
@@ -0,0 +1,498 @@
...
+
+    this.zkClient =
+        CuratorFrameworkFactory.newClient(Preconditions.checkNotNull(this.zkConnectString),
+            new ExponentialBackoffRetry(1000, 3));
+
+    this.zkClient.start();
+    try {
+      this.zkClient.blockUntilConnected(60, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      LOGGER.error(String.format("Can't connect to zookeeper: %s", zkConnectString), e);
```
rajathprasad, line 102: Should we be continuing even if connection to zk is not successful?

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
```diff
@@ -0,0 +1,498 @@
...
+                });
+          }
+        });
+
+    this.executorServices = new ArrayList<>(NUM_PARALLEL_THREADS);
+
+    for (int i = 0; i < NUM_PARALLEL_THREADS; ++i) {
+      this.executorServices.add(Executors.newSingleThreadExecutor());
+    }
+    this.clusterResourcePartitionLock = new ArrayList<>(128);
```
rajathprasad, line 160: Why 128?

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
```diff
@@ -0,0 +1,498 @@
...
+        LeaderEventsHistory.class, SerializationProtocol.COMPACT, CompressionAlgorithm.GZIP);
+    this.maxEventsToKeep = maxEventsToKeep;
+
+    this.zkClient =
+        CuratorFrameworkFactory.newClient(Preconditions.checkNotNull(this.zkConnectString),
+            new ExponentialBackoffRetry(1000, 3));
+
+    this.zkClient.start();
+    try {
+      this.zkClient.blockUntilConnected(60, TimeUnit.SECONDS);
```
rajathprasad, line 100: Is it possible that calling this constructor could delay the state transition for up to a minute?

gopalrajpurohit: No, this will be created when the jvm is started... All updates to zk will be in. background threads... so that should not affect the state transitions themselves...

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
gopalrajpurohit: This should have been same as number of parallel threads used for updating zk. We use SingleThreadExecutor... hence we select which executor should run a given task of updating the zk leader events history for given resource.

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
gopalrajpurohit: No, if the zk provided is something we can't connect to when the process start... we can terminate the process instead... making sure correct zk is provided for storing leader events.. else it shouldn't be provided at all. Throwing runtime exception in this case.

rajathprasad (approved):
lgtm

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
```diff
@@ -0,0 +1,499 @@
...
+  private final CuratorFramework zkClient;
+  private final Optional<Integer> maxEventsToKeep;
+  private final Codec<LeaderEventsHistory, byte[]> leaderEventsHistoryCodec;
+  private final List<ExecutorService> executorServices;
+
+  private final LoadingCache<String, LoadingCache<String,
+      MergeableReadWriteStore<LeaderEventsHistory, LeaderEvent>>>
+      zkStoreCache;
+
+  private final LoadingCache<String, Cache<String, LeaderEventsHistory>> localHistoryCache;
```
kangnanli, line 79: what is the purpose of this "locakHistoryCache", I have not seen usage of this use in this PR.

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
```diff
@@ -0,0 +1,499 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BiConsumer;
+
+public class LeaderEventHistoryStore implements Closeable {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(LeaderEventHistoryStore.class);
+  private static final int NUM_PARALLEL_THREADS = 64;
```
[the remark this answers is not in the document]
gopalrajpurohit: I will run in both Participant as well as Spectator... Participant doesn't have default 40 threads. That is helix's default limit on number of threads used for callback to statemodels for state transitions. Here the LeaderEventHistoryStore maintains it's own executor service and provides async way to merge leader events to Participant (state transitions). So it will create it's own threads. Moreover we don't want to block participant state transitions while we merge data to zk for leader events history.

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/LeaderEventHistoryStore.java
gopalrajpurohit: line number 234.

It is used to prune out potentially duplicate events from locally known history of events. This helps remove any dups like in Spectator when it receives ExternalView, for a leader that never changes it's state... we could be publishing same event for same state of leader multiple times... We are not interested in such events. We are only interested in events when from an observer's perspective the state of a leader has changed.

PR #426, opened 2021-01-25
Implement zkstore to read / write data from zk with read-modify-write semantics

gopalrajpurohit:


gopalrajpurohit:
Note: The pom changes are required to perform proper exclusion of zk being pulled in with different version than required, causing test cases unrelated to this change to fail.

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ZkMergeableEventStore.java
```diff
@@ -0,0 +1,238 @@
...
+  private InterProcessMutex interProcessPartitionMutex;
+
+
+  public ZkMergeableEventStore(
+      CuratorFramework zkClient,
+      final String clusterName,
+      final String resourceName,
+      final String partitionName,
+      final Codec<R, byte[]> recordCodec,
+      final Supplier<R> zeroSupplier,
```
kangnanli, line 56: what is reason to have a zeroSupplier other than just 0?

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ZkMergeableEventStore.java
```diff
@@ -0,0 +1,238 @@
...
+public class ZkMergeableEventStore<R, E> implements MergeableReadWriteStore<R, E> {
+
+  private final AtomicBoolean pathExists;
+  private final String eventHistoryPath;
+  private final String partitionLockPath;
+  private final BatchMergeOperator<R> mergeOperator;
+  private final Supplier<R> zeroSupplier;
+  private final Codec<R, byte[]> recordCodec;
+  private final CuratorFramework zkClient;
+  private final Semaphore ipMutexGuard = new Semaphore(1);
```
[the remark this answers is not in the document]
gopalrajpurohit: This Semaphore is to limit at most 1 thread can execute this code in specific instance (specific to resource + partition) at 1 host.

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ZkMergeableEventStore.java
gopalrajpurohit: for integer zero is 0. For LeaderEventsHistory it is new LeaderEventsHistory() which has empty list of events...

gopalrajpurohit: This. store requires zeroSupplier, since it doesn't know what zero means for the template argument.

In our case, you can consider set of all LeaderEventsHistory as a monoid with the merge operator being the associative binary operatorion and identify element being new LeaderEventsHistory().

kangnanli (approved):
lgtm

PR #428, opened 2021-01-26
add fail count for configGenerator

kangnanli:
We have stats monitor to repot Null ExternalView during configGenerator. But, other exceptions might also caused the configGenerator to fail, which is not stats yet. 

Here, wrap the config generation code inside a high level try catch block, so config generation failure due to any exceptions, including the Null ExternalView will be stats.


gopalrajpurohit (approved):
lgtm...

PR #430, opened 2021-01-26
Fix running junit4 and disable testng to run junit tests

gopalrajpurohit:
After upgraing Junit to 4.x, we saw the junit tests were not running and only testng tests ran. In order to make sure both junit and testng tests run through their respective providers... this change is necessary.

### DIFF
--- a/cluster_management/pom.xml
+++ b/cluster_management/pom.xml
@@ -49,6 +49,32 @@
         </executions>
       </plugin>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


kangnanli:
Can u share why this solve the problem? How do you find the solution, can u share some links or details?

gopalrajpurohit:
> Can u share why this solve the problem? How do you find the solution, can u share some links or details?

https://maven.apache.org/surefire/maven-surefire-plugin/examples/testng.html

Basically searched and found this solution. I tested it locally, since .travis configuration we skip the tests.

kangnanli (approved):
lgtm

PR #432, opened 2021-01-31
Allow Spectator to publish leader events messages

gopalrajpurohit:

PR #435, opened 2021-02-01
Implementation for ExternalView based event logging for Spectator

gopalrajpurohit:
Breaking out implementation for ExternalView based logging of leader events from https://github.com/pinterest/rocksplicator/pull/432/ ... so that https://github.com/pinterest/rocksplicator/pull/432/ remains actual bindings in applications to enable logging from spectator.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ExternalViewLeaderEventsLoggerImpl.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ExternalViewLeaderEventsLoggerImpl.java
@@ -1,0 +1,322 @@
+package com.pinterest.rocksplicator.eventstore;
+
+import com.pinterest.rocksplicator.thrift.eventhistory.LeaderEventType;
+import com.pinterest.rocksplicator.utils.ExternalViewUtils;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.cache.RemovalListener;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.math.IntMath;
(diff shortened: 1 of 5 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ExternalViewLeaderEventsLoggerImpl.java
```diff
@@ -0,0 +1,322 @@
...
+                  || (previouslyKnownLeader.state == LeaderState.LEADER_UP
+                          && (previouslyKnownLeader.leaderInstanceId == null || !host
+                  .equals(previouslyKnownLeader.leaderInstanceId)))) {
+                /**
+                 * We publish the event as LEADER_DOWN if
+                 * either previously there was no information about this partition
+                 * or previously the known leader was up
+                 * or previously the leader host was not known
+                 * or if previous leader was known, it has changed now.
+                 */
```
rajathprasad, line 271: Can you update comments here to explain the LEADER_UP state?

rajathprasad (approved):
LGTM. The multiple if-else blocks were a bit confusing to read (especially on github without matching parenthesis), as a future task we should consider refactoring it.

cluster_management/src/main/java/com/pinterest/rocksplicator/eventstore/ExternalViewLeaderEventsLoggerImpl.java
gopalrajpurohit: If the spectator has just come up, it won't have cached previous leader known...
if previous event we found that some leader was up state..
or if (it was known to be down) ... but we didn't know which host was leader... hence update..
or if the known leader that is down has changed... In all of the above cases... there is a change in either known state or actual leader that is down... hence we publish the event...

Basic idea is to only report each of the state/leader only once, the first time it is encountered... and subsequently supress same kind of event... as it is immaterial. Next time we publish is when there is a material change in either the status of known leader or the leader itself changed...

PR #436, opened 2021-02-01
json shardmap conversion utililies to be used for posting client leader events.

gopalrajpurohit:
This basically converts a JSON shardmap config into internal strongly typed representation for easily iterating over all data for posting leader events from client's perspective. The shard_map file will be watched for changes and post the leader events when status of leader changes in the shard_map config.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/Instance.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/Instance.java
@@ -3,6 +3,7 @@ package com.pinterest.rocksplicator.shardmap;
 import java.util.Objects;
 
 public class Instance {
+
   private final String host;
   private final int port;
   private final String domain;
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/Instance.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/Instance.java
@@ -1,0 +1,54 @@
+package com.pinterest.rocksplicator.shardmap;
+
+import java.util.Objects;
+
+public class Instance {
+  private final String host;
+  private final int port;
+  private final String domain;
+  private final String instanceId;
+
+  public Instance(String hostWithDomainName) {
+    String[] parts = hostWithDomainName.split(":");
+    this.host = parts[0];
+    this.port = Integer.parseInt(parts[1]);
+    this.domain = parts[2];
+    this.instanceId = String.format("%s_%d", this.host, this.port);
+  }
+
+  public String getInstanceId() {
+    return instanceId;
+  }
+
+  public String getHost() {
+    return host;
+  }
+
+  public int getPort() {
+    return port;
+  }
+
+  public String getDomain() {
+    return domain;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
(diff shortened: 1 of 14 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/ResourceMap.java
```diff
@@ -0,0 +1,24 @@
+package com.pinterest.rocksplicator.shardmap;
+
+import java.util.List;
+import java.util.Set;
+import javax.swing.Painter;
```
kangnanli, line 5: why do we nee Painter here?

cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/Instance.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public int getPort() {
+    return port;
+  }
+
+  public String getDomain() {
+    return domain;
+  }
+
+  @Override
```
kangnanli, line 35: why we have Override here? Instance seem not inherit from other class.

cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/JsonShardMap.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+        String encodedInstanceStr = (String) key;
+        Instance instance = new Instance(encodedInstanceStr);
+
+        JSONArray partitionsOnInstance = (JSONArray) resourceMapObj.get(key);
+        for (Object encodedPartitionObj : partitionsOnInstance) {
+          String encodedPartitionStr = (String) encodedPartitionObj;
+          String parts[] = encodedPartitionStr.split(":");
+          Partition partition = new Partition(resourceName, parts[0]);
+          ReplicaState replicaState = ReplicaState.ONLINE;
```
kangnanli, line 87: why default as online?

cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/ResourceMap.java
gopalrajpurohit: My bad. IntelliJ auto import mistake. Will fix this.

cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/Instance.java
gopalrajpurohit: All Java classes implicitly extends Object class which is root of all java classes. This method is being overridden so that this works with containers like Map/Set.

cluster_management/src/main/java/com/pinterest/rocksplicator/shardmap/JsonShardMap.java
gopalrajpurohit: If the partition info doesn't contain ':' then it cannot be Master/Slave or Leader/Follower model. Since this entry exists in the JSON ShardMap, it must be ONLINE. However, this is for completeness, in representation of a Partition replica state. We are only interested in serving partitions. And shardmap only contains serving partitions. This is also what client's assume.

kangnanli (approved):

PR #438, opened 2021-02-03
Implement re-watchable ConfigNotifier class to drive notifications for config file change after decoding the raw bytes

gopalrajpurohit:
In order to publish client events, it needs a FileWatcher implementation that polls for timestamp modification every second.
Moreover current implementation of ConfigStore doesn't provide notification. It is a store that basically holds data after decoding. This is not sufficient for shard_map notification. Hence ConfigNotifier forwards notification from underlying poller and decodes the data and forwards it to notification callback. It also provides more context around the meta data of the file, like the timestamp of the file being watched when the poller polled for changes in the file, when it actually got notified, and how much time it took to deserialize raw bytes into format required by client. This is helpful in understanding the time lost in decoding raw bytes as well as the time the file actually got updated. This will help with understanding the latency once shard_map is posted to given url and then the contents are actually received on client side.


rajathprasad (approved):
Did a quick pass, lgtm

PR #439, opened 2021-02-03
[replication errors] Add a script to get the current leader of a partition

rajathprasad:
This is part 1 of the changes to fix the replication errors we are seeing in read/write clusters.

**Brief description of how we are planning to solve it:**
When a replication error is seen in rocksdb_replicator, we will query helix to get the latest leader
the partition, and reset the upstream to this latest leader.

This change contains a java script to return the IP of the leader's replica.
I also added a main function to make it easier for standalone testing. But we will be eventually
using the `getLeaderHostIP` function.

The partition_name currently includes the resource name too (since that's how helix stores it)
I will take a call later if it's cleaner to just take in the partition number.

Testing:
Ran the script with various input (and error) combinations:
1. Verified that the expected response is got for good cases.
2. Verified that an empty string was returned for all other cases (
invalid zk, cluster, resource, partition)

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/helix_client/HelixClient.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/helix_client/HelixClient.java
@@ -1,0 +1,167 @@
+/// Copyright 2017 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author rajathprasad (rajathprasad@pinterest.com)
+//
+
+package com.pinterest.rocksplicator.helix_client;
+
+import java.util.Map;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.GnuParser;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.OptionBuilder;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.helix.HelixAdmin;
+import org.apache.helix.manager.zk.ZKHelixAdmin;
+import org.apache.helix.model.ExternalView;
+import org.apache.log4j.BasicConfigurator;
+import org.apache.log4j.ConsoleAppender;
+import org.apache.log4j.Level;
+import org.apache.log4j.PatternLayout;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class HelixClient {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HelixClient.class);
+  private static final String zkServer = "zkSvr";
+  private static final String cluster = "cluster";
+  private static final String resource = "resource";
+  private static final String partition = "partition";
+
+  private static Options constructCommandLineOptions() {
+    Option zkServerOption =
+        OptionBuilder.withLongOpt(zkServer).withDescription("Provide zookeeper addresses").create();
+    zkServerOption.setArgs(1);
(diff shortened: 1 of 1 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Given the names of a cluster, resource and partition, this function returns the IP of the
+   * instance which hosts the leader replica of the partition. Returns an empty string in case
+   * an error or if there is no leader. 
+   */
+  public static String getLeaderHostIP(String zkConnectString,
+                                       String clusterName,
+                                       String resourceName,
+                                       String partitionName) {
```
gopalrajpurohit, line 99: Can we make sure rocksplicator-replicator code doesn't need to know about the partitionNaming format helix uses? Is is possible?

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      LOG.error("State map :" + stateMap);
+
+      if (!stateMap.containsValue("MASTER") && !stateMap.containsValue("LEADER")) {
+        LOG.error("No leader found");
+        return leaderHostIP;
+      }
+
+      for (Map.Entry<String, String> instanceNameAndRole : stateMap.entrySet()) {
```
gopalrajpurohit, line 128: !nitpick

```
for (Map.Entry<String, String> stateEntry: stateMap.entrySet()) {
  String state = stateEntry.get.Value();
  if (ExternalViewUtils.isMaster(state) || ExternalViewUtils.isLeader(state)) {
    leaderInstanceId = stateEntry.getKey();    // <- NOT THAT THIS IS HELIX INSTANCE ID OF THE PARTICIPANT and not just IP.
    break;
  }
}
```

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
```diff
@@ -0,0 +1,164 @@
...
+   * an error or if there is no leader. 
+   */
+  public static String getLeaderHostIP(String zkConnectString,
+                                       String clusterName,
+                                       String resourceName,
+                                       String partitionName) {
+    try {
+      LOG.error("Starting participant with ZK:" + zkConnectString);
+      HelixAdmin helixAdmin = new ZKHelixAdmin(zkConnectString);
+      String leaderHostIP = "";
```
gopalrajpurohit, line 103: s/leaderHostIP/leaderInstanceId/g

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
```diff
@@ -0,0 +1,164 @@
...
+   * instance which hosts the leader replica of the partition. Returns an empty string in case
+   * an error or if there is no leader. 
+   */
+  public static String getLeaderHostIP(String zkConnectString,
+                                       String clusterName,
+                                       String resourceName,
+                                       String partitionName) {
+    try {
+      LOG.error("Starting participant with ZK:" + zkConnectString);
+      HelixAdmin helixAdmin = new ZKHelixAdmin(zkConnectString);
```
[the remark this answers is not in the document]
rajathprasad: Thanks, done.

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
rajathprasad: In the replicator code we have only the DB name (which has different formatting compared to the partition name). So once I get to the replicator code I'll figure out the best way to handle this.

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
rajathprasad: Done, thanks.

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
rajathprasad: Done.

ghanpipi (approved):
lgtm

cluster_management/src/main/java/com/pinterest/rocksplicator/HelixClient.java
gopalrajpurohit: Thanks.

gopalrajpurohit (approved):

PR #441, opened 2021-02-03
Allow Client to publish leader events messages

gopalrajpurohit:


cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -77,10 +80,13 @@
   private static final String handoffEventHistoryzkSvr = "handoffEventHistoryzkSvr";
   private static final String handoffEventHistoryConfigPath = "handoffEventHistoryConfigPath";
   private static final String handoffEventHistoryConfigType = "handoffEventHistoryConfigType";
+  private static final String  handoffClientEventHistoryJsonShardMapPath = "handoffClientEventHistoryJsonShardMapPath";
 
   private static HelixManager helixManager = null;
   private static LeaderEventsLogger staticParticipantLeaderEventsLogger = null;
   private static LeaderEventsLogger staticSpectatorLeaderEventsLogger = null;
+  private static LeaderEventsLogger staticClientLeaderEventsLogger = null;
+  private static ClientShardMapLeaderEventLoggerDriver staticClientShardMapLeaderEventLoggerDriver = null;
```
kangnanli, line 89: What is this "Client" vs the Participant and Spectator ?

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -218,26 +235,31 @@ public static void main(String[] args) throws Exception {
...
-              zkEventHistoryStr, clusterName, resourceConfigPath, resourceConfigType,
-              Optional.of(128));
+      staticSpectatorLeaderEventsLogger = new LeaderEventsLoggerImpl(instanceName,
+          zkEventHistoryStr, clusterName, resourceConfigPath, resourceConfigType, Optional.of(128));
+
+      if (shardMapPath != null && !shardMapPath.isEmpty()) {
+        staticClientLeaderEventsLogger = new LeaderEventsLoggerImpl(instanceName,
+            zkEventHistoryStr, clusterName, resourceConfigPath, resourceConfigType, Optional.empty());
+
+        staticClientShardMapLeaderEventLoggerDriver = new ClientShardMapLeaderEventLoggerDriver(
```
kangnanli, line 255: can u remind us what is the "Driver" vs the EventsLogger?

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
gopalrajpurohit: In helix, usually the client is a Spectator that directly listens to helix. In case of Rocksplicator, we only have one Spectator and it is Spectator's responsibility to distribute the shard_map to all interesting parties. This way the client (e.g.. is a separate Root Layer) doesn't depend on helix. So in case of Rocksplicator, The root layer is the client which is interested in updated shard_map information for the purpose of routing requests to specific participant nodes. Participant is instance that actually holds some partitions (shards) and hence some state of the system (e.g.. storage node). Spectator is a translater. Now, in order  to log events from Client's perspective we need to watch the shard_map file that some Spectator has published for given cluster. Since we can't put this in the Root layer (which doesn't exist in rocksplicator itself and that we are not looking for all the clients observations)... we mimick this part of Client in Spectator. Since Spectator itself can run besides Participant, wherever Spectator runs, we also run a partial Client side mimicking the observer of Shard_map to log the delays introduced by in-between systems after Spectator posts the shard_map. This is important to profile, since this can be external system with not a well defined SLA or may not be suitable for the particular use case of distribution of latest shard_map to all interest parties.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
gopalrajpurohit: Driver here is driving the notification on changes in the file. LeaderEventsLogger is the main class common in all cases. We use special implementation that encapsulates LeaderEventsLogger for specific needs. E.g.. ClientShardMapLeaderEventsLogger will process ShardMap and convert it into interest events to log for LeaderEventsLogger. Similarly in case of Spectator, we use ExternalViewBasedLeaderEventsLogger that converts the ExternalView based notifications into interesting events. ExternalView is being driven by helix notifications. We don't have one for local file watching... so ClientShardMapLeaderEventLoggerDriver basically does that part of driving notifications from changes in shard_map file and forwarding it down the line.

kangnanli (approved):

PR #446, opened 2021-02-04
[WORK IN PROGRESS]: Allow participants to offline the partitions before exiting

gopalrajpurohit:
This is work in progress.. and need to be tested thoroughly. However I am putting this out early to gather any feedback.

Basically, we observe that when Participant JVM is shutting down, it simply disconnects the helixManager. This doesn't provide any chance to the participant to transition to offline all the existing top state / second-top state partition states
for a clean shutdown.

There are few improvements still that can be done.
1. disableTheInstance while going down only when a given flag says so... This is so that we don't cause an incidents during a rollback when this change goes out in production.
2. Avoid re-enabling an instance if there is an instanceTag that specifically denies it to be re-enabled. This is to ensure that if there is a maintenance activity to be performed to rotate out a Participant instance, a re-start of the Participant process on the instance the instance that is marked for removal is not re-enabled.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
@@ -18,18 +18,13 @@
 
 package com.pinterest.rocksplicator;
 
+import com.pinterest.rocksplicator.eventstore.ExternalViewLeaderEventsLoggerImpl;
 import com.pinterest.rocksplicator.eventstore.LeaderEventsLogger;
 import com.pinterest.rocksplicator.eventstore.LeaderEventsLoggerImpl;
-import com.pinterest.rocksplicator.eventstore.ExternalViewLeaderEventsLoggerImpl;
 import com.pinterest.rocksplicator.monitoring.mbeans.RocksplicatorMonitor;
 import com.pinterest.rocksplicator.task.BackupTaskFactory;
 import com.pinterest.rocksplicator.task.DedupTaskFactory;
 import com.pinterest.rocksplicator.task.RestoreTaskFactory;
-
-import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
@@ -44,7 +39,6 @@ import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
 import org.apache.helix.InstanceType;
 import org.apache.helix.manager.zk.HelixManagerShutdownHook;
-import org.apache.helix.manager.zk.ZKHelixAdmin;
 import org.apache.helix.model.HelixConfigScope;
 import org.apache.helix.model.Message;
 import org.apache.helix.model.builder.HelixConfigScopeBuilder;
@@ -60,6 +54,11 @@ import org.apache.log4j.Level;
 import org.apache.log4j.PatternLayout;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 
 
 public class Participant {
@@ -82,6 +81,8 @@ public class Participant {
   private static LeaderEventsLogger staticParticipantLeaderEventsLogger = null;
   private static LeaderEventsLogger staticSpectatorLeaderEventsLogger = null;
   private static StateModelFactory<StateModel> stateModelFactory;
+  private static String staticClusterName = null;
+  private static String staticInstanceId = null;
   private final RocksplicatorMonitor monitor;
 
   private static Options constructCommandLineOptions() {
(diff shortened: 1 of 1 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -444,4 +369,209 @@ public void run() {
...
+     * force controller to issue state transitions to existing current states of all partitions
+     * hosted by this participant) to move to OFFLINE state.
+     *
+     * 4. Wait until all partition replicas hosted by this participant to have moved to OFFLINE state.
+     *
+     * 5. Now we are free to disconnect to HelixManager, so that all listeners and message callback
+     * handlers are properly un-registered.
+     */
+    if (helixManager != null && helixManager.isConnected()) {
+      disableInstance(helixManager.getClusterManagmentTool(), staticClusterName, staticInstanceId);
```
gopalrajpurohit, line 406: should we otherwise,

disable instance
add disabled instance tag

wait for resources to be offlined
unregister the statemodelfactoriies
disconnect from helixManager.

//use separate helixAdmin, not one from helixManager above.
re-enable the instance (to ensure that when it comes up, the instance is already enabled)
remove the disabled instance-tag.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -444,4 +369,209 @@ public void run() {
...
+     * we need a chance to cleanly offline the currently top-state and second
+     * top-state resources (e.g.. all partitions state should be brought to offline
+     * state) before going down. That will be a graceful shutdown. However, when
+     * helixManager disconnects, it releases all message handlers / listeners there after
+     * there is no way for resources to be brought down through state transition.
+     *
+     * Ideal way will be whereby
+     *
+     * 1. We setup instanceTag "disabled" so that spectator has a first chance to be notified that
+     * this participant is going down without bringing down the hosted partition replicas.
```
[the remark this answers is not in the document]
gopalrajpurohit: The way currently Spectator is written, it relies solely on disable instance tag. That too is not complete as it depends on deploy script and environment setup correctly for Spectator to remove the host that is scheduled to go down.

I am also thinking of adding a flag to only disable the instance, if the flag is set. This way it won't affect currently deployed systems. Disabling instance basically allowed a clean shutdown. Currently this PR is not setup to actually monitor all online partitions to have offlined. That is bit of complicated change, so I wanted to keep that separate from this PR, and in this PR we simply wait for 5 seconds before exiting from jvm.

We will also modify Spectator to remove any disabled hosts from consideration for generating the shard_map. But I wanted to keep this PR as small as possible. Don't want to make too many changes in single PR.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -444,4 +369,209 @@ public void run() {
...
+        new ExceptionalCondition() {
+          @Override
+          public boolean apply() throws Exception {
+            // When the instance config tag contains the "disabled" tag.
+            return helixAdmin.getInstanceConfig(clusterName, instanceId).containsTag("disabled");
+          }
+        }, 10);
+
+    /**
+     * Here we need a mechanism to wait for all instances to go offline. However currently that logic
```
rajathprasad, line 473: Why is it complicated? Can we look at the current state for the instance and check if everything is offline?

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
```diff
@@ -444,4 +369,209 @@ public void run() {
...
+                    + " to force spectator to consider this instance in shard_map "
+                    + "potentially even befoer any partitions have been active ");
+            helixAdmin.removeInstanceTag(clusterName, instanceId, "disabled");
+          }
+        },
+        new ExceptionalCondition() {
+          @Override
+          public boolean apply() throws Exception {
+            // When the instance config tag contains the "disabled" tag.
+            return helixAdmin.getInstanceConfig(clusterName, instanceId).containsTag("disabled");
```
rajathprasad, line 504: Should this be "when the instance config tag does not contain the "disabled" tag" since we are trying to remove the tag?

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
gopalrajpurohit: Exactly.. that is too much of the code to do it along with this PR. I want to add that and other function like maintenance tag... so that if the node is to be taken down for scheduled maintenance or rotation... we don't allow participant to successfully deploy / restart on such instances..

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
gopalrajpurohit: great catch.

cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
gopalrajpurohit: Ok. I tried to get ZNRecord for the instance, but can't get anything out of it. There isn't any direct api available that provides the information needed to ensure that all partitions are offlined. 

One way is to iterate through all resources and their corresponding externalViews to make sure the instance doesn't have any non-offlined partition. However that approach is too risky, for one that externalViews are bulky in nature and getting all externalViews in a large cluster during deploy could potentially increase the zk load more then is necessary.

So, for now, we will wait it out for 5 seconds which seems sufficient time to get the data offlined.

In the same way, we also don't have any way to see if all partitions are fully operational. The reason is that when an instance comes up it may not have any assigned partition or controller may have rebalanced them to other participants. Hence there is no straight way to say that all partitions have moved from offline state to next.

We might want to consider enabling a delay once a statemodel has been registered with helixManager to give it a time
to online (or move them to follower, bootstrap or slave state transition) partitions for pre-existing data partitions.

gopalrajpurohit: I have tested on dev cluster to make sure this change does disable the instance and offline existing top state partitions when going down. While coming back up, it does wait to enable the instance and remove the disabled flag.

It also make sure that if "maintenance" instance tag has been applied to an instance, it won't allow the Participant to successfully online.

gopalrajpurohit:
Putting this PR on hold for a while.

Why?

This feature has a risk with rollback to older versions. If this change is deployed, the client's will have to necessarily perform multiple deploy cycles before enabling this feature in order to ensure the rollbacks never happen before this PR. Since this change will leave the state of an instance in disabled mode (if that feature is enabled), a currently deploy system if rolledback to previous version will not know that it needs to enable the instance (since that code didn't exist earlier).

To make this PR safe for rollback scenarios, I think that we shouldn't leave the state of instance in disabled mode while going down. The purpose for us to disable the instance was to force helix to offline the existing partitions. So, what we can do is

1. disable the instance
2. disconnect from helix
3. enable the instance back (using helixAdmin not associated with the helixManager that connects the participant) and then exit cleanly.

@newpoo : Your feedback is appreciated. Do you think we should go with above outlined approach?

newpoo:
Hi Gopal, thanks for seeking my suggestions. I didn't follow the recent development in rocksplicator helix code. Can you please work with the team to figure out the best way moving forward?

PR #448, opened 2021-02-09
Add a function in helix_client to call the getleaderInstanceId function.

rajathprasad:
This is a continuation of
https://github.com/pinterest/rocksplicator/commit/0d7f168a6850e18744257bee2a86270a011c8a8a

This PR adds a c++ function to call the `getleaderInstanceId` Java
function in the running JVM.

Testing done:
In a test service, I added a perioidic call to this new cpp function and
verified that the correct results were returned.

### DIFF
--- a/rocksdb_admin/helix_client.cpp
+++ b/rocksdb_admin/helix_client.cpp
@@ -85,6 +85,36 @@ JNIEnv* createVM(const std::string& class_path) {
   return env;
 }
 
+JNIEnv* getRunningVM(){
+    JNIEnv* env = nullptr;
+    JavaVM* jvm;
+    jsize nVMs;
+
+    // Get the a reference to the created javaVM
+    JNI_GetCreatedJavaVMs(NULL, 0, &nVMs);
+    JavaVM** buffer = new JavaVM*[nVMs];
+    JNI_GetCreatedJavaVMs(buffer, nVMs, &nVMs);
+
+    if (nVMs != 1) {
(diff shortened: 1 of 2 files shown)


kangnanli (approved):
lgtm. 

Does the running JVM allow concurrent access from different cpp method (or different cpp threads) at the same time?

rajathprasad:
> lgtm.
> 
> Does the running JVM allow concurrent access from different cpp method (or different cpp threads) at the same time?

Yes, I tested this on a test cluster with a 1000 partition resource. Didn't see any issues. Will test it more later.

gopalrajpurohit:
> lgtm.
> 
> Does the running JVM allow concurrent access from different cpp method (or different cpp threads) at the same time?

+1. Will be safe bet to obtain a lock to guarantee it is safe. We don't make much calls into jvm.. so it should not be issue with performace/ contention.

gopalrajpurohit (approved):
Apart from concern raised by @kangnanli , lgtm. Make sure usage is safe in multi-threaded access.

rajathprasad:
> +1. Will be safe bet to obtain a lock to guarantee it is safe. We don't make much calls into jvm.. so it should not be issue with performace/ contention.

Thanks @kangnanli  and @gopalrajpurohit. Reading the documentation (https://docs.oracle.com/javase/9/docs/specs/jni/invocation.html#attachcurrentthread) I dont see an explicit mention that it is not thread safe. I will do some more testing. As suggested, I might just add a lock to be safe.

PR #450, opened 2021-02-11
[rocksdb_admin] enable option to backup/restore with meta

kangnanli:
During state transition, such as Offline->Follower,  the Follower replica will get a full snapshot of its Leader. 
Here, we also make Follower to make a copy of its Leader's DBMetaData (db_name, s3_bucket, s3_path, etc.)
- this make sense, since Follower is literally get a full copy of its leader, it should also inherit its DBMetaData

**code change to this PR**:
- during backup, we read the Leader's DBMetadata and push to s3. 
- during restore, we get the rocksdb::BackupInfo list from s3, and restore from the latest version, and store the latest version's metadata locally to DBMetadata. 

*Impact to our view of mutable DB*:
- currently, mutable DB (read-write system) does not have a local DBMetaData; while readonly DB (ie. readonly system) has a local DBMetaData to store the s3_path of the serving data. 
- Here, we re-design our view of mutable DB:
- - each mutable DB will have two tiers: a mutable tier locally, and an immutable tier at S3. 
- - each mutable DB will have a local DBMetaData to store its 2nd tier data info (e.g. version, s3_path, etc.)
-- during serving, if we need to combine this two tiers, we will ingest the tier2 data into tier1 through ingestion process, and update the local DBMetaData to indicate the "combination process", e.g. ingested or not. 

*Useful*:
- this is useful: if we ingest data to the Leader and store the ingested data's s3_path into Leader's DBMetaData, then, when Follower get the full snapshot from its Leader. Since the "snapshot" already contains the ingested data, the follower should have its DBMetaDat indicate a "s3_path" to indicate it contains the ingested data.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -644,7 +644,7 @@ bool AdminHandler::backupDBHelper(const std::string& db_name,
                                   const bool enable_backup_rate_limit,
                                   const uint32_t backup_rate_limit,
                                   const bool share_files_with_checksum,
-                                  bool backup_with_meta,
+                                  bool include_meta,
                                   AdminException* e) {
   CHECK(env_holder != nullptr);
   db_admin_lock_.Lock(db_name);
@@ -679,7 +679,7 @@ bool AdminHandler::backupDBHelper(const std::string& db_name,
   }
   std::unique_ptr<rocksdb::BackupEngine> backup_engine_holder(backup_engine);
 
-  if (backup_with_meta) {
+  if (include_meta) {
     std::string db_meta;
     const auto meta = getMetaData(db_name);
     if (!EncodeThriftStruct(meta, &db_meta)) {
@@ -800,14 +800,14 @@ void AdminHandler::async_tm_backupDB(
   LOG(INFO) << "HDFS Backup " << request->db_name << " to " << full_path;
   AdminException e;
   const bool share_files_with_checksum = request->__isset.share_files_with_checksum && request->share_files_with_checksum;
-  const bool backup_with_meta = request->__isset.backup_with_meta && request->backup_with_meta;
+  const bool include_meta = request->__isset.include_meta && request->include_meta;
   if (!backupDBHelper(request->db_name,
                       full_path,
                       std::unique_ptr<rocksdb::Env>(hdfs_env),
                       request->__isset.limit_mbs,
                       request->limit_mbs,
                       share_files_with_checksum,
-                      backup_with_meta,
+                      include_meta,
                       &e)) {
     callback.release()->exceptionInThread(std::move(e));
     common::Stats::get()->Incr(kHDFSBackupFailure);
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -686,6 +686,7 @@ bool AdminHandler::backupDBHelper(const std::string& db_name,
       LOG(ERROR) << "Failed to encode DBMetaData";
       return false;
     } else {
+      LOG(INFO) << "Create new backup with encoded meta: " << db_meta;
       return backup_engine->CreateNewBackupWithMetadata(db->rocksdb(), db_meta).ok();
     }
   } else {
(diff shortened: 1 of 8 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -995,6 +1031,7 @@ void AdminHandler::async_tm_backupDBToS3(
                         request->__isset.limit_mbs,
                         request->limit_mbs,
                         false, // disable checksum support for s3 till checkpoint backup to s3 also support it
+                        false, // checkpoint backup does not support with meta yet
```
rajathprasad, line 1034: We will not be able to checkpoint feature for UMS tables? Is that okay?

rocksdb_admin/tests/CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ endforeach(testsourcefile ${TEST_SOURCES})
 
 add_executable(hbase_backup_integration_test hbase_backup_integration_test.cpp)
 target_link_libraries(hbase_backup_integration_test gtest rocksdb_admin boost_filesystem rocksdb_glogger kafka_consumer rdkafka++)
+# uncomment following line to include the integration test
+#add_test(NAME hbase_backup_integration_test COMMAND hbase_backup_integration_test)
```
rajathprasad, line 21: Any reason for the tests being commented out?

rocksdb_admin/admin_handler.cpp
```diff
@@ -719,8 +731,19 @@ bool AdminHandler::restoreDBHelper(const std::string& db_name,
   }
   std::unique_ptr<rocksdb::BackupEngine> backup_engine_holder(backup_engine);
 
+  std::vector<rocksdb::BackupInfo> backup_infos;
```
rajathprasad, line 734: How many backups are typically stored? Asking since this is a new operation and wondering if it'll affect performance.

rocksdb_admin/admin_handler.cpp
```diff
@@ -736,6 +759,17 @@ bool AdminHandler::restoreDBHelper(const std::string& db_name,
     return false;
   }
 
+  DBMetaData meta;
+  if (!DecodeThriftStruct(backup_infos.back().app_metadata, &meta)) {
```
rajathprasad, line 763: Is this safe if no app_metadata was backed up?

rocksdb_admin/tests/CMakeLists.txt
kangnanli: Yes, the integration test will indeed upload data to our s3 bucket. Thus, we need comment it out for package. But, for local actual test, we need to uncomment to run the test.

rocksdb_admin/admin_handler.cpp
kangnanli: It the app_metadata is empty, then, DBMetaData will be default constructed. This is ok, bcz, when we "writeMetaData", we write with local "db_name".

rocksdb_admin/tests/admin_handler_test.cpp
```diff
@@ -106,37 +108,66 @@ TEST(AdminHandlerTest, MetaData) {
...
+  // EXPECT_TRUE(handler.writeMetaData(db_name, s3_bucket, s3_path, timestamp_ms));
+
+  // meta = handler.getMetaData(db_name);
+  // EXPECT_EQ(meta.db_name, db_name);
+  // EXPECT_TRUE(meta.__isset.s3_bucket);
+  // EXPECT_EQ(meta.s3_bucket, s3_bucket);
+  // EXPECT_TRUE(meta.__isset.s3_path);
+  // EXPECT_EQ(meta.s3_path, s3_path);
+  // EXPECT_TRUE(meta.__isset.last_kafka_msg_timestamp_ms);
+  // EXPECT_EQ(meta.last_kafka_msg_timestamp_ms, timestamp_ms);
```
rajathprasad: Please fix (uncomment or remove) before landing.

rajathprasad (approved):

kangnanli (commented):
*Integration test*
The change is deployed to test cluster with S3 or HDFS as the cloud backbone to verify:
- back/restore with meta works ok during shard movement
- backup, restore API call through admin client is working ok. (ie. send RPC through a python admin client to test cluster to trigger the backup or restore).

kangnanli:
*Integration test*
The change is deployed to test cluster with S3 or HDFS as the cloud backbone to verify:
- back/restore with meta works ok during shard movement
- backup, restore API call through admin client is working ok. (ie. send RPC through a python admin client to test cluster to trigger the backup or restore).

ghanpipi (approved):

PR #454, opened 2021-02-16
Refactor Publisher code for shard_map

gopalrajpurohit:
We want to introduce a new publisher for publishing per resource level shard_map publishing to provided zk. In order to keep the upcoming changes clean, need to refactor existing publish code.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/DedupingShardMapPublisher.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/DedupingShardMapPublisher.java
@@ -1,0 +1,51 @@
+/// Copyright 2021 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+/// http://www.apache.org/licenses/LICENSE-2.0
+
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+//
+// @author Gopal Rajpurohit (grajpurohit@pinterest.com)
+//
+
+package com.pinterest.rocksplicator.publisher;
+
+import org.json.simple.JSONObject;
+import org.slf4j.Logger;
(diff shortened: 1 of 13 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/DedupingShardMapPublisher.java
```diff
@@ -0,0 +1,58 @@
...
+    this.lastPostedContent = null;
+  }
+
+  @Override
+  public void publish(
+      final Set<String> validResources,
+      final List<ExternalView> externalViews,
+      final JSONObject jsonShardMap) {
+    String newContent = jsonShardMap.toString();
+    if (lastPostedContent != null && lastPostedContent.equals(newContent)) {
```
rajathprasad, line 52: Might be missing this, but where is `lastPostedContent` updated with the latest content?

cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/ParallelShardMapPublisher.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public void publish(final Set<String> validResources,
+                      final List<ExternalView> externalViews,
+                      final T shardMap) {
+    for (ShardMapPublisher<T> shardMapPublisher : shardMapPublishers) {
+      try {
+        shardMapPublisher.publish(validResources, externalViews, shardMap);
+      } catch (Throwable throwable) {
+        // Ensure that failure of one publisher doesn't effect the other
```
rajathprasad, line 44: should we log the exception here?

cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/DedupingShardMapPublisher.java
gopalrajpurohit: Already fixed. Added unit tests for wrapper implementations as well.

cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/ParallelShardMapPublisher.java
gopalrajpurohit: Refreshed. logging exception here wouldn't be much useful, since the publisher implementation details are now known.. e.g.. which publisher failed? The implementation publishers themselves should be logging this information, which are deduping, httppost and localfiledump publishers... log individual states as per implementation.

rajathprasad (approved):

PR #455, opened 2021-02-17
Per Resource Compressed ShardMap publisher to ZK

gopalrajpurohit:
Depends on: https://github.com/pinterest/rocksplicator/pull/461
Tested with next PR: https://github.com/pinterest/rocksplicator/pull/456 . The publish is successful.


rajathprasad (approved):
Looks good!

PR #457, opened 2021-02-19
get the highest empty level of rocksdb

kangnanli:
Add a util function into applicationDB to get the highest empty level of RocksDB. 

This is useful to check readiness of backfill since Rocksdb's ingest_behind only work if the highest level is empty (ie. level6 is empty).

### DIFF
--- a/rocksdb_admin/tests/application_db_test.cpp
+++ b/rocksdb_admin/tests/application_db_test.cpp
@@ -1,0 +1,196 @@
+/// Copyright 2016 Pinterest Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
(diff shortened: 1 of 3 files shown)


rajathprasad (approved):

rocksdb_admin/tests/application_db_test.cpp
```diff
@@ -0,0 +1,196 @@
...
+using rocksdb::Status;
+using std::list;
+using std::make_unique;
+using std::pair;
+using std::string;
+using std::to_string;
+using std::unique_ptr;
+using std::vector;
+
+class SimpleMergeOperator : public AssociativeMergeOperator {
```
ghanpipi, line 54: why is the MergeOperator needed in the test?

kangnanli: Yes, it is not needed. Removed.

ghanpipi (approved):

PR #462, opened 2021-02-20
Zk Based shardMap downloader with java agent

gopalrajpurohit:
Depends on https://github.com/pinterest/rocksplicator/pull/461

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/config/PollingFileWatcher.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/config/PollingFileWatcher.java
@@ -63,9 +63,11 @@ public class PollingFileWatcher implements FileWatcher<byte[]> {
   private static final HashFunction HASH_FUNCTION = Hashing.md5();
   private static final int DEFAULT_POLL_PERIOD_MILLIS = 10000;
   private static final int DEFAULT_HIGH_RESOLUTION_POLL_PERIOD_MILLIS = 1000;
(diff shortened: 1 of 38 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/config/PollingFileWatcher.java
```diff
@@ -121,6 +123,21 @@ public static PollingFileWatcher defaultPerSecondWatcher() {
     return HIGH_RESOLUTION_INSTANCE;
   }
 
+  /**
+   * Creates the default ConfigFileWatcher instance on demand.
+   */
+  public static PollingFileWatcher defaultExpensiveHighResolutionWatcher() {
+    if (EXPENSIVE_HIGH_RESOLUTION_INSTANCE == null) {
```
rajathprasad, line 130: How is `EXPENSIVE_HIGH_RESOLUTION_INSTANCE` different? More frequent polling?

rajathprasad (approved):
lgtm

cluster_management/src/main/java/com/pinterest/rocksplicator/config/PollingFileWatcher.java
gopalrajpurohit: Yes... This is only used in testing... not in prod code. In the test, it is to ensure we get faster notification when the file is downloaded from zk... In prod we don't use this...

PR #463, opened 2021-02-20
Disable/Enable auto compaction through SetOptions

kangnanli:
Add test to verify:
- we can disable/re-enable auto compaction with SetOptions
note: we can also re-enable the auto compaction with EnableAutoCompaction. 

This is for:
disable/enable auto compaction for a segment before/after data backfill. 

Since rocksplicator already have setDBOptions API in rocksdb_admin.thrift, we will just use this API to disable/enable auto compaction.

### DIFF
--- a/rocksdb_admin/tests/application_db_test.cpp
+++ b/rocksdb_admin/tests/application_db_test.cpp
@@ -23,11 +23,14 @@
 #include <vector>
 
 #include "boost/filesystem.hpp"
+#include "gflags/gflags.h"
 #include "gtest/gtest.h"
 #include "rocksdb/db.h"
 #include "rocksdb/sst_file_writer.h"
 #include "rocksdb_admin/application_db.h"
 #include "rocksdb_replicator/rocksdb_replicator.h"
+
+DEFINE_bool(log_to_stdout, false, "Enable output some assisting logs to stdout");
 
 namespace admin {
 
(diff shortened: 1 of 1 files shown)


rocksdb_admin/tests/application_db_test.cpp
```diff
@@ -64,65 +77,169 @@ class ApplicationDBTestBase : public testing::Test {
...
+    last_options_ = options;
+    DB* db;
+    if (FLAGS_log_to_stdout) {
+      cout << "Open DB with db_path: " << db_path_ << endl;
+    }
+    auto status = DB::Open(options, db_path_, &db);
+    ASSERT_TRUE(status.ok()) << status.ToString();
+    ASSERT_TRUE(db != nullptr);
+    db_ = make_unique<ApplicationDB>(
+        db_name_, shared_ptr<DB>(db), replicator::DBRole::SLAVE, nullptr);
```
gopalrajpurohit, line 121: Will this always be SLAVE role?

kangnanli: By default, a new DB is always created as "S" role. 
For all the tests now in this test file, it only operate at DB level, it does not have any replication logic, so, it does not require different roles like "M" or "S". 

If future tests need rely on the replication, ie. need both "M" and "S", then, we can add another helper function to change role.

ghanpipi (approved):

PR #464, opened 2021-02-21
[rocksdb_replicator] Add a new thrift field to identify when a leader…

rajathprasad:
… db is moved.

We use the error code `SOURCE_NOT_FOUND` in two cases:
1. An instance has no information about the db in the replicate request.
2. An instance identifies it is no longer hostting the db requested.

Although the difference is subtle, we want to handle these two
situations differently: we would like to reset the leader only in the
first case (since a new leader might not yet be available in the second
case).

To handle this difference, I'm adding a new field to the `ErrorCode`
struct. Using the new field will be part of a subsequent PR.

### DIFF
--- a/rocksdb_replicator/thrift/replicator.thrift
+++ b/rocksdb_replicator/thrift/replicator.thrift
@@ -60,6 +60,7 @@ enum ErrorCode {
   OTHER = 0,
   SOURCE_NOT_FOUND = 1, # could not find the upstream db
   SOURCE_READ_ERROR = 2,
+  SOURCE_REMOVED = 3, # leader db has moved to a different instance
 }
 
 exception ReplicateException {


kangnanli:
I have not fully understand these two cases, can u elaborate:
1. what is "have no information"? At what condition or at which logic a instance to decide it "has no information"?
2. Is this when instance compare ExternalView with current hosting ApplicationDbs in applicationDBManager?

Also, what you mean by "since a new leader might not yet be available in the second case"? 
- do you mean, the time interval: after the old leader ->follower/offline, but, before follower promote to leader?

rajathprasad:
> I have not fully understand these two cases, can u elaborate:
> 
> 1. what is "have no information"? At what condition or at which logic a instance to decide it "has no information"?
> 2. Is this when instance compare ExternalView with current hosting ApplicationDbs in applicationDBManager?
> 
Every instance maintains an internal map of the DB's it hosts. Now, if a DB has been removed but still in the map, then the instance knows that is has been recently removed.
> Also, what you mean by "since a new leader might not yet be available in the second case"?
> 
> * do you mean, the time interval: after the old leader ->follower/offline, but, before follower promote to leader?
Yes, since it was very recently removed, the chances of a new leader being elected and available in external view is low.

kangnanli:
Can u still comment on:
What kinds of situation SOURCE_NOT_FOUND triggered?

I find two code reference about this error, it seems to be:
1. "has no information" https://github.com/pinterest/rocksplicator/blob/master/rocksdb_replicator/replicated_db.cpp#L266
2. db is removed: https://github.com/pinterest/rocksplicator/blob/master/rocksdb_replicator/replicator_handler.cpp#L34
Are these right?

kangnanli (approved):

rajathprasad:
Closing this PR since it's part of https://github.com/pinterest/rocksplicator/pull/471

PR #467, opened 2021-02-23
writeMetaData at all DB creation time

kangnanli:
Each Rocksplicator process has a metaDB inside its Admin Handler. It stores one entry for each DB it manages. 
For OnlineOffline/Boostrap state model, the MetaDB stores the db_name, and s3_path of the serving data. 

Here, we extend this to LeaderFollower statemodel (ie. read-write tables) as well, each DB, regardless of its statemodel:
- always have en initial entry (ie. only write db_name to meta_db) in Admin Handler's metaDB. 
- always write the s3_path to metadb if bootstrap or ingest from a s3_path
- new DB will inherit its leader's metaDAta (#450 )

note: for easy code review, I have a **FIXME** comment above each `writeMetaData` to indicate which following API it corresponds to. Will remove them when code reviewed. 

Our code is **comprehensive** to make sure all DB creation will have a `DBMetaData` write to `meta_db`:
There are 7 places in AdminHandler that a DB is DB::Open:
```
DB::open
- openMetaDB
- createDBBasedOnConfig (getRocskdbFuture) // it reload DBs from local config_shardmap when process restart
- addDB
- restoreDB (open after restore)
- restoreDBFromS3 with checkpoint
- AddS3SstFilestoDB
- clearDB with reopen
```
and 5 places in AdminHandler that a DB is added to ApplicationDBManager
```
db_manager_->addDB
- restoreDBfromS3 -checkpoint
- restoreDBHelper
- changeDBRoleAndUpstream (removeDb, then addDB; only role/upstream change, no need to update DBMetaData)
- clearDB
- addS3SstFilestoDB
```
And, we have add `writeMetaData` at 6 places, which corresponds to all places a DB::Open is called, except that AdminHandler reload the shardMap when process restart. 
```
Add DBMetadata (ie. writeMetaData)
- CreateDBBasedOnConfig (No need to writeMetaData since AdminHandler only reload the existing DB when process restart)
- admin::addDB
- restoreDB - checkpoint (write dummy DBmetaData till backup-checkpoint also backup the meta)
- restoreDBHelper (S3/hdfs)
- addS3SstFilesToDB 
- clearDB with reopen dB
- startMessageIngestion
```

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -1202,8 +1202,9 @@ void AdminHandler::async_tm_restoreDBFromS3(
     }
 
     // FIXME: restoreDBFromS3-checkpoint
-    if (!writeMetaData(request->db_name, request->s3_bucket, request->s3_backup_dir)) {
-      std::string errMsg = "RestoreDBFromS3 failed to write DBMetaData from request for " + request->db_name;
+    if (!writeMetaData(request->db_name, "", "")) {
+      // TODO: enable backup with meta for checkpoint, then, restore will writ the meta from the backup
+      std::string errMsg = "RestoreDBFromS3 failed to write DBMetaData for " + request->db_name;
       SetException(errMsg, AdminErrorCode::DB_ADMIN_ERROR, &callback);
       common::Stats::get()->Incr(kS3RestoreFailure);
       return;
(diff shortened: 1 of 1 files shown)


ghanpipi (approved):

kangnanli:
For test:
deploy build to test cluster with OnlineOffline or LeaderFollower statemodel. 
- verify data ingestion works ok for OnlineOffline
- verify new resource add ok for both
- verify shard movement ok for both

kangnanli:
This PR introduced a *bug* to OnlineOffline/Bootstrap statemodel, which ingest data from s3. 
The bug is fixed in https://github.com/pinterest/rocksplicator/pull/488

PR #469, opened 2021-02-23
support ingest behind for addS3SstFilesToDB api

kangnanli:
API: addS3SstFilesToDB already execute ingesting sst files to DB. By default, it is ingest ahead and used for boostrap DBs for OnlineOffline/Bootstrap statemodel. 
Here, we extend this API to ingest sst files behind when request.ingest_behine=true.
This is useful when we backfill data to existing DB, we will ingest sst files behind existing data.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -1420,9 +1420,6 @@ void AdminHandler::async_tm_clearDB(
     LOG(INFO) << "Open DB: " << request->db_name;
     admin::AdminException e;
     e.errorCode = AdminErrorCode::DB_ADMIN_ERROR;
-    if (request->__isset.allow_ingest_behind && request->allow_ingest_behind) {
-      options.allow_ingest_behind = true;
-    }
     auto db = GetRocksdb(db_path, options);
     if (db == nullptr) {
       e.message = "Failed to open DB: " + request->db_name;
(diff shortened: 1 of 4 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -1509,6 +1513,24 @@ void AdminHandler::async_tm_addS3SstFilesToDB(
...
+  if (ingest_behind) {
+    if (!db->rocksdb()->GetDBOptions().allow_ingest_behind) {
+      e.message = request->db_name + " DBOptions.allow_ingest_behind false";
+      LOG(ERROR) << "DBOptions.allow_ingest_behind false, can't ingest behind " << request->db_name;
+      callback.release()->exceptionInThread(std::move(e));
+      return;
+    }
+    if (db->getHighestEmptyLevel() != db->rocksdb()->NumberLevels() - 1) {
+      // note: default num levels for DB is 7 (0, 1, ..., 6)
+      std::string errMsg = "The Lmax of DB is not empty, skipp add files " + request->db_name;
```
rajathprasad, line 1526: Typo: skip

rajathprasad (approved):

rocksdb_admin/admin_handler.cpp
kangnanli: fixed it.

PR #471, opened 2021-02-24
[rocksdb_replicator] Reset upstream when the replication error is SOURCE_NOT_FOUND

rajathprasad:
…RCE_NOT_FOUND

This is the last diff in the task to fix replication errors on
followers. In this diff, we use the previously implemeneted function to
get the latest upstream for a follower.

Some specifics:
1. Since we do not want to overload the ZX server for external views,
only a small (right now 10) percentage of errors are querying helix.
In the rocksdb_replicator, there is already a delay in the next pull if
there is a replication error (default delay of 5 seconds). So the risk
of overloading ZK is low.
2. Add new a type of replication error: SOURCE_REMOVED. Prior to this
change, if a leader instance was aware that it was no longer hosting a
DB, it would return the same SOURCE_NOT_FOUND error. We want to
differentiate this case to further reduce load on ZK: SOURCE_REMOVED
errors are not eligible for upstream resets since a new leader might not
have been elected yet. We will rely on SOURCE_NOT_FOUND errors for
querying ZK.
3. Adding new gflags for zookeeper server string helix cluster name.
Instead of passing down the flags through a series of constructors, I'm
creating two gflags to set these values. In the service code, I will set
the values from glfags of the service. So, users of the service will be
agnostic of the new gflags.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -5,7 +5,9 @@ AUX_SOURCE_DIRECTORY(./ COMMON_FILES)
 
 add_library(common ${COMMON_FILES})
 
-target_link_libraries(common jemalloc glog gflags ssl wangle folly aws-cpp-sdk-core aws-cpp-sdk-s3 boost_system boost_filesystem thrift thriftcpp2 jsoncpp_lib_static zstd)
+target_link_libraries(common jemalloc jvm glog gflags ssl wangle
+        folly aws-cpp-sdk-core aws-cpp-sdk-s3 boost_system
+        boost_filesystem thrift thriftcpp2 jsoncpp_lib_static zstd)
 
 add_subdirectory(rocksdb_glogger)
 add_subdirectory(stats)
--- a/common/helix_client.cpp
+++ b/common/helix_client.cpp
@@ -16,7 +16,7 @@
 // @author bol (bol@pinterest.com)
 //
 
-#include "rocksdb_admin/helix_client.h"
+#include "common/helix_client.h"
 
 #include "jni.h"
 
@@ -26,7 +26,7 @@
 
 #include "common/network_util.h"
 
-DECLARE_int32(port);
+DEFINE_int32(port, 9090, "Port of the server");
 DEFINE_string(helix_log_config_file_path, "",
     "Participant log config file path "
     "e.g: /etc/config/helix/log4j.properties");
@@ -204,7 +204,7 @@ std::string JStringToString(JNIEnv* env, jstring jstr) {
 
 } // namespace
 
-namespace admin {
+namespace common {
 
 void JoinCluster(const std::string& zk_connect_str,
                  const std::string& cluster,
@@ -306,4 +306,4 @@ std::string GetLeaderInstanceId(
     return JStringToString(env, jstr);
 }
 
-}  // namespace admin
+}  // namespace common
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -22,7 +22,10 @@
 #include <string>
 #include <vector>
 
+#include "common/helix_client.h"
+#include "common/segment_utils.h"
 #include "folly/MoveWrapper.h"
+#include "folly/Random.h"
 #include "rocksdb_replicator/replicator_stats.h"
 #include "rocksdb_replicator/rocksdb_replicator.h"
 
@@ -53,6 +56,11 @@ DECLARE_int32(replicator_idle_iter_timeout_ms);
 DEFINE_bool(emit_stat_for_leader_behind,
             false,
             "Flag to control whether to emit a stat when the leader is behind the follower during a sync request.");
+DEFINE_string(replicator_zk_cluster, "", "Zookeeper cluster");
+DEFINE_string(replicator_helix_cluster, "", "Helix cluster");
+DEFINE_int32(replication_error_reset_upstream_percentage, 10,
+             "what percentage of replication errors should query helix for the latest leader");
+DECLARE_int32(rocksdb_replicator_port);
 
 
 namespace {
@@ -141,6 +149,44 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
           + FLAGS_replicator_client_server_timeout_difference_ms));
 }
 
+/*
+ * Helper function to reset the upstream IP after querying for latest leader from helix.
+ */
+void RocksDBReplicator::ReplicatedDB::resetUpstream() {
+  if (FLAGS_replicator_zk_cluster.empty() || FLAGS_replicator_helix_cluster.empty()) {
+    LOG(ERROR) << "[resetUpstream] ZK cluster or helix cluster name not provided.";
+    return;
+  }
+
+  // Query helix only FLAGS_replication_error_reset_upstream_percentage %ge of the requests.
+  // This is to avoid excessive load to Helix (zk)
+  if (folly::Random::rand32(0, 99) >= FLAGS_replication_error_reset_upstream_percentage) {
+    LOG(ERROR) << "[resetUpstream] Skip checking latest upstream for " << db_name_;
+    return;
+  }
+
+  auto segment_name = common::DbNameToSegment(db_name_);
+  auto helix_partition_name = common::DbNameToHelixPartitionName(db_name_);
+  LOG(ERROR) << "[resetUpstream] Zookeeper: " << FLAGS_replicator_zk_cluster << " cluster: " <<
+    FLAGS_replicator_helix_cluster << " segment: " << segment_name
+     << " helix_partition_name: " << helix_partition_name;
+  auto leader_id = common::GetLeaderInstanceId(
+    FLAGS_replicator_zk_cluster, FLAGS_replicator_helix_cluster, segment_name, helix_partition_name);
+  LOG(ERROR) << "[resetUpstream] Leader for " << helix_partition_name << " is " << leader_id;
+
+  if (!leader_id.empty()) {
+    std::vector<std::string> tokens;
+    folly::split("_", leader_id, tokens);
+    if (tokens.size() == 2 && tokens[0] != upstream_addr_.getAddressStr()) {
+      incCounter(kReplicatorLeaderReset, 1, db_name_);
+      LOG(ERROR) << "[resetUpstream] Resetting upstream for " << db_name_ << " to " << tokens[0];
+      upstream_addr_.setFromIpPort(tokens[0], FLAGS_rocksdb_replicator_port);
+      // Update client with new upstream address.
+      client_ = client_pool_->getClient(upstream_addr_);
+    }
+  }
+}
+
 void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
   CHECK(role_ == DBRole::SLAVE);
   ReplicateRequest req;
@@ -168,9 +214,16 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
             t.exception().throwException();
 #endif
           } catch (const ReplicateException& ex) {
-            LOG(ERROR) << "ReplicateException: " << static_cast<int>(ex.code)
+            LOG(ERROR) << "ReplicateException: (upstream): " << db->upstream_addr_.getAddressStr() << " " << static_cast<int>(ex.code)
                        << " " << ex.msg;
             incCounter(kReplicatorRemoteApplicationExceptions, 1, db->db_name_);
+            
+            if (ex.code == ErrorCode::SOURCE_NOT_FOUND) {
+              // This could happen if this db missed a request about the latest upstream.
+              // So try to reset it.
+              incCounter(kReplicatorRemoteApplicationExceptionsNotFound, 1, db->db_name_);
+              db->resetUpstream();
+            }
           } catch (const std::exception& ex) {
             LOG(ERROR) << "std::exception: " << ex.what();
             incCounter(kReplicatorConnectionErrors, 1, db->db_name_);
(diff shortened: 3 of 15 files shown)


rocksdb_replicator/replicated_db.cpp
```diff
@@ -141,6 +148,42 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
           + FLAGS_replicator_client_server_timeout_difference_ms));
 }
 
+/*
+ * Helper function to reset the upstream IP after querying for latest leader from helix.
+ */
+void RocksDBReplicator::ReplicatedDB::resetUpstream(std::shared_ptr<ReplicatedDB> db) {
+  if (FLAGS_replicator_zk_cluster.empty() || FLAGS_replicator_helix_cluster.empty()) {
+    LOG(ERROR) << "[resetUpstream] ZK cluster or helix cluster name not provided.";
```
newpoo, line 156: return?

rocksdb_replicator/rocksdb_replicator.h
```diff
@@ -127,7 +128,7 @@ class RocksDBReplicator {
     std::shared_ptr<rocksdb::DB> db_;
     folly::Executor* const executor_;
     const DBRole role_;
-    const folly::SocketAddress upstream_addr_;
+    folly::SocketAddress upstream_addr_;
```
newpoo, line 131: I don't fully remember the threading model for this code. Will take a closer look at it.

Can you please think through its threading model and the interaction between replicatedDB, replicator and db manager in case you haven't yet.

newpoo (commented):
flush some quick comments.

rocksdb_replicator/replicated_db.cpp
```diff
@@ -141,6 +148,42 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
...
+  }
+
+  // Query helix only FLAGS_replication_error_reset_upstream_percentage %ge of the requests.
+  // This is to avoid excessive load to Helix (zk)
+  if (common::GetRandInt(0, 99) >= FLAGS_replication_error_reset_upstream_percentage) {
+    LOG(ERROR) << "[resetUpstream] Skip checking latest upstream for " << db->db_name_;
```
kangnanli, line 162: when replication_error happen, what kind of rate of error is? 
Will this write too many error logs into logfile?

rocksdb_replicator/thrift/replicator.thrift
```diff
@@ -60,6 +60,7 @@ enum ErrorCode {
   OTHER = 0,
   SOURCE_NOT_FOUND = 1, # could not find the upstream db
   SOURCE_READ_ERROR = 2,
+  SOURCE_REMOVED = 3,
```
kangnanli, line 63: where/when will we use this in the code?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -152,7 +195,7 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
   std::weak_ptr<ReplicatedDB> weak_db = shared_from_this();
   auto options = rpc_options_;
   client_->future_replicate(options, req).via(executor_)
-    .then([weak_db = std::move(weak_db)] (folly::Try<ReplicateResponse>&& t) {
+    .then([weak_db = std::move(weak_db), this] (folly::Try<ReplicateResponse>&& t) {
```
kangnanli, line 198: why need to pass "this"? I did not see how to use it inside the lambda.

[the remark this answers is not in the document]
rajathprasad: Since `resetUpstream()` is part of the class, I needed to capture `this`. Do you have a suggestion on a better way to do it? Thanks for the info about the weak_ptr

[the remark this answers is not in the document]
rajathprasad: Oh, I guess I can do `db->resetUpstream()` instead.

rocksdb_replicator/replicated_db.cpp
rajathprasad: Error rate is quite high, but once an error is seen it is not retried for 5 seconds, so I think this should be okay.

rocksdb_replicator/replicated_db.cpp
rajathprasad: Done, thanks for the catch.

rocksdb_replicator/rocksdb_replicator.h
rajathprasad: Yes, will look into it.

rocksdb_replicator/thrift/replicator.thrift
rajathprasad: Since both the thrift file change and using the new field cannot be part of the same diff (will cause problems during rollout), I removed the part where this is used. This will be used [here](https://github.com/pinterest/rocksplicator/blob/master/rocksdb_replicator/replicated_db.cpp#L266)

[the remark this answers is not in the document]
rajathprasad: Done

common/random_number_util.h
```diff
@@ -0,0 +1,30 @@
...
+//
+// Utility declares and functions for time.
+
+#pragma once
+
+#include <random>
+
+namespace common {
+
+int GetRandInt(int low, int high) {
```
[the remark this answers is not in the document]
rajathprasad: Done, thanks for the suggestion

rocksdb_admin/utils.cpp
```diff
@@ -47,4 +47,12 @@ int ExtractShardId(const std::string& db_name) {
   }
 }
 
+/*
+ * Utility function to convert a rocskdb instance name (test00100) to helix
+ * partition name (test_100)
+ */
```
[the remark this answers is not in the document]
rajathprasad: Done

common/CMakeLists.txt
```diff
@@ -5,7 +5,9 @@ AUX_SOURCE_DIRECTORY(./ COMMON_FILES)
 
 add_library(common ${COMMON_FILES})
 
-target_link_libraries(common jemalloc glog gflags ssl wangle folly aws-cpp-sdk-core aws-cpp-sdk-s3 boost_system boost_filesystem thrift thriftcpp2 jsoncpp_lib_static zstd)
+target_link_libraries(common jemalloc jvm glog gflags ssl wangle
```
rajathprasad, line 8: Added jvm

ghanpipi (approved):

rocksdb_replicator/rocksdb_replicator.h
rajathprasad: @newpoo Didn't find any issues with removing the const here. Application DB stores a copy of the upstream, and it will be out of sync with the upstream here, but that's not used once it is set, so it should be fine. Even within the replicator, once the upstream is set, it is not used for anything but sending replication requests, so should be fine too. Please lmk what you think

common/helix_client.cpp
```diff
@@ -26,7 +26,7 @@
 
 #include "common/network_util.h"
 
-DECLARE_int32(port);
+DEFINE_int32(port, 9090, "Port of the server");
```
newpoo, line 29: why do we need this change?

rocksdb_replicator/thrift/replicator.thrift
```diff
@@ -60,6 +60,7 @@ enum ErrorCode {
   OTHER = 0,
   SOURCE_NOT_FOUND = 1, # could not find the upstream db
   SOURCE_READ_ERROR = 2,
+  SOURCE_REMOVED = 3, # upstream db was recently removed
```
newpoo, line 63: Can you please add some comments that this is an optimization to reduce the helix reads? And it's no harm if the upstream fails to differentiate between SOURCE_NOT_FOUND VS. SOURCE_REMOVED.

rocksdb_replicator/replicated_db.cpp
```diff
@@ -23,6 +23,9 @@
 #include <vector>
 
 #include "folly/MoveWrapper.h"
+#include "folly/Random.h"
+#include "common/helix_client.h"
+#include "common/segment_utils.h"
```
newpoo, line 28: sort them?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -141,6 +148,43 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
...
+  auto segment_name = common::DbNameToSegment(db->db_name_);
+  auto helix_partition_name = common::DbNameToHelixPartitionName(db->db_name_);
+  LOG(ERROR) << "[resetUpstream] Zookeeper: " << FLAGS_replicator_zk_cluster << " cluster: " <<
+    FLAGS_replicator_helix_cluster << " segment: " << segment_name
+     << " helix_partition_name: " << helix_partition_name;
+  auto leader_id = common::GetLeaderInstanceId(
+    FLAGS_replicator_zk_cluster, FLAGS_replicator_helix_cluster, segment_name, helix_partition_name);
+  LOG(ERROR) << "[resetUpstream] Leader for " << helix_partition_name << " is " << leader_id;
+
+  if (leader_id.length() > 0) {
```
newpoo, line 176: nit: !leader_id.empty()

rocksdb_replicator/replicated_db.cpp
```diff
@@ -141,6 +148,43 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
...
+    std::vector<std::string> tokens;
+    folly::split("_", leader_id, tokens);
+    if (tokens.size() == 2 && tokens[0].compare(db->upstream_addr_.getAddressStr()) != 0) {
+      incCounter(kReplicatorLeaderReset, 1, db->db_name_);
+      LOG(ERROR) << "[resetUpstream] Resetting upstream for " << db->db_name_ << " to " << tokens[0];
+      uint16_t port = atoi(tokens[1].c_str());
+      db->upstream_addr_.setFromIpPort(tokens[0], port);
```
newpoo, line 183: I guess leader_id got from helix contains service port, while replicator uses a separate port for replication.
Sorry, this is a bit confusion.
You may take a look at this code: https://github.com/pinterest/rocksplicator/blob/b70224c1e3f1b8ff7ba1d0b854cacb5efa5b7263/rocksdb_admin/admin_handler.cpp#L281

rocksdb_replicator/replicated_db.cpp
```diff
@@ -141,6 +148,43 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
...
+    FLAGS_replicator_helix_cluster << " segment: " << segment_name
+     << " helix_partition_name: " << helix_partition_name;
+  auto leader_id = common::GetLeaderInstanceId(
+    FLAGS_replicator_zk_cluster, FLAGS_replicator_helix_cluster, segment_name, helix_partition_name);
+  LOG(ERROR) << "[resetUpstream] Leader for " << helix_partition_name << " is " << leader_id;
+
+  if (leader_id.length() > 0) {
+    std::vector<std::string> tokens;
+    folly::split("_", leader_id, tokens);
+    if (tokens.size() == 2 && tokens[0].compare(db->upstream_addr_.getAddressStr()) != 0) {
```
newpoo, line 179: nit:

`tokens.size() == 2 && tokens[0] != db->upstream_addr_.getAddressStr()`

rocksdb_replicator/replicated_db.cpp
```diff
@@ -215,15 +266,18 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
           auto eb = db->client_->getChannel()->getEventBase();
           // It is very bad if we fail to rescheudle a pull request, we'd prefer
           // crashing.
-          eb->runInEventBaseThread([eb, weak_db = std::move(weak_db)] {
-              eb->runAfterDelay([weak_db = std::move(weak_db)] {
+          auto delay = FLAGS_replicator_pull_delay_on_error_ms;
+          // Randomize the delay so that helix (zk) is not overloaded from ext view requests.
+          auto randomized_delay = folly::Random::rand32(delay, delay * 2);
+          eb->runInEventBaseThread([eb, weak_db = std::move(weak_db), randomized_delay] {
+              eb->runAfterDelay([weak_db = std::move(weak_db), randomized_delay] {
```
newpoo, line 273: this lambda doesn't need to capture randomized_delay?

how about we just compute randomized_delay in the lambda and avoid capturing it in any of the lambdas

newpoo (commented):
overall lgtm, a few minor comments.

rocksdb_replicator/replicated_db.cpp
```diff
@@ -168,9 +212,16 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
...
-            LOG(ERROR) << "ReplicateException: " << static_cast<int>(ex.code)
+            LOG(ERROR) << "ReplicateException: (upstream): " << db->upstream_addr_.getAddressStr() << " " << static_cast<int>(ex.code)
                        << " " << ex.msg;
             incCounter(kReplicatorRemoteApplicationExceptions, 1, db->db_name_);
+            
+            if (ex.code == ErrorCode::SOURCE_NOT_FOUND) {
+              // This could happen if this db missed a request about the latest upstream.
+              // So try to reset it.
+              incCounter(kReplicatorRemoteApplicationExceptionsNotFound, 1, db->db_name_);
+              db->resetUpstream(db);
```
newpoo, line 223: do you need to update client_ somewhere?

common/helix_client.cpp
rajathprasad: In general, the refactor is because of the new dependency of rocksdb_replicator on helix_client. Since rocksdb_admin already depends on rockdb_replicator, it would've created a circular dependency without this refactor. So I moved helix_client to common.
In particular, I had to define the PORT here since this is the common file, all other files which need to use this flag will include this common file.

rocksdb_replicator/replicated_db.cpp
rajathprasad: Oh I had missed this, thanks

rocksdb_replicator/replicated_db.cpp
rajathprasad: Done

newpoo (approved):
Looks great, thanks!

PR #473, opened 2021-02-25
Update ShardMapPublisherBuilder to alow zk based publishing

gopalrajpurohit:


cluster_management/src/main/java/com/pinterest/rocksplicator/publisher/ZkBasedPerResourceShardMapPublisher.java
```diff
@@ -285,6 +284,9 @@ private void removeResourceFromZkGZIPCompressedPerResourceShardMap(
             continue;
           }
 
+          // First remove the local cache
+          this.latestResourceToConfigMap.remove(resourceName);
```
gopalrajpurohit, line 288: This is a bug fix... since we have a local per resource shardmap... we have to ensure that it is also part of cleanup process...

PR #476, opened 2021-02-25
Enable Embedded Spectator to publish zk data and Participant to download zk data to local file

gopalrajpurohit:
This won't work in following scenario

If the Application  api is directly inheriting from Rockdb_Admin api, and it supports routing layer within participant cluster, then the startup sequence for the application will be

1. Create a Router (this requires shardMap)
2. Create serviceHandler (requires router)
3. Create Server routing calls to ServiceHandler (we may potentially start getting requests here)
4. Start JVM Participant which will then download the shardMap as first step, which was required in step no. 1.

Ideal scene would have been, that application service api not extend from rocksdb_admin service api and create separate servers for both..

1. Create RocksdbAdmin handler
2. Start RocksdbAdmin server routing rocksdb_admin api calls to RockdbAdmin handler (now we are ready to serve admin requests)
3. Start JVM Participant, which starts with downloading the shardMap and joins helix cluster as participant and start performing state transitions in background.
4. Wait for shardMap to be downloaded
5. Create Router (we now have shardMap)
6. Create ServiceHandler for application api
7. Create Application Server routing api calls to ServiceHandler (now we are ready to receive api calls for application)

Unfortunately... the design pattern for existing application already do not confirm to proper startup sequence.. 

Hence it is not possible to embedd zk data download logic in the embedded jvm.. Other option is to deploy the zk-shardmap download agent as a side-car with the participant cluster as well.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
@@ -24,10 +24,12 @@ import com.pinterest.rocksplicator.eventstore.LeaderEventsLoggerImpl;
 import com.pinterest.rocksplicator.eventstore.ExternalViewLeaderEventsLoggerImpl;
 import com.pinterest.rocksplicator.monitoring.mbeans.RocksplicatorMonitor;
 import com.pinterest.rocksplicator.publisher.ShardMapPublisherBuilder;
+import com.pinterest.rocksplicator.shardmapagent.ClusterShardMapAgent;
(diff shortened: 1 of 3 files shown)


gopalrajpurohit:
As explained in the description, this mechanism with shardMap agent embedded in participant won't work without significant refactoring and also depends upon applications code changes... Hence Abandoning this approach altogether. Side card solution works perfectly fine with participants as well.. Hence no need to over-engineer a solution for embedded scenario here.

PR #478, opened 2021-02-26
[not ready review] add optional to indicate create DB with allow_ingest_behind

kangnanli:
There are two admin APIs that might create DB, which need to know whether create DB with allow_ingest_behind or not.


kangnanli:
Close this request. DB's options should be set at the `RocksDBOptionsGeneratorType`, instead  of adjust based on the request object.

PR #480, opened 2021-02-28
add ingestTask(factory) to support ingest data with helix task framework

kangnanli:
This change added ingestTask(Factory) into rocksplicator, specifically, it adds 3 files:
- ingestTaskFactory, which create ingestTask based on helixJobConfigs, which is created through helixRest api to helix-zoookeeper. The helixJobConfigs contain info: which s3_path of the ingested data, whether ingest_behind, ... 
- ingestTask, which is the unit to execute the ingestion. It get the externalView of the resource, since 1 partition will have 1 task, then each Task manage the ingestion to all the replicas of the same partition (ie. 3 replicas). The actual execution will be cpp code `AddS3SstfilesToDB`.
- TestIingestTaskFactory, test logic inside above two files. 

others are formatter change due to clang-format. We will introduce a checkfile later to unify the style.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -19,6 +19,7 @@
 package com.pinterest.rocksplicator;
 
 import com.pinterest.rocksdb_admin.thrift.AddDBRequest;
+import com.pinterest.rocksdb_admin.thrift.AddS3SstFilesToDBRequest;
 import com.pinterest.rocksdb_admin.thrift.Admin;
 import com.pinterest.rocksdb_admin.thrift.AdminErrorCode;
 import com.pinterest.rocksdb_admin.thrift.AdminException;
@@ -287,7 +288,7 @@ public class Utils {
       CheckDBRequest req = new CheckDBRequest(dbName);
       return client.checkDB(req);
     } catch (TException e) {
-      LOG.error("Failed to check DB: ", e.toString());
+      LOG.error("Failed to check DB: ", e);
       throw new RuntimeException(e);
     }
   }
@@ -310,7 +311,7 @@ public class Utils {
       req.setInclude_meta(true);
       client.backupDB(req);
     } catch (TException e) {
-      LOG.error("Failed to backup DB: ", e.toString());
+      LOG.error("Failed to backup DB: ", e);
       throw new RuntimeException(e);
     }
   }
@@ -327,7 +328,7 @@ public class Utils {
       req.setShare_files_with_checksum(shareFilesWithChecksum);
       client.backupDB(req);
     } catch (TException e) {
-      LOG.error("Failed to backup DB: ", e.toString());
+      LOG.error("Failed to backup DB: ", e);
       throw new RuntimeException(e);
     }
   }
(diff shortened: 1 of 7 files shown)


kangnanli:
**Test:**  Other than the unit test, and previous pull request #469 . Here, we deploy the private build into cluster, and verify that: 3 sequential ingestion to the same resource from same s3_path. Only the 1st one will actually execute the ingestion, the 2nd and 3rd job will return succ immediately so that we can avoid double ingestion.

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -356,7 +351,7 @@ public static void restoreRemoteOrLocalDB(String host, int adminPort, String dbN
           new RestoreDBRequest(dbName, hdfsPath, upsreamHost, (short) upstreamPort);
       client.restoreDB(req);
     } catch (TException e) {
-      LOG.error("Failed to restore DB: ", e.toString());
```
gopalrajpurohit, line 359: this should be 

LOG.error("Failed to restore DB: ", e)

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -381,14 +375,13 @@ public static void backupDBToS3(String host, int adminPort, String dbName, Strin
       req.setInclude_meta(true);
       client.backupDBToS3(req);
     } catch (TException e) {
-      LOG.error("Failed to backup DB: ", e.toString());
+      LOG.error("Failed to backup DB: " + e.toString());
```
gopalrajpurohit, line 378: ditto

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -397,7 +390,7 @@ public static void backupDBToS3WithLimit(String host, int adminPort, String dbNa
       req.setLimit_mbs(limitMbs);
       client.backupDBToS3(req);
     } catch (TException e) {
-      LOG.error("Failed to backup DB: ", e.toString());
+      LOG.error("Failed to backup DB: " + e.toString());
```
gopalrajpurohit, line 393: ditto

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -429,7 +420,24 @@ public static void restoreRemoteOrLocalDBFromS3(String host, int adminPort, Stri
           new RestoreDBFromS3Request(dbName, s3Bucket, s3Path, upsreamHost, (short) upstreamPort);
       client.restoreDBFromS3(req);
     } catch (TException e) {
-      LOG.error("Failed to restore DB: ", e.toString());
+      LOG.error("Failed to restore DB: " + e.toString());
```
gopalrajpurohit, line 423: ditto

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -461,7 +469,7 @@ public static boolean isMasterReplica(String host, int adminPort, String dbName)
       CheckDBResponse res = client.checkDB(req);
       return res.is_master;
     } catch (TException e) {
-      LOG.error("Failed to check DB: ", e.toString());
+      LOG.error("Failed to check DB: " + e.toString());
```
gopalrajpurohit, line 472: ditto

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -482,26 +490,22 @@ public static boolean isLeaderReplica(String host, int adminPort, String dbName)
       // TODO (rajathprasad): add a new field "is_leader" and use that instead.
       return res.is_master;
     } catch (TException e) {
-      LOG.error("Failed to check DB: ", e.toString());
+      LOG.error("Failed to check DB: " + e.toString());
```
gopalrajpurohit, line 493: ditto

rajathprasad (approved):

PR #488, opened 2021-03-11
don't re-writeMeta when adddb if previous meta exist

kangnanli:
This fix #467 , which we always write an empty DBMetadata at aysnc_tm_addDB. 
This introduce bug, because when deploy, the helix StateTransition OnlineFromOffline or FollowerFromOffline will re-addDB, thus, we can't re-write the meta if previous meta already exist.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -622,11 +622,18 @@ void AdminHandler::async_tm_addDB(
     }
   }
 
-  if (!writeMetaData(request->db_name, "", "")) {
-    std::string errMsg = "AddDB failed to write initial DBMetaData for " + request->db_name;
-    SetException(errMsg, admin::AdminErrorCode::DB_ADMIN_ERROR, &callback);
-    LOG(ERROR) << errMsg;
-    return;
+  // update meta if not exist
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -622,11 +622,18 @@ void AdminHandler::async_tm_addDB(
...
 
-  if (!writeMetaData(request->db_name, "", "")) {
-    std::string errMsg = "AddDB failed to write initial DBMetaData for " + request->db_name;
-    SetException(errMsg, admin::AdminErrorCode::DB_ADMIN_ERROR, &callback);
-    LOG(ERROR) << errMsg;
-    return;
+  // update meta if not exist
+  auto meta = getMetaData(request->db_name);
+  if (!meta.__isset.s3_bucket && !meta.__isset.s3_path &&
+      !meta.__isset.last_kafka_msg_timestamp_ms) {
```
gopalrajpurohit, line 628: Can you specify what you are checking here for? Are you trying to identify OnlineOffline / Bootstrap statemodels with these fields?

Can you please provide more description on why you need these checks here... in code comments pls.

rajathprasad (approved):

PR #489, opened 2021-03-12
Revert all latest changes rely on checkdb and only keep thrift changes

kangnanli:
We are using the thrift change data in a peer-to-peer way, which introduce various compatibility issue. 
For example, in a cluster, during deploy, old hosts might still serving build without the thrift change, while new hosts are serving build with the thrift change, then, the old->new read, or new->old read on the thrift change will break. 

We will have to revert all business logic using the thrift change, but only keep the thrift change itself. 
Then, we will rollout thrift change to all clusters, then, add business logic to use the thrift change.


gopalrajpurohit:
Can you please build local application and verify that this change doesn't break anything else on any of the statemodels... and that the change is good to deploy on last known good build, before pushing this change out to rocksplicator codebase.

rajathprasad (approved):

kangnanli:
**Read-write** cluster:
if we define a build before 03/02 as goodBuild, and current private build with the revert as curtBuild
for each build:
I test ingest data to segment
test cluster restart
test host replacement
OK: all transitions / shard movement recover ok
I also try deploy from goodBuild -> curtBuild, and curtBuild -> goodBuild, and make sure transitions go through ok

Same tests apply to ReadOnly clusters

PR #491, opened 2021-03-13
Only change the local reference to last known published value, if publish is actually successful

gopalrajpurohit:
We match the resource map with last known published value... and don't publish if it is same as previous one. However, when we cache the last resource map, we don't know if later on the publish will be actually be successful. Hence we must store the reference to last known publish value, only if the last publish is actually successful.


kangnanli (approved):
this make sense

PR #492, opened 2021-03-13
Share a single zkConnection across multiple clusters to save connections to zookeeper cluster

gopalrajpurohit:
In case of some clients that need routing informations for multiple stateful clusters, it needs to watch multiple clusters to the shardmap zk. Earlier we used one zk client per cluster. However all cluster information are available on same zk cluster. Means we can potentially share the same connection across each of the cluster specific agents (as all agents for all clusters are running in the same jvm)

Verified the result by running a local clientshardagent and download multiple cluster's shardmap. It works in both cases. When explicitly disabled... it will fallback to old behaviour of creating one zk client per cluster it needs to watch.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/spectator/ConfigGeneratorClusterSpectatorImpl.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/spectator/ConfigGeneratorClusterSpectatorImpl.java
@@ -169,8 +169,13 @@ public class ConfigGeneratorClusterSpectatorImpl implements ClusterSpectator {
     if (this.clusterShardMapAgent == null) {
       if (shardMapZkSvr != null && !shardMapZkSvr.isEmpty()) {
         if (shardMapDownloadDir != null && !shardMapDownloadDir.isEmpty()) {
+          /**
+           * There is no need to share the zk client, as number of spectator instances is expected
+           * to be really small and total number of zk clients created will be linear to number of
+           * participant clusters being watched in total by all spectator instances...
+           */
           this.clusterShardMapAgent =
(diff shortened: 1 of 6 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/spectator/ConfigGeneratorClusterSpectatorImpl.java
```diff
@@ -169,8 +169,13 @@ public synchronized void start() throws Exception {
...
       if (shardMapZkSvr != null && !shardMapZkSvr.isEmpty()) {
         if (shardMapDownloadDir != null && !shardMapDownloadDir.isEmpty()) {
+          /**
+           * There is no need to share the zk client, as number of spectator instances is expected
+           * to be really small and total number of zk clients created will be linear to number of
+           * participant clusters being watched in total by all spectator instances...
+           */
           this.clusterShardMapAgent =
-              new ClusterShardMapAgent(shardMapZkSvr, clusterName, shardMapDownloadDir);
+              new ClusterShardMapAgent(shardMapZkSvr, null, clusterName, shardMapDownloadDir);
```
kangnanli, line 178: ClusterShardMapAgent is used as sidecar on Participant to receive latest shardMap generated by Spectator, right?
If so, why does Spectator needs an ClusterShardMapAgent as well?

gopalrajpurohit: Inorder to be able to perform leaf-handoff analysis. The delay from publishing to zk and then the receiving those updates on client side are measured on Spectator for the purpose of leader-handoff analysis. Although, currently the leader handoff is not integrated in distributed spectator.

kangnanli (approved):
lgtm

PR #495, opened 2021-03-16
Add simple backoff mechanism for file_watcher

gopalrajpurohit:
FileWatcher is used to watch changes in config files or other model file changes in rocksplicator applications.
In case of configurations like shardmap, they are latency sensitive. A blanket 2 second delay in all cases can cause significant drop in SR when shardmap needs to be updated as soon as it is available on the disk.

In order to adapt the file_watcher to different needs, we add a simple backoff mechanism which in case of configurations should be quick to register a file for re-registration without having to necessarily wait for 2 seconds.

### DIFF
--- a/common/file_watcher.cpp
+++ b/common/file_watcher.cpp
@@ -46,7 +46,11 @@
  */
 // DEFINE_int32(recheck_removed_file_interval_ms, 2 * 1000,
 //              "The interval for checking removed files");
-const int kRecheckRemovedFileIntervalMs = 2000;
+
+const int MAX_TIME_INTERVAL_SLOTS = 10;
+const int* TIME_INTERVALS = new int[MAX_TIME_INTERVAL_SLOTS]{
+  5, 10, 20, 40, 80, 160, 320, 640, 1280, 2560
+};
 
 namespace common {
 FileWatcher* FileWatcher::Instance() {
@@ -193,6 +197,8 @@ std::string FileWatcher::ReadFileAndHash(const std::string& file_name,
 
 int FileWatcher::RegisterFile(const std::string& file_name,
                               const bool check_dup) {
+  LOG(INFO) << "Registering file " << file_name << " to watch for changes";
+
   if (check_dup && states_.count(file_name) == 1) {
     LOG(ERROR) << file_name << " is already being watched, RemoveFile() "
                << "first if you want to change its callback.";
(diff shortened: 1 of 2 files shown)


ghanpipi:
Currently we only use AddFile() to register a config to watch. Will you plan to use ScheduleRegisterMonitoredFile for the new shardmap config publish solution?

gopalrajpurohit:
> Currently we only use AddFile() to register a config to watch. Will you plan to use ScheduleRegisterMonitoredFile for the new shardmap config publish solution?

No. It is internal detail. When we use AddFile for registration, it Registers the file to be watched immediately. However when the file being watched changes, it can change in different ways... one way is an atomic move from temp file to the file being watched.. or delete the old file and move the temp file to the file being watched... or re-write an existing file being watched...

When the file_watcher gets notification on the fd, it re-opens the file, during which it may have changed the fd. In which case it automatically re-schedules it for opening after 2 seconds. 2 seconds delay for config is bit too much. Hence I am simply introducing a backoff mechanism, so that config files can be loaded faster, likely within 5-10 ms... instead of waiting for 2 seconds.

Waiting for 2 seconds before a file is reloaded, for shardmap can have substantial impact at higher SR. Hence closing that gap.

example logs are

`
I0316 08:24:23.142529 22213 file_watcher.cpp:200] Registering file /data/xvdb/rocksplicator/zkshardmap/realpin-devtest to watch for changes
I0316 08:24:23.142568 22213 file_watcher.cpp:215] Registered file /data/xvdb/rocksplicator/zkshardmap/realpin-devtest to watch for changes
I0316 08:24:23.142668 22213 file_watcher.cpp:269] File change detected for /data/xvdb/rocksplicator/zkshardmap/realpin-devtest
E0316 08:24:23.537463 22213 file_watcher.cpp:166] File removed or unmounted: /data/xvdb/rocksplicator/zkshardmap/realpin-devtest
E0316 08:24:23.537489 22213 file_watcher.cpp:232] Schedule registering file /data/xvdb/rocksplicator/zkshardmap/realpin-devtest in 5 ms.
I0316 08:24:23.542639 22213 file_watcher.cpp:200] Registering file /data/xvdb/rocksplicator/zkshardmap/realpin-devtest to watch for changes
I0316 08:24:23.542665 22213 file_watcher.cpp:215] Registered file /data/xvdb/rocksplicator/zkshardmap/realpin-devtest to watch for changes
I0316 08:24:23.542765 22213 file_watcher.cpp:269] File change detected for /data/xvdb/rocksplicator/zkshardmap/realpin-devtest
`

ghanpipi:
Got it. mostly due to the fd_ will be changed after we re-download a new config file. Yes. 2 second is too heavy for the latency sensitive config files. Also previously there was flag to control this delay but shu change it to be hard coded.

ghanpipi (approved):

common/file_watcher.cpp
```diff
@@ -46,7 +46,11 @@
  */
 // DEFINE_int32(recheck_removed_file_interval_ms, 2 * 1000,
 //              "The interval for checking removed files");
-const int kRecheckRemovedFileIntervalMs = 2000;
+
+const int MAX_TIME_INTERVAL_SLOTS = 10;
+const int* TIME_INTERVALS = new int[MAX_TIME_INTERVAL_SLOTS]{
+  5, 10, 20, 40, 80, 160, 320, 640, 1280, 2560
+};
```
kangnanli, line 53: do you need deallocate these?

gopalrajpurohit: Not until the process goes down.. so not absolutely necessary.

PR #500, opened 2021-03-23
Add a segment tag for db sst size.

kangnanli:
Previously, for db (ie. a shard), we only have "db" tag, which makes it really hard to get the aggregated size of a whole segment. Plus, we have appended "version" to the segment in some of the use cases, which make the db_name as "segment""delimiter""version""shardNumber". 
Here, we added a util function `DbNameToSegmentAndVersion` to parse "segment" and "version" from a "db_name" and add them as tags to db_sst_size.

### DIFF
--- a/common/segment_utils.cpp
+++ b/common/segment_utils.cpp
@@ -35,23 +35,24 @@ std::string DbNameToSegment(const std::string& db_name) {
   return segment;
 }
 
-void DbNameToSegmentAndVersion(const std::string& db_name, std::string* seg,
-                               std::string* v, const std::string v_deli) {
+void DbNameToSegmentAndVersion(const std::string& db_name, std::string* segment,
+                               std::string* version,
+                               const std::string version_delim) {
   if (db_name.size() <= kShardLength) {
-    *seg = db_name;
+    *segment = db_name;
     return;
   }
-  if (!v_deli.empty()) {
-    auto iter = db_name.rfind(v_deli);
+  if (!version_delim.empty()) {
+    auto iter = db_name.rfind(version_delim);
     if (iter != std::string::npos) {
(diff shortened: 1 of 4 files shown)


ghanpipi:
@richardzhangrui is that ok for us to introduce these tags as the cardinality will increase a lot

kangnanli:
Add Tests:

**Test for Read-write system**, which has no <version>:
```
$ curl http://localhost:9999/stats.txt | grep total_sst_file_size
:  total_sst_file_size segment=test3 version= db=test300008: 3594
--  total_sst_file_size segment=test2 version= db=test200006: 0
```

**Test for Readonly system**, which has a timestamp <version>
```
$ curl http://localhost:9999/stats.txt | grep total_sst_file_size
   total_sst_file_size segment=ads version=1616533959 db=ads---161653395900145: 5436004
```

richardzhangrui:
how large can the cardinality be? there will be a host tag added by metric agent

kangnanli:
> how large can the cardinality be? there will be a host tag added by metric agent

Without current change, the sst size metrics cardinality is already num(shardNumber) * num(host), which could be ~10K * 1K already and hitting the cardinality (we have already observed this when query stats). 
Now, with current change, it add two more tags `segment`, `version`, which could increase the cardinality, but, our metrics query will most likely be "segment=*, version=, host=, db" to get the overall segment size. Thus, I think the "cardinality" might still be ok.

kangnanli:
> how large can the cardinality be? there will be a host tag added by metric agent

Do you have concern regarding `cardinality` of this sst size metrics? Considering we will mostly only query `segment=*` and leave get segment overall size. The query will actually have smaller unique values compared to previous query `db=*`

ghanpipi (approved):

PR #501, opened 2021-03-24
Revert revert 484

kangnanli:
This PR revert  the 484 revert in #489 

Basically, during Offline->Follower/S transition, it get Upstream's status from `checkDB` and compare with local status from local `checkDB`. Specifically, it upstreamStatus.db_metas is different localStatus, which means local DB need to rebuild from the upstream.


kangnanli:
Add some integration test details:
in #484 , we have already tested the logic works as expected (but, we did not verify the backward/forward compatibility during deploy). 
Thus, here, we test the backward/forward compatibility of build deployment. 
- on read-write system, we deploy 4 builds including a private build of current code change, and observe 1) live traffic sr remains >95% during all deployment; 2) on ERROR shards during deployment.

ghanpipi (approved):

rajathprasad (approved):

PR #509, opened 2021-04-08
Catch up to leader before completing offline -> follower transi…

rajathprasad:
…tion.

Current behavior in offline -> follower transition for a new offline
shard:
1. Find current leader.
2. Issue backup of DB.
3. Restore DB
4. Done.

Steps 2 and 3 can take up to an hour (and for some cases even more). So
the new follower could be lagging behind the leader for hours worth of
realtime updates.

Helix sometimes promotes a newly provisioned follower immediately to a
leader for load balancing. This causes the follower -> leader transition
to wait until the follower catches up to all the real time updates it
does not have. This causes drop in availability since there is no leader
for this shard during this period of time.

Proposed solution and behavior:
1. Find current leader.
2. Issue backup of DB.
3. Restore DB
4. Wait until the new follower has caught up to the leader (with a small
threshold)
5. Done.

With the proposed solution, future follower -> leader transitions can
complete quicker, improving availability.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -91,6 +91,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
 
   private static final Logger LOG = LoggerFactory.getLogger(LeaderFollowerStateModelFactory.class);
   private static final String LOCAL_HOST_IP = "127.0.0.1";
+  private static final int LEADER_CATCH_UP_THRESHOLD = 100;
 
   private final String host;
   private final int adminPort;
@@ -249,7 +250,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
                     ") from " + hostWithHighestSeq + " for " + dbName);
             localSeq = newLocalSeq;
             if (highestSeq <= localSeq) {
-              LOG.error(dbName + " catched up!");
+              LOG.error(dbName + " caught up!");
               break;
             }
             LOG.error("Slept for " + String.valueOf(i + 1) + " seconds for replicating " + dbName);
@@ -331,6 +332,37 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
       Utils.logTransitionCompletionMessage(message);
     }
 
+  public Map<String, String> getLiveHostAndRole(NotificationContext context, String dbName) {
+    HelixAdmin admin = context.getManager().getClusterManagmentTool();
+    ExternalView view = admin.getResourceExternalView(cluster, resourceName);
+    Map<String, String> stateMap = view.getStateMap(partitionName);
+
+    // find live replicas
+    Map<String, String> liveHostAndRole = new HashMap<>();
+    for (Map.Entry<String, String> instanceNameAndRole : stateMap.entrySet()) {
+      String role = instanceNameAndRole.getValue();
+      if (!role.equalsIgnoreCase("LEADER") &&
+          !role.equalsIgnoreCase("FOLLOWER") &&
+          !role.equalsIgnoreCase("OFFLINE")) {
+        continue;
+      }
+
+      String hostPort = instanceNameAndRole.getKey();
+      String host = hostPort.split("_")[0];
+      int port = Integer.parseInt(hostPort.split("_")[1]);
+
+      if (this.host.equals(host)) {
+        // myself
+        continue;
+      }
+
+      if (Utils.getLatestSequenceNumber(dbName, host, port) != -1) {
+        liveHostAndRole.put(hostPort, role);
+      }
+    }
+    return liveHostAndRole;
+  }
+
     /**
      * 3) Offline to Follower
      */
@@ -347,33 +379,8 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
           // open the DB if it's currently not opened yet
           Utils.addDB(dbName, adminPort, "FOLLOWER");
 
-          HelixAdmin admin = context.getManager().getClusterManagmentTool();
-          ExternalView view = admin.getResourceExternalView(cluster, resourceName);
-          Map<String, String> stateMap = view.getStateMap(partitionName);
-
           // find live replicas
-          Map<String, String> liveHostAndRole = new HashMap<>();
-          for (Map.Entry<String, String> instanceNameAndRole : stateMap.entrySet()) {
-            String role = instanceNameAndRole.getValue();
-            if (!role.equalsIgnoreCase("LEADER") &&
-                !role.equalsIgnoreCase("FOLLOWER") &&
-                !role.equalsIgnoreCase("OFFLINE")) {
-              continue;
-            }
-
-            String hostPort = instanceNameAndRole.getKey();
-            String host = hostPort.split("_")[0];
-            int port = Integer.parseInt(hostPort.split("_")[1]);
-
-            if (this.host.equals(host)) {
-              // myself
-              continue;
-            }
-
-            if (Utils.getLatestSequenceNumber(dbName, host, port) != -1) {
-              liveHostAndRole.put(hostPort, role);
-            }
-          }
+          Map<String, String> liveHostAndRole = getLiveHostAndRole(context, dbName);
 
           // Find upstream, prefer Leader
           String upstream = null;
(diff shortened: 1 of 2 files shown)


newpoo:
The high level idea sounds good to me.

Do we know how much time does "Wait until the new follower has caught up to the leader (with a small threshold)" take for a large db?

Another idea is to set FLAGS_replicator_max_updates_per_response to be a larger number, such as 1000? If it can reduce the catch up time to be low enough, it might be an easier fix.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -331,6 +332,37 @@ public void onBecomeFollowerFromLeader(Message message, NotificationContext cont
...
+      int port = Integer.parseInt(hostPort.split("_")[1]);
+
+      if (this.host.equals(host)) {
+        // myself
+        continue;
+      }
+
+      if (Utils.getLatestSequenceNumber(dbName, host, port) != -1) {
+        liveHostAndRole.put(hostPort, role);
+      }
```
kangnanli, line 361: do u recall why we check this seq num here?

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -471,6 +478,46 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
           Utils.restoreLocalDBFromS3(adminPort, dbName, s3Bucket, s3Path, snapshotHost,
               snapshotPort);
         }
+
+        try {
+          LOG.error(dbName + " Catching up live updates to leader");
+          Map<String, String> liveHostAndRole = getLiveHostAndRole(context, dbName);
```
kangnanli, line 484: are we expecting the stateMap changed since the last getExternalView? We are not here, right?
- if so, can we catch the externalView instead of get from zk again.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -471,6 +478,46 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
...
+          // Find upstream, prefer Leader
+          String upstream = null;
+          for (Map.Entry<String, String> instanceNameAndRole : liveHostAndRole.entrySet()) {
+            String role = instanceNameAndRole.getValue();
+            if (role.equalsIgnoreCase("LEADER")) {
+              upstream = instanceNameAndRole.getKey();
+              break;
+            } else {
+              upstream = instanceNameAndRole.getKey();
```
kangnanli, line 493: if somehow, you returned a `liveHostAndRole` in order: follower, follower, offline. then, won't this logic set the upstream to offline?

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -471,6 +478,46 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
...
+            leader_seq_num = Utils.getLatestSequenceNumber(dbName, upstreamHost, upstreamPort);
+
+            // If leader sequence number is within threshold, consider caught up.
+            if (leader_seq_num < local_seq_num + LEADER_CATCH_UP_THRESHOLD) {
+              LOG.error(dbName + " caught up!");
+              break;
+            }
+            TimeUnit.SECONDS.sleep(1);
+          }
+        } catch (Exception e) {
```
kangnanli, line 516: there is no explicit Exception throw from the catch up step, right? Then, this catch won't catch anything.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -471,6 +478,46 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
...
+
+          String upstreamHost = (upstream == null ? LOCAL_HOST_IP : upstream.split("_")[0]);
+          LOG.error(dbName + " got leader " + upstreamHost);
+          int upstreamPort =
+              (upstream == null ? adminPort : Integer.parseInt(upstream.split("_")[1]));
+
+          long local_seq_num;
+          long leader_seq_num;
+          // wait for up to 10 mins
+          for (int i = 0; i < 600; ++i) {
```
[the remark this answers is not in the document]
kangnanli: We could add a stats here to indicate whether such scenario occurs or not. Otherwise, since no exception throw, we won't know whether this 10mis catch up or not till we grep the log.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
rajathprasad: Not sure of the exact scenario, but it seems to be filter out hosts which are not accessible or dead. But those should be reflected in the external view too.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
rajathprasad: Since backup/restore can take up to an hour (and there no is mutex lock), high chances of external view changing. So we would want to get it again.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
rajathprasad: `getLiveHostAndRole` does not return offline replicas.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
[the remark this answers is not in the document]
rajathprasad: Yes, I can add stats here.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
rajathprasad: I thought there could be exceptions while getting latest seq number from leader. But that function is already catching Thrift exceptions. I'll leave it here to catch any other exceptions, we want this catch up to be a best effort.

rajathprasad:
> The high level idea sounds good to me.
> 
> Do we know how much time does "Wait until the new follower has caught up to the leader (with a small threshold)" take for a large db?
Catch up time depends on two factors:
1. Time it took for a shard to be backed up from leader and restored to host.
2. Update rate for that shard (how many updates per second).

In the current follower->leader transition, we have a time limit of 10 minutes for the catch up to finish (or else its set to ERROR state). And we have not observed hitting that limit (except on rare occasions) So for most cases catch up should be within 10 minutes.
> 
> Another idea is to set FLAGS_replicator_max_updates_per_response to be a larger number, such as 1000? If it can reduce the catch up time to be low enough, it might be an easier fix.
Yes, that will be easier. But conceptually, we would want the replica to be as close to the leader as possible before promoting it to follower (from offline). It's okay if offline->follower takes longer. But we want follower->leader take as little time as possible (since that affects availability)

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
kangnanli: it does? 
```
      if (!role.equalsIgnoreCase("LEADER") &&
          !role.equalsIgnoreCase("FOLLOWER") &&
          !role.equalsIgnoreCase("OFFLINE")) {
        continue;
      }
```

rajathprasad: Hmm, I read that wrong. 
This is the same code which we use to choose a leader to backup and restore from. 
Let me read it again and see why it's being done this way.

rajathprasad: I've added a check to skip any offline replicas.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -471,6 +478,46 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
           Utils.restoreLocalDBFromS3(adminPort, dbName, s3Bucket, s3Path, snapshotHost,
               snapshotPort);
         }
+
+        try {
```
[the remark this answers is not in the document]
kangnanli: I think this is OK. Since we are assuming same seqno among replicas are holding the same data, then, catch from any other is ok.

kangnanli (approved):
lgtm

PR #510, opened 2021-04-09
Add debug logs to Follower->Leader state model implementations.

rajathprasad:
We are seeing issues in production where the state transitions are
getting stuck, and it's difficult to reproduce in test setups. So,
temporarily add some extra logs for debugging. Will remove once issue is
resolved.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -187,6 +187,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
       leaderEventsCollector.addEvent(LeaderEventType.PARTICIPANT_LEADER_UP_INIT, null);
 
       Utils.logTransitionMessage(message);
+      final String dbName = Utils.getDbName(partitionName);
 
       try (Locker locker = new Locker(partitionMutex)) {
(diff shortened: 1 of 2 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -300,6 +304,7 @@ public void onBecomeLeaderFromFollower(Message message, NotificationContext cont
             e);
       } finally {
         leaderEventsCollector.commit();
```
kangnanli, line 306: one more before commit?

cluster_management/src/main/java/com/pinterest/rocksplicator/MasterSlaveStateModelFactory.java
```diff
@@ -297,6 +301,7 @@ public void onBecomeMasterFromSlave(Message message, NotificationContext context
             e);
       } finally {
         leaderEventsCollector.commit();
```
kangnanli, line 303: same here

kangnanli (approved):

cluster_management/src/main/java/com/pinterest/rocksplicator/MasterSlaveStateModelFactory.java
rajathprasad: There's a log statement just before, at the end of the try block, so that it was unnecessary here.

PR #511, opened 2021-04-14
Fix restoreTask issue - change leader role back to leader since by default db will be follower after restore

kangnanli:
During restore (ie. `restoreDB` api in admin_handler), the db will be added back to `db_manager` as `replicator::DBRole::SLAVE`. Then, after restore, the 3 replicas of a shard will have 1 leader 2 followers in its helix externalView, but, the interval `dbRole` at `db_manager` will all be `replicator::DBRole::SLAVE`. Thus, we need change the previous leader back to leader role.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -90,6 +90,10 @@ public class Utils {
         Integer.parseInt(partitionName.substring(lastIdx + 1)));
   }
 
+  /**
+   * @param dbName e.g. "p2p100001"
+   * @return e.g. "p2p1_1"
+   */
   public static String getPartitionName(String dbName) {
     String resourceName = dbName.substring(0, dbName.length() - 5);
     int partitionNumber = Integer.parseInt(dbName.substring(dbName.length() - 5));
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -164,7 +164,7 @@ public class Utils {
 
   public static void addRemoteOrLocalDB(String host, int adminPort, String dbName,
                                         String dbRole,
-                                        boolean overwrite_if_fail_add)
+                                        boolean overwriteIfFailAdd)
       throws RuntimeException {
     if (!dbRole.equals("SLAVE") && !dbRole.equals("FOLLOWER") && !dbRole.equals("NOOP")) {
       throw new RuntimeException("Invalid db role requested for new db " + dbName + " : " + dbRole);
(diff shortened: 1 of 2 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -143,60 +143,64 @@ public static void closeRemoteOrLocalDB(String host, int adminPort, String dbNam
...
-          LOG.error("Trying to overwrite open " + dbName);
-          req.setOverwrite(true);
-          client.addDB(req);
+        if (overwrite_if_fail_add) {
+          LOG.error("Failed to open " + dbName, e);
+          if (e.errorCode == AdminErrorCode.DB_ERROR) {
+            LOG.error("Trying to overwrite open " + dbName);
+            req.setOverwrite(true);
+            client.addDB(req);
+          }
```
rajathprasad, line 192: Is this change related to the restore fix which is described in this PR?

cluster_management/src/main/java/com/pinterest/rocksplicator/task/RestoreTask.java
```diff
@@ -129,13 +130,24 @@ protected void executeRestore(String host, int adminPort, String dbName, String
                                 boolean useS3Store, String s3Bucket)
       throws RuntimeException {
     try {
+      CheckDBResponse
+          checkDBResponse =
+          Utils.checkRemoteOrLocalDB(host, adminPort, dbName, false, null, null);
+      boolean isLeader = checkDBResponse.isIs_master();
```
rajathprasad, line 136: Is it possible that a leader can change (from a helix POV) while the restoring is taking place?

kangnanli: Simple Restore is basically overwriting any existing data on disk, thus, it does not matter leader change or not. 
the best practice to use it will be enable maintenance mode first before restore.

kangnanli: @rajathprasad any concern even if we have maintenance model on?

rajathprasad: But we cannot enforce that, right?

kangnanli:
Add some integration test info:
The service with the code change is deployed to test cluster, and restore from an existing backups. The restore succeed, and SR was only diminished during the restore, and recovered upon the restore complete.

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -90,6 +90,12 @@ public static String getDbName(String partitionName) {
         Integer.parseInt(partitionName.substring(lastIdx + 1)));
   }
 
+  public static String getPartitionName(String dbName) {
```
rajathprasad, line 93: Can you add an example like the function above?

rajathprasad (approved):

PR #515, opened 2021-05-04
add dbmeta files for checkpoint backup/restore

kangnanli:
There are two changes in this PR:

- Rocksplicator uses Helix to manage shard movement. When a new replica bootstrap, it involves Backup&Restore. By default, it relies on rocksdb::backupengine to manage the backup and restore, which also enables to backup/restore with customized meta. This is useful if during the new replica boostrap, some info (ie. meta) need to pass from Leader replica to the new replica. In addition, admin handler also build a customized "backup/restore" process bypassing the rocksdb::backupengine by adopting rocksdb::checkpoint directly; however, this customized "backup/restore" does not support backup/restore with meta. 
-- Thus, in this PR: at backup path, it encodes data (ie. meta) into a string and save to a local "dbmeta" file; then, it will be uploaded to s3 along with other files in the checkpoint. At the restore path, it will download all files from s3, and check whether local restored DB has a `dbmeta` file, if so, it will decode the meta data and remove the `dbmeta` file.

- S3 supports eventual consistency. To guarantee restore path lists all the objects backuped at backup path, a `_SUCCESS` file is appended as the last object during backup path; then, restore path will check the existence of `_SUCCESS` to ensure the backup's integrity.

### DIFF
--- a/common/file_util.cpp
+++ b/common/file_util.cpp
@@ -8,26 +8,6 @@ namespace filesystem = boost::filesystem;
 
 namespace common {
 
-void FileUtil::touch(const std::string& path) {
-  std::ofstream file(path);
-  file.close();
-}
-
-std::string FileUtil::getSuccessFilePath(const std::string& base_dir) {
-  if (!filesystem::exists(base_dir)) {
-    throw std::runtime_error(
-        "Failed to get success file path since base dir not exist, " +
-        base_dir);
-  }
-  return base_dir + ((base_dir.back() == '/') ? "_SUCCESS" : "/_SUCCESS");
-}
-
-std::string FileUtil::createSuccessFile(const std::string& base_dir) {
-  std::string path = FileUtil::getSuccessFilePath(base_dir);
-  touch(path);
-  return path;
-}
-
 std::string FileUtil::createFileWithContent(const std::string& base_dir,
                                             const std::string& filename,
                                             const std::string& content) {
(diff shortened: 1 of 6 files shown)


ghanpipi:
I remember aws supports  strong consistency for list operations now: https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-s3-now-delivers-strong-read-after-write-consistency-automatically-for-all-applications/#:~:text=Dec%201%2C%202020-,Amazon%20S3%20now%20delivers%20strong%20read%2Dafter%2Dwrite,consistency%20automatically%20for%20all%20applications.&text=S3%20also%20provides%20strong%20consistency,bucket%20with%20all%20changes%20reflected. 
so do we still need the _success marker?

kangnanli:
> I remember aws supports strong consistency for list operations now: https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-s3-now-delivers-strong-read-after-write-consistency-automatically-for-all-applications/#:~:text=Dec%201%2C%202020-,Amazon%20S3%20now%20delivers%20strong%20read%2Dafter%2Dwrite,consistency%20automatically%20for%20all%20applications.&text=S3%20also%20provides%20strong%20consistency,bucket%20with%20all%20changes%20reflected.
> so do we still need the _success marker?

Nope. That's cool. Then, we can just remove the _SUCCESS step.

ghanpipi:
lgtm

ghanpipi (approved):

PR #517, opened 2021-05-14
Add a metric for when the router config file watch failed

ghanpipi:


newpoo (commented):
I suspect this won't help. Because the process would crash immediately and it won't even have the time to send the stats to the monitoring system.

Is it possible to ensure no duplicated router objects are created with the same config file?

ghanpipi:
Yes, you are right. I forgot our metrics report is asynchronous. Will add the checker in service layer when trying to register a router file duplicately.

PR #524, opened 2021-06-23
Handle the case of all Replicas going OFFLINE

premkumr:
Basic Design
============
The leader stores its latest sequence number once in a while in ZK.
During transition to leader, if the remote  sequence number is higher
than the local seq number by more than the tolerance value, the
transition will fail and the partition will go into ERROR state.

When Helix allocates a new shard in this scenario, disallow the shard from becoming a leader. For this, we would need to know that the shard previously had data (seq. No > 0). This info needs to be stored in a central place that could be updated by the active instances and could be looked up by the new instance. 

When the old instance comes back online, Helix will promote the offline partition to FOLLOWER/LEADER.

We will discuss this solution in detail.

OVERVIEW
============

STORE LAST.STATE
-------------------
Partition state  = [Seq no, Timestamp, UpdateCount etc]
Seq No = Latest / Highest number as seen by the instance
Timestamp = lastmod of ts of the state
UpdateCount = no.of times, the state has been updated (Optional)
State is stored in ZK [last.state]

FLOW
------
On Slave -> Master
Fetch last known state from ZK [last.state]
Update data from host with highest seq no.
T = 0.95..1.0 (Tolerance)
Transition Condition (TC): current.seqno >= T * last.state.seqno
If TC
Then  allow transition
Else fail the transition

On Master
Every 10-15 min, update the last.state with local info.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -22,6 +22,9 @@ import com.pinterest.rocksdb_admin.thrift.CheckDBResponse;
 import com.pinterest.rocksplicator.eventstore.LeaderEventsCollector;
 import com.pinterest.rocksplicator.eventstore.LeaderEventsLogger;
 import com.pinterest.rocksplicator.thrift.eventhistory.LeaderEventType;
+import com.pinterest.rocksplicator.utils.PartitionState;
+import com.pinterest.rocksplicator.utils.PartitionStateManager;
+import com.pinterest.rocksplicator.utils.PartitionStateUpdater;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
@@ -104,6 +107,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
   private final boolean useS3Backup;
   private final String s3Bucket;
   private final LeaderEventsLogger leaderEventsLogger;
+  public final PartitionStateUpdater partitionStateUpdater;
 
   public LeaderFollowerStateModelFactory(
       final String host,
@@ -122,6 +126,8 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         new ExponentialBackoffRetry(1000, 3));
     zkClient.start();
     this.leaderEventsLogger = leaderEventsLogger;
+    this.partitionStateUpdater = new PartitionStateUpdater(this.zkClient, this.cluster, this.adminPort);
+    this.partitionStateUpdater.start();
   }
 
   @Override
@@ -136,9 +142,9 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         zkClient,
         useS3Backup,
         s3Bucket,
-        leaderEventsLogger);
+        leaderEventsLogger,
+        partitionStateUpdater);
   }
-
 
   public static class LeaderFollowerStateModel extends StateModel {
 
@@ -154,7 +160,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
     private final String s3Bucket;
     private InterProcessMutex partitionMutex;
     private final LeaderEventsLogger leaderEventsLogger;
-
+    public final PartitionStateUpdater partitionStateUpdater;
 
     /**
      * State model that handles the state machine of a single replica
@@ -168,7 +174,8 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         final CuratorFramework zkClient,
         final boolean useS3Backup,
         final String s3Bucket,
-        final LeaderEventsLogger leaderEventsLogger) {
+        final LeaderEventsLogger leaderEventsLogger,
+        PartitionStateUpdater partitionStateUpdater) {
       this.resourceName = resourceName;
       this.partitionName = partitionName;
       this.host = host;
@@ -180,6 +187,8 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
       this.partitionMutex = new InterProcessMutex(zkClient,
           getLockPath(cluster, resourceName, partitionName));
       this.leaderEventsLogger = leaderEventsLogger;
+      this.partitionStateUpdater = partitionStateUpdater;
+      LOG.error("prem: LeaderFollowerStateModel has been instantiated [" + partitionName + "]");
     }
 
     /**
@@ -193,6 +202,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
 
       Utils.logTransitionMessage(message);
       final String dbName = Utils.getDbName(partitionName);
+      PartitionStateManager stateManager = new PartitionStateManager(cluster,resourceName, partitionName, zkClient);
 
       try (Locker locker = new Locker(partitionMutex)) {
         HelixAdmin admin = context.getManager().getClusterManagmentTool();
@@ -220,6 +230,8 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         long localSeq = Utils.getLocalLatestSequenceNumber(dbName, adminPort);
         String hostWithHighestSeq = null;
         long highestSeq = localSeq;
+        long numActiveReplicas = 1;
+        long totalReplicas = stateMap.keySet().size();
         for (String instanceName : stateMap.keySet()) {
           String hostName = instanceName.split("_")[0];
           int port = Integer.parseInt(instanceName.split("_")[1]);
@@ -233,6 +245,9 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
             LOG.error("Failed to get latest sequence number from " + hostName + " for " + dbName);
             continue;
           }
+
+          numActiveReplicas++;
+
           if (highestSeq < seq) {
             highestSeq = seq;
             hostWithHighestSeq = hostName;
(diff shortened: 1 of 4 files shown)


rajathprasad (commented):
Can you make the same changes to MasterSlaveStateModelFactory.java. 
We want to keep the changes in sync.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -220,6 +230,8 @@ public void onBecomeLeaderFromFollower(Message message, NotificationContext cont
         long localSeq = Utils.getLocalLatestSequenceNumber(dbName, adminPort);
         String hostWithHighestSeq = null;
         long highestSeq = localSeq;
+        long numActiveReplicas = 1;
+        long totalReplicas = stateMap.keySet().size();
```
rajathprasad, line 234: I've sometimes seen helix not assign a replica to a shard. Example: {shard1: {host1:L, host2:F}. In that case, will totalReplicas be counted as 2? Is that okay?

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -266,10 +281,32 @@ public void onBecomeLeaderFromFollower(Message message, NotificationContext cont
           }
         }
 
+        LOG.error("prem: local.seq:" + localSeq + " replica:[" + numActiveReplicas + "/"+ totalReplicas + "]");
+        // 3-Node failure case
+        if (localSeq <= 0 || numActiveReplicas < totalReplicas/2) {
```
rajathprasad, line 286: Can you add a comment on why we have this condition?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+  public synchronized boolean removeLeader(String resource, String partition) {
+    LeaderInfo li = new LeaderInfo();
+    li.partition = partition;
+    li.resource = resource;
+    boolean success = leaders.remove(li) != null;
+    notify();
+    return success;
+  }
+
+  public void run() {
```
rajathprasad, line 100: I'm not familiar with the `Thread` class, is `run` automatically called when the class is constructed?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  public synchronized boolean addLeader(String resource, String partition) {
+    LeaderInfo li = new LeaderInfo();
+    li.partition = partition;
+    li.resource = resource;
+    if (leaders.containsKey(li)) {
+      return false;
+    }
+    leaders.put(li, new PartitionState());
```
rajathprasad, line 86: Should we add concurrency control around this since this function can be called simultaneously for multiple shards?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+          } catch(Exception e) {
+            LOG.error(String.format("Exception while storing state : %s",e));
+          }
+        }
+
+        // calculate wait time
+        runCount++;
+
+        if (runCount % 8 == 0) {
+          waitTime = minWaitMs;
```
rajathprasad, line 133: Reset runCount too?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionState.java
```diff
@@ -0,0 +1,82 @@
...
+    this.lastMod = System.currentTimeMillis();
+  }
+
+  public boolean isValid() {
+    return seqNum > -1 && lastMod > -1;
+  }
+
+  @Override
+  public String toString() {
+    return String.format("[seq.num=%d lastmod=%d]", seqNum, lastMod);
```
rajathprasad, line 55: Is %d acceptable for longs?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateManager.java
```diff
@@ -0,0 +1,131 @@
...
+              oldState.deserialize(stateData);
+              LOG.error("got.state: " + oldState);
+              return oldState;
+
+            } catch (KeeperException.NoNodeException ex) {
+              LOG.error(String.format("Path (%s) doesn't exist yet: ", statePath));
+              // By throwing the OperationTimeoutException, we force the retry, as NoNode exception is
+              // not retryable.
+              if (retryCount.get() < max_retries) {
+                throw new KeeperException.OperationTimeoutException();
```
rajathprasad, line 79: What happens the first time a segment is set up? Will this path not exist? In that case will this exception be thrown?

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
premkumr: For now it is fine. We can discuss corner cases on this later

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
premkumr: The idea was to make sure we check the remote state only when needed and not for all cases. LocalSeq <=0 is the primary case. I wanted to also check when the active replicas is less that 50% of the total replicas. But after looking at the numbers, this may not be worth it

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
premkumr: No. it was to be started in the ctor. It is currently done as `this.partitionStateUpdater.start();`

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
premkumr: It is already there. Java concurrency is simply used via the keyword `synchronized` in the function signature

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
premkumr: We could do that. But not needed as every 8 times, it will be in the same state. And we can use this later to see how many times this has been run

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionState.java
premkumr: Looks like its how it works in Java

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateManager.java
premkumr: Yes, the exception will be thrown. The request will be retried upto 3 times. After that an empty state will be returned

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
rajathprasad: I think it would be better to either remove this check, or get the total replicas using Helix API? This might cause a bug which will be hard to track later on.

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateManager.java
rajathprasad: I may be missing something here. If the exception is thrown, won't the state transition fail?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
rajathprasad: Did not know that, thank you!

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionState.java
kangnanli: maybe just use `seqnum` in the string format, to be consistent, instead of `seq.num`

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionState.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public PartitionState deserialize(String serializedData) {
+    if (serializedData == null) {
+      clear();
+    } else {
+      JSONParser parser = new JSONParser();
+      try {
+          JSONObject obj = (JSONObject) parser.parse(serializedData);
+          seqNum = ((Long)obj.get("seqnum")).intValue();
+          lastMod = ((Long)obj.get("lastmod")).intValue();
```
kangnanli, line 74: any reason to convert Long to Int? other than using `longValue()` to get `long`

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateManager.java
```diff
@@ -0,0 +1,131 @@
...
+      final CuratorFramework zkClient) {
+        this.cluster = cluster;
+        this.resourceName = resourceName;
+        this.partitionName = partitionName;
+        this.zkClient = zkClient;
+        oldState = new PartitionState();
+    }
+
+    public String getStatePath() {
+      return String.format("/pstate/%s/%s/%s",cluster,resourceName,partitionName);
```
kangnanli, line 53: maybe use `partitionstate` in the zk path to make it clear

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.HashMap;
+
+import org.apache.curator.framework.CuratorFramework;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PartitionStateUpdater extends Thread {
+  private static final Logger LOG = LoggerFactory.getLogger(PartitionStateUpdater.class);
+  private final AtomicBoolean running = new AtomicBoolean(false);
+  HashMap<LeaderInfo, PartitionState> leaders = new HashMap<LeaderInfo, PartitionState>();
```
kangnanli, line 32: rename to `leaderPStates`?

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateUpdater.java
kangnanli: should we reset `waitTime` for every 8 runs, or should we reset it for every leader add/removal? 
- if later, which might tell us, how often a leader info is changed

kangnanli (approved):

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionState.java
premkumr: Done.

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -180,6 +187,8 @@ public LeaderFollowerStateModel(
       this.partitionMutex = new InterProcessMutex(zkClient,
           getLockPath(cluster, resourceName, partitionName));
       this.leaderEventsLogger = leaderEventsLogger;
+      this.partitionStateUpdater = partitionStateUpdater;
+      LOG.error("prem: LeaderFollowerStateModel has been instantiated [" + partitionName + "]");
```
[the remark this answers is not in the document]
premkumr: Done

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionState.java
premkumr: Done

cluster_management/src/main/java/com/pinterest/rocksplicator/utils/PartitionStateManager.java
premkumr: Done

PR #525, opened 2021-06-23
Refactor replicator code to allow a pluggable DB to support CDC

iprentic:
Include one for existing rockstore functionality, and one that I will use to test the replication by printing the replication messages on the command line

### DIFF
--- a/rocksdb_replicator/rockdb_wrapper.cpp
+++ b/rocksdb_replicator/rockdb_wrapper.cpp
@@ -1,0 +1,32 @@
+#include "rocksdb_replicator/rocksdb_wrapper.h"
+
+namespace replicator {
+uint64_t RocksDbWrapper::LatestSequenceNumber() { return db_->GetLatestSequenceNumber(); }
+rocksdb::Status RocksDbWrapper::WriteToLeader(const rocksdb::WriteOptions& options,
+                                              rocksdb::WriteBatch* updates) {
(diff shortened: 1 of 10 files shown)


rocksdb_replicator/rockdb_wrapper.cpp
```diff
@@ -0,0 +1,32 @@
...
+uint64_t RocksDbWrapper::LatestSequenceNumber() { return db_->GetLatestSequenceNumber(); }
+rocksdb::Status RocksDbWrapper::WriteToLeader(const rocksdb::WriteOptions& options,
+                                              rocksdb::WriteBatch* updates) {
+  return db_->Write(options, updates);
+}
+rocksdb::Status RocksDbWrapper::GetUpdatesFromLeader(
+    rocksdb::SequenceNumber seq_number, std::unique_ptr<rocksdb::TransactionLogIterator>* iter) {
+  return db_->GetUpdatesSince(seq_number, iter);
+}
+bool RocksDbWrapper::HandleReplicateResponse(Update update) {
```
rajathprasad, line 13: will `update` be copied? Can we avoid that?

rajathprasad (approved):
lgtm

rocksdb_replicator/rockdb_wrapper.cpp
iprentic: I will change it to a pointer

PR #526, opened 2021-06-29
S3concurrent

randallk:
Added S3Concurrent class which uses aws sdk transfermanager to support concurrent, partitioned uploads.
Previous per-host upload speeds didn't exceed ~70MBps.  With changes, upwards of 500MBps are available.  
Large files are partitioned into smaller ranges for concurrent uploads.

### DIFF
--- a/common/rocksdb_env_s3.h
+++ b/common/rocksdb_env_s3.h
@@ -59,12 +59,19 @@ class S3FatalException : public std::exception {
 class S3Env : public Env {
 
  public:
-  explicit S3Env(const std::string& s3_key_prefix,
+  explicit S3Env(const std::string& s3_bucket,
+                 const std::string& s3_key_prefix,
                  const std::string& local_directory,
-                 std::shared_ptr<common::S3Util> s3_util) :
+                 std::shared_ptr<common::S3Util> s3_util,
+                 common::S3Concurrent* const s3_conc,
+                 const std::string& sync_group_id) :
+      s3_bucket_(s3_bucket),
       s3_key_prefix_(s3_key_prefix),
       local_directory_(local_directory),
-      s3_util_(std::move(s3_util)) {
+      s3_util_(std::move(s3_util)),
+      s3_conc_(s3_conc),
+      sync_group_id_(sync_group_id)
+  {
     posix_env_ = Env::Default();
   }
 
@@ -186,12 +193,18 @@ class S3Env : public Env {
   Env* GetBaseEnv() { return posix_env_; }
 
  private:
+  // s3 bucket
+  const std::string s3_bucket_;
   // s3 object key prefix
   const std::string s3_key_prefix_;
   // Local directory to temporarily store the files downloaded from/uploading to S3
   const std::string local_directory_;
+
+  const std::string sync_group_id_;
+
   // S3 util used for accessing AWS S3
   std::shared_ptr<common::S3Util> s3_util_;
+  common::S3Concurrent* s3_conc_{};
   // This object is derived from Env, but not from
   // posixEnv. We have posixEnv as an encapsulated
   // object here so that we can use posix timers,
--- a/common/rocksdb_env_s3.h
+++ b/common/rocksdb_env_s3.h
@@ -199,7 +199,7 @@ class S3Env : public Env {
   const std::string s3_key_prefix_;
   // Local directory to temporarily store the files downloaded from/uploading to S3
   const std::string local_directory_;
-
+  // string common to all files in a db backup operation
   const std::string sync_group_id_;
 
   // S3 util used for accessing AWS S3
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -585,7 +585,7 @@ bool S3Concurrent::setUploadMBps(uint64_t upload_MBps) {
 
 bool S3Concurrent::enqueuePutObject(const string& s3_bucket, const string& key,
                                     const string& local_path, const string& sync_group_id){
-  LOG(INFO) << "S3Concurrent::enqueuePutObject sync_group_id: "<< sync_group_id << " local_path: "  << local_path << " key: " << key;
+  LOG(INFO) << "S3Concurrent::enqueuePutObject" << " key: " << key;
   {
     std::lock_guard<std::mutex> lock(handles_mutex_);
     auto sync_collection_ptr = handles_.find(sync_group_id);
@@ -607,7 +607,7 @@ bool S3Concurrent::Sync(const string& sync_group_id) {
   uint64_t num_files_transferred = 0;
   uint64_t num_bytes_transferred = 0;
   uint64_t num_failures = 0;
-  LOG(INFO) << "S3Concurrent::Sync sync_group: " << sync_group_id;
+  LOG(INFO) << "S3Concurrent::Sync()";
   vector<std::shared_ptr<Aws::Transfer::TransferHandle>> sync_group;
   {
     std::lock_guard<std::mutex> lock(handles_mutex_);
@@ -622,12 +622,12 @@ bool S3Concurrent::Sync(const string& sync_group_id) {
   for(auto &h: sync_group){
     h->WaitUntilFinished();
     if (h->GetStatus() == Aws::Transfer::TransferStatus::COMPLETED) {
-      LOG(INFO) << "S3Concurrent::Sync sync_group: " << sync_group_id << " transfer success: " << h->GetTargetFilePath();
+      LOG(INFO) << "S3Concurrent::Sync success: " << h->GetTargetFilePath();
       Stats::get()->Incr(kS3PutObject);
       num_bytes_transferred += h->GetBytesTransferred();
       num_files_transferred ++;
     } else {
-      LOG(ERROR) << "S3Concurrent::Sync sync_group: " << sync_group_id << " transfer fail: " << h->GetTargetFilePath();
+      LOG(INFO) << "S3Concurrent::Sync fail "<< h->GetTargetFilePath();
       failures_.push_back(h);
       num_failures++;
       LOG(ERROR)
@@ -636,7 +636,7 @@ bool S3Concurrent::Sync(const string& sync_group_id) {
     }
   }
 
-  LOG(INFO) << "S3Concurrent sync_group: " << sync_group_id
+  LOG(INFO) << "S3Concurrent::Sync complete "
             << " transferred: " << num_files_transferred
             << " failures: " << num_failures;
   return num_files_transferred == sync_group.size();
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -625,7 +625,7 @@ bool S3Concurrent::Sync(const string& sync_group_id) {
       LOG(INFO) << "S3Concurrent::Sync success: " << h->GetTargetFilePath();
       Stats::get()->Incr(kS3PutObject);
       num_bytes_transferred += h->GetBytesTransferred();
-      num_files_transferred ++;
+      num_files_transferred++;
     } else {
       LOG(INFO) << "S3Concurrent::Sync fail "<< h->GetTargetFilePath();
       failures_.push_back(h);
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -508,35 +508,32 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
 }
 
 
-S3Concurrent::S3Concurrent(const string &bucket, const int upload_MBps)
-  : bucket_(bucket),
-    upload_MBps_(upload_MBps),
-    executor_(2),  // n_threads
+S3Concurrent::S3Concurrent(const int upload_MBps)
+  : executor_(2),  // n_threads
     transfer_config_(&executor_) {
 
-  const uint64_t MiB = 1024*1024;
-  const uint64_t max_heap_size = 400 * MiB;
-  const uint64_t buffer_size = 25 * MiB;
-  int max_connections = 16; // upper limit, good for 10Gbps.
-
+  const uint64_t MB = 1000*1000;
+  const uint64_t max_heap_size = 400 * MB;
+  const uint64_t buffer_size = 25 * MB;
+  int max_connections = 16; // reasonable upper limit, good for 10Gbps.
   // throttle concurrency according to desired rate.
-  // 2 threads allows for some file interleaving.
-  // approx 70 MB/sec speed per connection.
-  if (upload_MBps_ != 0) {
-    max_connections = std::min(max_connections, std::max(1, (int)(upload_MBps_/70)));
+  // 2 threads allows for some file interleaving while limiting cpu exposure.
+  if (upload_MBps != 0) {
+    const int MBps_per_connection_est = 70;
+    max_connections = std::min(max_connections, std::max(1, (int)(upload_MBps/MBps_per_connection_est)));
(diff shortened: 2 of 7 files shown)


common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+  client_config_.maxConnections = max_connections;
+  client_config_.region = "us-east-1";
+
+  if(s3_client_)
+    s3_client_.reset();
+  s3_client_ = std::make_shared<Aws::S3::S3Client>(client_config_);
+  transfer_config_.s3Client = s3_client_;
+  transfer_config_.transferBufferMaxHeapSize = max_heap_size;
+  transfer_config_.bufferSize = buffer_size;
+  transfer_config_.transferInitiatedCallback =
```
kangnanli, line 544: will this be called for each task/file for the transfer manager?

rocksdb_admin/admin_handler.cpp
```diff
@@ -483,6 +483,9 @@ AdminHandler::AdminHandler(
       LOG(INFO) << "Stopping DB deletioin thread ...";
     });
   }
+
+  int upload_MBps = 600; // set as fixed for now.
+  s3_conc_ = std::make_shared<common::S3Concurrent>("pinterest-jackson", upload_MBps);
```
kangnanli, line 488: we need to use the bucket passed by the gflag to support for pii cluster, which will use `pinterest-bristol`

rocksdb_admin/admin_handler.cpp
```diff
@@ -483,6 +483,9 @@ AdminHandler::AdminHandler(
       LOG(INFO) << "Stopping DB deletioin thread ...";
     });
   }
+
+  int upload_MBps = 600; // set as fixed for now.
```
kangnanli, line 487: yeah, there is a gflag for this

rocksdb_admin/admin_handler.cpp
```diff
@@ -986,70 +983,15 @@ void AdminHandler::async_tm_backupDBToS3(
...
-          common::Stats::get()->Incr(kS3BackupFailure);
-          return;
-        }
+    for (const auto& file : checkpoint_files) {
+      if (file == "." || file == "..") {
+        continue;
       }
+      s3_conc_->enqueuePutObject(formatted_s3_dir_path + file,
+                                 formatted_checkpoint_local_path + file,
+                                 //request->s3_bucket,
```
kangnanli, line 993: nite: remove this line

common/rocksdb_env_s3.h
```diff
@@ -186,12 +193,18 @@ class S3Env : public Env {
   Env* GetBaseEnv() { return posix_env_; }
 
  private:
+  // s3 bucket
+  const std::string s3_bucket_;
```
kangnanli, line 197: do we need this? Seem not used

common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+  const uint64_t max_heap_size = 400 * MiB;
+  const uint64_t buffer_size = 25 * MiB;
+  int max_connections = 16; // upper limit, good for 10Gbps.
+
+  // throttle concurrency according to desired rate.
+  // 2 threads allows for some file interleaving.
+  // approx 70 MB/sec speed per connection.
+  if (upload_MBps_ != 0) {
+    max_connections = std::min(max_connections, std::max(1, (int)(upload_MBps_/70)));
+  }
```
kangnanli, line 527: will throttle by `max_connections` more safe or throttle by actual network speed, regardless to avoid overloading host.

common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+            << " max_heap_size MiB: " << max_heap_size / MiB
+            << " buffer_size MiB: " << buffer_size / MiB;
+  int socket_timeout_ms = 3000;
+  client_config_.connectTimeoutMs = socket_timeout_ms;
+  client_config_.requestTimeoutMs = socket_timeout_ms;
+  client_config_.maxConnections = max_connections;
+  client_config_.region = "us-east-1";
+
+  if(s3_client_)
+    s3_client_.reset();
```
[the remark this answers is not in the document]
randallk: I'll take out the reset() calls, won't be called in constructor.

common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+        << "Failed to download file. s3 bucket: "  << th->GetBucketName()
+        << ", s3 key: " << th->GetKey()
+        << ", target_path: " << th->GetTargetFilePath()
+        << ", error: " << th->GetLastError().GetMessage();
+    };
+  if(context_)
+    context_.reset();
+  context_ = Aws::MakeShared<Aws::Client::AsyncCallerContext>("upload");
+  if(transfer_manager_)
+    transfer_manager_.reset();
```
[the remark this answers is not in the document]
randallk: I'll take out the reset() calls, not called in constructor.

common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+
+S3Concurrent::S3Concurrent(const string &bucket, const int upload_MBps)
+  : bucket_(bucket),
+    upload_MBps_(upload_MBps),
+    executor_(2),  // n_threads
+    transfer_config_(&executor_) {
+
+  const uint64_t MiB = 1024*1024;
+  const uint64_t max_heap_size = 400 * MiB;
+  const uint64_t buffer_size = 25 * MiB;
```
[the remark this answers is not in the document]
randallk: AWS config uses MiB for buffer sizes (despite calling them 'MB'), eg.
https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#multipart-threshold
so I'm following the convention there as it's setting buffer sizes within sdk objects.
The sizes are experimentally derived.  These were the settings able to achieve highest bandwidth.  Less memory, and fewer or more threads showed lower bandwidth.
There's possibility for more refinement, but the extremely high variance of the bandwidth numbers (due to network or service congestion likely) .  These were able to achieve 850MB/sec+ in good runs, and allow limiting by only varying max_connections.

common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+  // 2 threads allows for some file interleaving.
+  // approx 70 MB/sec speed per connection.
+  if (upload_MBps_ != 0) {
+    max_connections = std::min(max_connections, std::max(1, (int)(upload_MBps_/70)));
+  }
+  LOG(INFO) << "S3Concurrent() upload_MBps: " << upload_MBps_
+            << " max_connections: " << max_connections
+            << " max_heap_size MiB: " << max_heap_size / MiB
+            << " buffer_size MiB: " << buffer_size / MiB;
+  int socket_timeout_ms = 3000;
```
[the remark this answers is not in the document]
randallk: It's the same value as s3util is using for socket timeout.

common/s3util.cpp
```diff
@@ -493,4 +507,124 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+  client_config_.connectTimeoutMs = socket_timeout_ms;
+  client_config_.requestTimeoutMs = socket_timeout_ms;
+  client_config_.maxConnections = max_connections;
+  client_config_.region = "us-east-1";
+
+  if(s3_client_)
+    s3_client_.reset();
+  s3_client_ = std::make_shared<Aws::S3::S3Client>(client_config_);
```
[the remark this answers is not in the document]
randallk: I'm not using the s3client directly, only allocating and configuring it for the TransferManager.  Future changes to CustomizedClient could break assumptions of TransferManager class, so better to stick with stock sdk class.

common/s3util.cpp
randallk: No, one S3Concurrent object (and therefore TransferManager) is created for the AdminHandler process, and is reused.

common/s3util.cpp
randallk: Docs recommend using max_connections to manage bandwidth usage:
https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#max-bandwidth
Using AWS DefaultRateLimiter as well as targeting the concurrency would allow for a bit of safety while avoiding the problems with too many blocked threads.
I'll add it in and test.

common/s3util.cpp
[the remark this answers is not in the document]
randallk: Is there a reason to use the customized class here?

common/s3util.cpp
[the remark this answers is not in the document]
randallk: Did some testing with MB (1e6) multiples for buffer sizes, they work fine.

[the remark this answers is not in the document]
kangnanli: Can u share what buffer sizes tested and a brief comparison, just for further reference

rocksdb_admin/admin_handler.cpp
randallk: I just see s3_download_limit_mb, nothing for upload.

rocksdb_admin/admin_handler.cpp
randallk: I don't see the gflag for bucket ... s3_bucket, s3_bucket_backup just used in tests.

common/s3util.cpp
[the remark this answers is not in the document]
kangnanli: The `CustomizedS3Client` does not change any default functions. It add 1 more API `MakeHttpRequest`. 
Actually, it is ok to do your current way. 
Those place need `MakeHttpRequest` will still call with `CustomziedS3Client` from `std::unique_ptr<CutomizedS3Client> s3Client;`; and current change only replies those. APIs `PutObject`s

common/s3util.cpp
```diff
@@ -493,4 +507,128 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+  const uint64_t max_heap_size = 400 * MB;
+  const uint64_t buffer_size = 25 * MB;
+  int max_connections = 16; // reasonable upper limit, good for 10Gbps.
+  // throttle concurrency according to desired rate.
+  // 2 threads allows for some file interleaving while limiting cpu exposure.
+  if (upload_MBps != 0) {
+    const int MBps_per_connection_est = 70;
+    max_connections = std::min(max_connections, std::max(1, (int)(upload_MBps/MBps_per_connection_est)));
+    rate_limiter_ = Aws::MakeShared<Aws::Utils::RateLimits::DefaultRateLimiter<>>("S3Concurrent", upload_MBps * MB);
+    client_config_.writeRateLimiter = rate_limiter_;
```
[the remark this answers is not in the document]
kangnanli: I just realized, this PR only handles the `Backup`. Then, let's add similar change to `Restore` ie. write path in another PR then.

rocksdb_admin/admin_handler.cpp
```diff
@@ -483,6 +483,9 @@ AdminHandler::AdminHandler(
       LOG(INFO) << "Stopping DB deletioin thread ...";
     });
   }
+
+  const int upload_MBps = 600; // fixed upper bw limit for now.
```
kangnanli, line 487: As @rajathprasad suggest, besides `s3_download_limit_mb` gflag in `admin_handler.cpp`, we could add another one for backup instead of hardcoded 600.

rocksdb_admin/admin_handler.cpp
kangnanli: There is a gflag for which s3 bucket to use: https://github.com/pinterest/rocksplicator/blob/master/common/helix_client.cpp#L35. But, use the bucket specified in the request should be good.

rocksdb_admin/admin_handler.cpp
kangnanli: yeah, currently, there is limit setup for the write path from EC2 to S3: https://github.com/pinterest/rocksplicator/blob/master/rocksdb_admin/admin_handler.cpp#L1572. 
Should we add 1 more flag?

randallk: Done, added s3_upload_limit_mb

rocksdb_admin/admin_handler.cpp
randallk: Done, added gflags setting with 600 as default.

rocksdb_admin/admin_handler.cpp
randallk: OK will just use bucket from the request.

common/s3util.cpp
randallk: Ratelimiter added.

common/s3util.cpp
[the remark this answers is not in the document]
randallk: Sounds good.

common/s3util.cpp
```diff
@@ -488,9 +489,160 @@ shared_ptr<S3Util> S3Util::BuildS3Util(
...
+    h->WaitUntilFinished();
+    if (h->GetStatus() == Aws::Transfer::TransferStatus::COMPLETED) {
+      LOG(INFO) << "S3Concurrent::Sync success: " << h->GetTargetFilePath();
+      Stats::get()->Incr(kS3PutObject);
+      num_bytes_transferred += h->GetBytesTransferred();
+      num_files_transferred ++;
+    } else {
+      LOG(INFO) << "S3Concurrent::Sync fail "<< h->GetTargetFilePath();
+      failures_.push_back(h);
+      num_failures++;
```
kangnanli, line 632: Can u lint the code changes from your editor? 
e.g. the `num_files_transferered ++`, but, `num_failures++`, the former has an extra space before `++`

rocksdb_admin/admin_handler.cpp
```diff
@@ -911,38 +914,32 @@ void AdminHandler::async_tm_backupDBToS3(
...
-        folly::stringPrintf("Concurrent uploading/downloading limit hits %d by %s",
-                            n, request->db_name.c_str());
-    SetException(err_str, AdminErrorCode::DB_ADMIN_ERROR, &callback);
-    LOG(ERROR) << err_str;
-    common::Stats::get()->Incr(kS3BackupFailure);
-    return;
-  }
-
   common::Timer timer(kS3BackupMs);
-  LOG(INFO) << "S3 Backup " << request->db_name << " to " << request->s3_backup_dir;
```
kangnanli, line 931: what is reason to remove this log? this is a log message to inform, it is about to backup the whole db, seem ok to remain

rocksdb_admin/admin_handler.cpp
```diff
@@ -1095,9 +1035,13 @@ void AdminHandler::async_tm_backupDBToS3(
...
   }
-
+  if (!s3_conc_->Sync(sync_group_id)) {
+    callback.release()->exceptionInThread(std::move(e));
+    common::Stats::get()->Incr(kS3BackupFailure);
+    return;
+  }
   LOG(INFO) << "S3 Backup is done for " << request->db_name
-            << " with latency(ms) " << timer.getElapsedTimeMs();
+            << " with latency(ms)" << timer.getElapsedTimeMs();
```
kangnanli, line 1044: lint: let's still keep the space after `(ms)`

kangnanli (approved):

PR #530, opened 2021-07-26
Add a function to ping the JVM.

rajathprasad:
We are seeing instances where helix participant is not processing any
new state transitions. Adding a 'ping' function for debugging purposes
to check if the JVM is alive.
Piggybacking this JVM call with the checkDB API (which is also used for
debugging).

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Participant.java
@@ -334,6 +334,10 @@ public class Participant {
     }
   }
 
+  public static void ping() {
+    LOG.error("[participant] Received ping");
+  }
+
   private Participant(String zkConnectString, String clusterName, String instanceName,
                       String stateModelType, int port, String postUrl, boolean useS3Backup,
                       String s3BucketName, boolean runSpectator)
--- a/common/helix_client.cpp
+++ b/common/helix_client.cpp
@@ -306,4 +306,35 @@ std::string GetLeaderInstanceId(
     return JStringToString(env, jstr);
 }
 
+void PingJVM() {
+    auto env = getRunningVM();
+
+    if (env == nullptr) {
+        LOG(ERROR) << "Failed to get env";
+        return;
+    }
+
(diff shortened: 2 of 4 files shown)


kangnanli:
The ping funct looks good to me. Why not include this logic into the adminHandler::ping() since it make more sense.

rajathprasad:
> The ping funct looks good to me. Why not include this logic into the adminHandler::ping() since it make more sense.

I didn't know that the function exists, will change it to that. Thanks.

rajathprasad:
@kangnanli the ping API in adminHandler has no return type (void), so I cannot return a failure for JVM issues. So I'd prefer keeping it this way for now. Ideally we would want to change the ping API or add a new one with a return type.

jaricftw:
> @kangnanli the ping API in adminHandler has no return type (void), so I cannot return a failure for JVM issues. So I'd prefer keeping it this way for now. Ideally we would want to change the ping API or add a new one with a return type.

Isn't the same for `checkDB`?
```
void AdminHandler::async_tm_checkDB(
    std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
      CheckDBResponse>>> callback,
    std::unique_ptr<CheckDBRequest> request) {
```

I am also +1 on Kangnan's earlier question on why not adding the PingJVM logic in the `AdminHandler::ping` handler

rajathprasad:
> > @kangnanli the ping API in adminHandler has no return type (void), so I cannot return a failure for JVM issues. So I'd prefer keeping it this way for now. Ideally we would want to change the ping API or add a new one with a return type.
> 
> Isn't the same for `checkDB`?
> 
> ```
> void AdminHandler::async_tm_checkDB(
>     std::unique_ptr<apache::thrift::HandlerCallback<std::unique_ptr<
>       CheckDBResponse>>> callback,
>     std::unique_ptr<CheckDBRequest> request) {
> ```
> 
> I am also +1 on Kangnan's earlier question on why not adding the PingJVM logic in the `AdminHandler::ping` handler

`CheckDB` has a return type `CheckDBResponse`. But `ping` does not.

PR #537, opened 2021-08-10
Pass Config options per dataset for Writes

premkumr:
- Allow service config options to be passed from the service


rocksdb_admin/tests/utils_test.cpp
```diff
@@ -1,54 +0,0 @@
-#include "rocksdb_admin/utils.h"
-
-#include <boost/filesystem.hpp>
-#include <string>
-#include "common/file_util.h"
-#include "gtest/gtest.h"
-#include "rocksdb_admin/gen-cpp2/rocksdb_admin_types.h"
-
-namespace filesystem = boost::filesystem;
```
rajathprasad, line 9: Any reason for deleting this test?

rajathprasad (approved):
LGTM!

rocksdb_admin/tests/utils_test.cpp
premkumr: Actually no sure .. Looks like a mistake . Let me add this back in .

PR #538, opened 2021-08-12
Initialize atomic variables to 0.

rajathprasad:
According to the docs at https://en.cppreference.com/w/cpp/atomic/atomic/atomic, variables are not initialized. So initializing them in the constructor here.


neildge:
lg2m

PR #540, opened 2021-08-16
Set admin_handler.rocksdb_options to protected.

randallk:
rocksdb_options has filters that are useful to make available to admin_handler subclasses.
In particular the ttl_compaction_filter has segment ttl data that can be used by rockstorewidecolumn_rw_handler.


randallk (commented):
For upcoming changes to rockstorewidecolumn to give access to ttl_compaction_filter object.

PR #544, opened 2021-08-28
Add option to read upstream cluster zk from add db request

iprentic:
instead of from command-line args

This will allow us to observe multiple rocksplicator clusters from one rocksobserver cluster

### DIFF
--- a/cdc_admin/cdc_admin.thrift
+++ b/cdc_admin/cdc_admin.thrift
@@ -34,7 +34,7 @@ struct AddObserverRequest {
   # if this db is already being observed, an ALREADY_EXIST error is thrown
   1: required string db_name,
   2: required string upstream_ip,
-  # The following are needed if the replicator zk and helix cluster will not be provided via command-line args
+  # provide the upstream replicator's zk and helix cluster through request instead of command-line args
   3: optional string replicator_zk_cluster,
   4: optional string replicator_helix_cluster,
 }
--- a/cdc_admin/cdc_admin.thrift
+++ b/cdc_admin/cdc_admin.thrift
@@ -34,6 +34,9 @@ struct AddObserverRequest {
   # if this db is already being observed, an ALREADY_EXIST error is thrown
   1: required string db_name,
   2: required string upstream_ip,
+  # The following are needed if the replicator zk and helix cluster will not be provided via command-line args
+  3: optional string replicator_zk_cluster,
+  4: optional string replicator_helix_cluster,
 }
 
 struct AddObserverResponse {
--- a/cdc_admin/cdc_admin_handler.cpp
+++ b/cdc_admin/cdc_admin_handler.cpp
@@ -116,13 +116,15 @@ void CDCAdminHandler::async_tm_addObserver(
   // add the db to db_manager
   std::string err_msg;
   replicator::DBRole role = replicator::DBRole::SLAVE;
+  const std::string replicator_zk_cluster = request->__isset.replicator_zk_cluster ? request->replicator_zk_cluster : std::string("");
+  const std::string replicator_helix_cluster = request->__isset.replicator_helix_cluster ? request->replicator_helix_cluster : std::string("");
 
   // TODO(indy): Read sequence # from kafka
   // And create a db wrapper based on the internal producer
   std::unique_ptr<replicator::TestDBProxy> db_wrapper =
       std::make_unique<replicator::TestDBProxy>(request->db_name, 0);
   if (!db_manager_->addDB(
-          request->db_name, std::move(db_wrapper), role, std::move(upstream_addr), &err_msg)) {
+          request->db_name, std::move(db_wrapper), role, std::move(upstream_addr), replicator_zk_cluster, replicator_helix_cluster, &err_msg)) {
     e.errorCode = CDCAdminErrorCode::ADMIN_ERROR;
     e.message = std::move(err_msg);
     callback.release()->exceptionInThread(std::move(e));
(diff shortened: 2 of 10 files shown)


cdc_admin/cdc_admin_handler.cpp
```diff
@@ -116,13 +116,15 @@ void CDCAdminHandler::async_tm_addObserver(
   // add the db to db_manager
   std::string err_msg;
   replicator::DBRole role = replicator::DBRole::SLAVE;
+  const std::string replicator_zk_cluster = request->__isset.replicator_zk_cluster ? request->replicator_zk_cluster : std::string("");
+  const std::string replicator_helix_cluster = request->__isset.replicator_helix_cluster ? request->replicator_helix_cluster : std::string("");
```
kangnanli, line 120: nit: line seem really long, lint it a little bit?

cdc_admin/cdc_admin.thrift
```diff
@@ -34,6 +34,9 @@ struct AddObserverRequest {
   # if this db is already being observed, an ALREADY_EXIST error is thrown
   1: required string db_name,
   2: required string upstream_ip,
+  # The following are needed if the replicator zk and helix cluster will not be provided via command-line args
```
kangnanli, line 37: ```# provide the upstream replicator's zk and helix cluster through request instead of command-line args```

rocksdb_replicator/rocksdb_replicator.cpp
```diff
@@ -101,22 +101,19 @@ ReturnCode RocksDBReplicator::addDB(const std::string& db_name,
   std::shared_ptr<DbWrapper> db_wrapper(
     new RocksDbWrapper(db_name, std::move(db))
   );
-  return addDB(db_name, db_wrapper, role, upstream_addr, replicated_db);
+  return addDB(db_name, db_wrapper, role, upstream_addr, replicated_db = replicated_db);
```
kangnanli, line 104: Java take keyword arg? `replicated_db = replicated_db`?

rocksdb_replicator/rocksdb_replicator.cpp
```diff
@@ -101,22 +101,19 @@ ReturnCode RocksDBReplicator::addDB(const std::string& db_name,
...
                                     const folly::SocketAddress& upstream_addr,
-                                    ReplicatedDB** replicated_db) {
-  // Change this
-  // Oh you already did :) 
-  // Well sort of
-  // Make it so you pass in the wrapper
-  // OR can pass in the db and then you make it # overloading # c++ # expert
+                                    ReplicatedDB** replicated_db,
+                   const std::string& replicator_zk_cluster,
+                   const std::string& replicator_helix_cluster) {
```
kangnanli, line 113: lint

rocksdb_replicator/rocksdb_replicator.cpp
iprentic: Don't worry this is in cpp :)

cdc_admin/cdc_admin.thrift
```diff
@@ -34,6 +34,9 @@ struct AddObserverRequest {
   # if this db is already being observed, an ALREADY_EXIST error is thrown
   1: required string db_name,
   2: required string upstream_ip,
+  # provide the upstream replicator's zk and helix cluster through request instead of command-line args
+  3: optional string replicator_zk_cluster,
+  4: optional string replicator_helix_cluster,
```
rajathprasad, line 39: Should we make these required?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -134,13 +134,17 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
     folly::Executor* executor,
     const DBRole role,
     const folly::SocketAddress& upstream_addr,
-    common::ThriftClientPool<ReplicatorAsyncClient>* client_pool)
+    common::ThriftClientPool<ReplicatorAsyncClient>* client_pool,
+    const std::string& replicator_zk_cluster,
+    const std::string& replicator_helix_cluster)
```
rajathprasad, line 139: Should we create another constructor with these two new arguments? That would prevent any changes to existing callers. WDYT?

cdc_admin/cdc_admin.thrift
iprentic: I think they should be optional so that we still support having them through the cluster command-line args. I think we should be conservative about adding required fields.

rocksdb_replicator/replicated_db.cpp
iprentic: They have default values:
```
                   const std::string& replicator_zk_cluster = "",
                   const std::string& replicator_helix_cluster = "");
```
(see https://github.com/pinterest/rocksplicator/pull/544/files/9579632255eea6b3b67e4edd8eddb781780d0cc8#diff-cb14f07e0183bff583746dbb5adf040de197c3cd1dfcb08369770004f083955bR187-R188) 
so existing callers should not be affected

rocksdb_replicator/rocksdb_replicator.cpp
kangnanli: o. sorry, I mean cpp. 
So, this is not keyboard arg as in python. You are assign value to itself, which is confusing.

iprentic: 🤦‍♀️  too much switching languages recently... fixed

kangnanli (approved):

rajathprasad (approved):

PR #547, opened 2021-09-08
Put replicator sequence number in Update sent from leader to follower

iprentic:
in replication

Can be used by followers that are not rocksdb instances (e.g. rocksobserver)

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -363,7 +363,6 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
           logMetric(kReplicatorGetUpdatesSinceMs, start < end ? end - start : 0,
                     db->db_name_);
         }
-
         if (use_cached_iter || status.ok() || status.IsNotFound()) {
           ReplicateResponse response;
           uint64_t read_bytes = 0;
@@ -372,6 +371,7 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
                ++i, iter->Next()) {
             auto result = iter->GetBatch();
             Update update;
+            update.set_seq_no(result.sequence);
             next_seq_no += result.writeBatchPtr->Count();
             const auto& str = result.writeBatchPtr->Data();
             read_bytes += str.size();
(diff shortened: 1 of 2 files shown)


rajathprasad (approved):

kangnanli:
Can u elaborate wha u mean by "Can be used by followers that are not rocksdb instances (e.g. rocksobserver)"?
- how the rocksobserver use it?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -372,6 +371,7 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
                ++i, iter->Next()) {
             auto result = iter->GetBatch();
             Update update;
+            update.set_seq_no(result.sequence);
```
kangnanli, line 374: do you need add seqno for all replication request, or, you will only need when leader->rocksobserver

iprentic:
> Can u elaborate wha u mean by "Can be used by followers that are not rocksdb instances (e.g. rocksobserver)"?
> 
> * how the rocksobserver use it?

We use it in RocksObserver if there is no sequence number in Kafka, to know where we have started replicating from. Since the sequence number of the first write batch in the writeahead log may not be 0 (and will not be if there is retention set up)

kangnanli (commented):
> > Can u elaborate wha u mean by "Can be used by followers that are not rocksdb instances (e.g. rocksobserver)"?
> > 
> > * how the rocksobserver use it?
> 
> We use it in RocksObserver if there is no sequence number in Kafka, to know where we have started replicating from. Since the sequence number of the first write batch in the writeahead log may not be 0 (and will not be if there is retention set up)

So, if a LeaderFollower resource is created at the same time with rocksdobserver, then, rocksobserver should always begin replicate from 0. 
if the LeaderFollower resource is created first, then, later, the rocksobserver is created, then, rocksobserver will begin replicate from the smalltest seqno in the leader's WAL? (which is this PR).

rocksdb_replicator/replicated_db.cpp
iprentic: We only need it for leader->rocksobserver, but on the leader we do not distinguish between different types of follower... we could add a field in the request type to indicate that we need the sequence number, but to keep it simple, I propose always returning it. What do you think?

PR #548, opened 2021-09-08
in CDC admin, accept db wrapper implementation passed in

iprentic:
By default, print the updates to the log files, but can be overridden to use any db_wrapper

### DIFF
--- a/cdc_admin/cdc_admin_handler.h
+++ b/cdc_admin/cdc_admin_handler.h
@@ -75,9 +75,7 @@ protected:
   common::ObjectLock<std::string> db_admin_lock_;
 
   virtual std::unique_ptr<replicator::DbWrapper> getDbWrapper(std::string db_name) {
-      std::unique_ptr<replicator::TestDBProxy> db_wrapper =
-          std::make_unique<replicator::TestDBProxy>(db_name, 0);
-      return db_wrapper;
+      return std::make_unique<replicator::TestDBProxy>(db_name, 0);
   }
 
 private:
--- a/cdc_admin/cdc_admin_handler.h
+++ b/cdc_admin/cdc_admin_handler.h
@@ -27,6 +27,7 @@
 #else
 #include "cdc_admin/gen-cpp2/CdcAdmin.h"
 #endif
+#include "rocksdb_replicator/test_db_proxy.h"
 
 namespace cdc_admin {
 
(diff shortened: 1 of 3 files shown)


cdc_admin/cdc_admin_handler.h
```diff
@@ -73,6 +74,12 @@ class CDCAdminHandler : virtual public CdcAdminSvIf {
   // of overriding some admin functions
   common::ObjectLock<std::string> db_admin_lock_;
 
+  virtual std::unique_ptr<replicator::DbWrapper> getDbWrapper(std::string db_name) {
+      std::unique_ptr<replicator::TestDBProxy> db_wrapper =
+          std::make_unique<replicator::TestDBProxy>(db_name, 0);
+      return db_wrapper;
+  }
+
```
rajathprasad, line 82: Instead of having a test implementation as default should we make this a pure virtual method (and make this an abstract class)?

iprentic: For the sake of the open source folks out there, I like having it work out of the box :) but I can change it if you disagree!

cdc_admin/cdc_admin_handler.h
```diff
@@ -73,6 +74,12 @@ class CDCAdminHandler : virtual public CdcAdminSvIf {
   // of overriding some admin functions
   common::ObjectLock<std::string> db_admin_lock_;
 
+  virtual std::unique_ptr<replicator::DbWrapper> getDbWrapper(std::string db_name) {
+      std::unique_ptr<replicator::TestDBProxy> db_wrapper =
+          std::make_unique<replicator::TestDBProxy>(db_name, 0);
+      return db_wrapper;
```
kangnanli, line 80: just `return std::make_unique<replicator::TestDBProxy>(db_name, 0);`?

kangnanli (commented):
I did not fully understand the purpose of this PR, wasn't it a code refactor?
can u illustrate what this means `By default, print the updates to the log files, but can be overridden to use any db_wrapper`?

iprentic:
> I did not fully understand the purpose of this PR, wasn't it a code refactor?
> can u illustrate what this means `By default, print the updates to the log files, but can be overridden to use any db_wrapper`?

This makes it a "virtual" method so subclasses can override method `getDbWrapper` to return any db wrapper

cdc_admin/cdc_admin_handler.h
```diff
@@ -73,6 +74,10 @@ class CDCAdminHandler : virtual public CdcAdminSvIf {
   // of overriding some admin functions
   common::ObjectLock<std::string> db_admin_lock_;
 
+  virtual std::unique_ptr<replicator::DbWrapper> getDbWrapper(std::string db_name) {
```
kangnanli, line 77: `const std:string&`

kangnanli (approved):

cdc_admin/cdc_admin_handler.h
rajathprasad: I would prefer the default be the generic CDC implementation where we actually publish the key and value to Kafka. Maybe add a TODO for doing that in the future?

iprentic: I feel it will be hard to create a generic kafka publisher in the open source repo since others may configure their service and kafka clusters differently... My preference would be to leave it like this for now, and we can remember it as an internal TODO :)

rajathprasad (approved):

PR #549, opened 2021-09-10
In CDC admin, handle exceptions from creating the DB wrapper instance

iprentic:


cdc_admin/cdc_admin_handler.cpp
```diff
@@ -122,20 +122,21 @@ void CDCAdminHandler::async_tm_addObserver(
...
+    if (!db_manager_->addDB(request->db_name,
+                            std::move(db_wrapper),
+                            role,
+                            std::move(upstream_addr),
+                            replicator_zk_cluster,
+                            replicator_helix_cluster,
+                            &err_msg)) {
+      e.errorCode = CDCAdminErrorCode::ADMIN_ERROR;
+      e.message = std::move(err_msg);
+      throw e;
```
kangnanli, line 136: why do we throw here? 
maybe just `callback.release()->exceptionInThread(std::move(ex));`?

cdc_admin/cdc_admin_handler.cpp
```diff
@@ -122,20 +122,21 @@ void CDCAdminHandler::async_tm_addObserver(
...
-                          replicator_zk_cluster,
-                          replicator_helix_cluster,
-                          &err_msg)) {
-    e.errorCode = CDCAdminErrorCode::ADMIN_ERROR;
-    e.message = std::move(err_msg);
-    callback.release()->exceptionInThread(std::move(e));
-    return;
+  try {
+    std::unique_ptr<replicator::DbWrapper> db_wrapper = getDbWrapper(request->db_name);
+    if (!db_manager_->addDB(request->db_name,
```
kangnanli, line 127: In what situation, this addDB will fail?

cdc_admin/cdc_admin_handler.cpp
iprentic: Thought I would just consolidate since we do exactly the same thing below, but we can do that too!

cdc_admin/cdc_admin_handler.cpp
iprentic:
```
    if (dbs_.find(db_name) != dbs_.end()) {
      if (error_message) {
        *error_message = db_name + " has already been added";
      }
      return false;
    }
```

kangnanli: 1. should we use `<string*>->assign`? I have never see this type `*error_message = db_name + " has already been added";` before. If it is common, can u link 1 or 2 examples.
2. When the db already exist, and `addDB` fail, do you really need to throw exception back? you will silently log the error and treat it as OK?

iprentic: @kangnanli actually this code is copied from `rocksdb_admin`: https://github.com/pinterest/rocksplicator/blob/master/rocksdb_admin/application_db_manager.cpp#L49
I feel it is good to keep consistent with the existing behavior. And to be aware if the same DB is being added twice, as that indicates something is wrong that we should debug, not silently let happen.

cdc_admin/cdc_admin_handler.cpp
```diff
@@ -122,20 +122,21 @@ void CDCAdminHandler::async_tm_addObserver(
...
+                            std::move(upstream_addr),
+                            replicator_zk_cluster,
+                            replicator_helix_cluster,
+                            &err_msg)) {
+      e.errorCode = CDCAdminErrorCode::ADMIN_ERROR;
+      e.message = std::move(err_msg);
+      callback.release()->exceptionInThread(std::move(e));
+    }
+  } catch (CDCAdminException& ex) {
+    callback.release()->exceptionInThread(std::move(ex));
```
kangnanli, line 139: do u still need try catch here? Line 135-136 seem already handled the exception to client.

iprentic: This PR is to handle an exception raised by `std::unique_ptr<replicator::DbWrapper> db_wrapper = getDbWrapper(request->db_name);` 😅 
In RocksObserver db wrapper implementation, this call may raise an exception (e.g. if the kafka configuration is not correct)

kangnanli (approved):

PR #553, opened 2021-10-27
Correctly pass in replication mode to the replication handler

premkumr:
Correctly pass in replication mode for replication mode replication handler to deal with.

Details
-------
- The Config was not passed into the replication handler
- So the threads were waiting at Numberbox::wait for the repl to complete
- But as the repl-handler did know about the repl-mode, it did not do a post
- So all the waits timedout thereby increasing the latency ..

Tests
------
- Checked with various rep modes[0,1,2]
- Made sure it is reflected correctly on writes and in the replication handler
- No latency spikes
- Also ran Repl Consistency checker to verify

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -95,8 +95,13 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
   logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
   
   // for now we have to support both till all clusters are migrated
-  auto replication_mode =  config.replication_mode > FLAGS_replicator_replication_mode ? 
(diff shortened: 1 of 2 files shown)


jaricftw:
Posting a discussion from slack so we don't lose the context:

With replication mode mutated on a per-request/write basis, it seems we can easily run into race conditions like:
1. writeA with mode 2, then immediately followed by writeB with mode 0
2. replicatorDB update its local mode to be 0
3. replicatorDB receives a `handleReplicateRequest` from a follower, since the current mode is 0, it won’t post the `max_seq_no_acked_`
4. writeA now times out

Alos, do we want to support dynamically setting replication mode (0/1/2) on a pre-request/write basis? What's the use case? I can imagine we do that on a per-resource/dataset level (based on dataset config) if needed, but probably shouldn't change that during runtime.

PR #554, opened 2021-11-05
implement az specific routing as mentioned in parameter

monilsanghavi:
The feature will enable the ability to select hosts based only on the az specified by the caller. In our service, we plan to use it to have expensive queries routed only to a specific leaf cluster so that the other replicas are not affected by them (for protection). In the future, we also plan to use this for routing to multiple specific clusters (for concurrency).

### DIFF
--- a/common/tests/thrift_router_test.cpp
+++ b/common/tests/thrift_router_test.cpp
@@ -535,8 +535,7 @@ TEST(ThriftRouterTest, SpecificAzTest) {
     EXPECT_EQ(h->nPings_.load(), 1);
   }
 
-  // Get the client from local az for specific shard
-  // All requests should hit the local az handler
+  // Get the client from specified az for specific shard
   for (int i = 0; i < 100; i ++) {
     v.clear();
     EXPECT_EQ(
@@ -547,8 +546,7 @@ TEST(ThriftRouterTest, SpecificAzTest) {
   }
   EXPECT_EQ(handlers[1]->nPings_.load(), 101);
 
-  // Get the client from local az
-  // All requests should hit the local az handler
+  // Get the client from specified az
   for (int i = 0; i < 100; i ++) {
     std::map<uint32_t, std::vector<std::shared_ptr<DummyServiceAsyncClient>>> m;
     m[2];
@@ -573,7 +571,6 @@ TEST(ThriftRouterTest, SpecificAzTest) {
 }
 
 TEST(ThriftRouterTest, SpecificAzMasterSlaveTest) {
-  FLAGS_always_prefer_local_host = true;
   updateConfigFile(g_config_v4);
   ThriftRouter<DummyServiceAsyncClient> router(
     "us-east-1a", g_config_path, common::parseConfig);
@@ -613,8 +610,7 @@ TEST(ThriftRouterTest, SpecificAzMasterSlaveTest) {
   }
   EXPECT_EQ(handlers[2]->nPings_.load(), 100);
 
-  // Get the client from local az
-  // All requests should hit the local az handler
+  // Get the client from host which is slave in the specified az
   for (int i = 0; i < 100; i ++) {
     v.clear();
     EXPECT_EQ(
@@ -625,6 +621,23 @@ TEST(ThriftRouterTest, SpecificAzMasterSlaveTest) {
   }
   EXPECT_EQ(handlers[3]->nPings_.load(), 100);
 
+  // Get the client from host which is master in the specified az
+  for (int i = 0; i < 100; i ++) {
+    v.clear();
+    EXPECT_EQ(
+        router.getClientsFor("user_pins", Role::MASTER, Quantity::ONE, 2, &v, "us-east-1b"),
+        ReturnCode::OK);
+    EXPECT_EQ(v.size(), 1);
+    v[0]->future_ping().get();
+  }
+  EXPECT_EQ(handlers[2]->nPings_.load(), 200);
+
+  // no hosts with the role in the az specified
+  EXPECT_EQ(
+    router.getClientsFor("user_pins", Role::MASTER, Quantity::ONE, 2, &v, "us-east-1a"),
+    ReturnCode::NOT_FOUND);
+  EXPECT_EQ(v.size(), 0);
+
   // stop all servers
   for (auto& s : servers) {
     s->stop();
@@ -633,7 +646,6 @@ TEST(ThriftRouterTest, SpecificAzMasterSlaveTest) {
   for (auto& t : thrs) {
     t->join();
   }
-  FLAGS_always_prefer_local_host = false;
 }
 
 TEST(ThriftRouterTest, ForeignAzTest) {
--- a/common/tests/thrift_router_test.cpp
+++ b/common/tests/thrift_router_test.cpp
@@ -535,6 +535,18 @@ TEST(ThriftRouterTest, SpecificAzTest) {
     EXPECT_EQ(h->nPings_.load(), 1);
   }
 
+  // Get the client from local az for specific shard
+  // All requests should hit the local az handler
+  for (int i = 0; i < 100; i ++) {
+    v.clear();
+    EXPECT_EQ(
+        router.getClientsFor("user_pins", Role::ANY, Quantity::ALL, 2, &v, "us-east-1c"),
+        ReturnCode::OK);
+    EXPECT_EQ(v.size(), 1);
+    v[0]->future_ping().get();
+  }
+  EXPECT_EQ(handlers[1]->nPings_.load(), 101);
+
   // Get the client from local az
   // All requests should hit the local az handler
   for (int i = 0; i < 100; i ++) {
@@ -547,7 +559,7 @@ TEST(ThriftRouterTest, SpecificAzTest) {
     EXPECT_EQ(m[2].size(), 1);
     m[2][0]->future_ping().get();
   }
-  EXPECT_EQ(handlers[1]->nPings_.load(), 101);
+  EXPECT_EQ(handlers[1]->nPings_.load(), 201);
 
   // stop all servers
   for (auto& s : servers) {
--- a/common/tests/thrift_router_test.cpp
+++ b/common/tests/thrift_router_test.cpp
@@ -572,6 +572,70 @@ TEST(ThriftRouterTest, SpecificAzTest) {
   FLAGS_always_prefer_local_host = false;
 }
 
+TEST(ThriftRouterTest, SpecificAzMasterSlaveTest) {
+  FLAGS_always_prefer_local_host = true;
+  updateConfigFile(g_config_v4);
+  ThriftRouter<DummyServiceAsyncClient> router(
+    "us-east-1a", g_config_path, common::parseConfig);
+
+  std::vector<shared_ptr<DummyServiceAsyncClient>> v;
+  shared_ptr<DummyServiceTestHandler> handlers[4];
+  shared_ptr<ThriftServer> servers[4];
+  unique_ptr<thread> thrs[4];
+
+  tie(handlers[0], servers[0], thrs[0]) = makeServer(8090);
+  tie(handlers[1], servers[1], thrs[1]) = makeServer(8091);
+  tie(handlers[2], servers[2], thrs[2]) = makeServer(8092);
+  tie(handlers[3], servers[3], thrs[3]) = makeServer(8093);
+  sleep(1);
+
+  EXPECT_EQ(router.getShardNumberFor("user_pins"), 4);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 0), 4);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 1), 4);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 2), 4);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 3), 4);
+
+  // no hosts in the az specified
+  EXPECT_EQ(
+    router.getClientsFor("user_pins", Role::ANY, Quantity::ALL, 2, &v, "us-east-1f"),
+    ReturnCode::NOT_FOUND);
+  EXPECT_EQ(v.size(), 0);
+
+  // Get the client from local az for specific shard
+  // All requests should hit the local az handler
+  for (int i = 0; i < 100; i ++) {
+    v.clear();
+    EXPECT_EQ(
+        router.getClientsFor("user_pins", Role::ANY, Quantity::ONE, 2, &v, "us-east-1b"),
+        ReturnCode::OK);
+    EXPECT_EQ(v.size(), 1);
+    v[0]->future_ping().get();
+  }
+  EXPECT_EQ(handlers[2]->nPings_.load(), 100);
+
+  // Get the client from local az
+  // All requests should hit the local az handler
+  for (int i = 0; i < 100; i ++) {
+    v.clear();
+    EXPECT_EQ(
+        router.getClientsFor("user_pins", Role::SLAVE, Quantity::ONE, 2, &v, "us-east-1b"),
+        ReturnCode::OK);
+    EXPECT_EQ(v.size(), 1);
+    v[0]->future_ping().get();
+  }
+  EXPECT_EQ(handlers[3]->nPings_.load(), 100);
+
+  // stop all servers
+  for (auto& s : servers) {
+    s->stop();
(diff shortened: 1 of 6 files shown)


common/thrift_router.cpp
```diff
@@ -51,6 +51,7 @@ bool parseHost(const std::string& str, common::detail::Host* host,
     return false;
   }
   auto group = (tokens.size() == 3 ? tokens[2] : "");
+  host->az = group;
```
rajathprasad, line 54: This is actually a combination of availability zone + placement group. So I don't think az is the right name for this. Did you want to filter by AZ or PG or both?

common/thrift_router.h
```diff
@@ -377,7 +382,8 @@ class ThriftRouter {
         const Role role,
         const unsigned rotation_counter,
         const std::string& segment,
-        const int shrink_target) {
+        const int shrink_target,
+        const std::string& specific_az) {
       std::vector<const Host*> v;
       if (role == Role::ANY && !FLAGS_always_prefer_local_host) {
```
rajathprasad, line 388: Does this mean `FLAGS_always_prefer_local_host` should be True to be able to get clients by certain AZ/PG?

rajathprasad (commented):
Can you add a brief description on when this feature will be useful and how you're planning to use it?

common/thrift_router.cpp
monilsanghavi: Actually it's ok to only be AZ aware and not look into PG. Do you suggest fishing out the AZ from the string?

common/thrift_router.h
monilsanghavi: yes that was the idea. But I can make it generic and have this as the first condition. Would you like me to do that?

monilsanghavi:
> Can you add a brief description on when this feature will be useful and how you're planning to use it?

The feature will enable the ability to select hosts based only on the az specified by the caller. In our service, we plan to use it to have expensive queries routed only to a specific leaf cluster so that the other replicas are not affected by them (for protection). In the future, we also plan to use this for routing to multiple specific clusters (for concurrency).

common/thrift_router.h
```diff
@@ -367,6 +371,7 @@ class ThriftRouter {
      * hosts first by
      * (1) Prefer master to slave, then (2) Prefer local to non-local.
      *
+     * Otherwise, if az is specified by caller, return hosts belonging to that az
```
jaricftw, line 374: comments here is a bit confusing, Can you update them to be more clear, now that we have three conditions 
- AZ aware
- prefer master, then prefer local
- prefer local only

jaricftw:
> > Can you add a brief description on when this feature will be useful and how you're planning to use it?
> 
> The feature will enable the ability to select hosts based only on the az specified by the caller. In our service, we plan to use it to have expensive queries routed only to a specific leaf cluster so that the other replicas are not affected by them (for protection). In the future, we also plan to use this for routing to multiple specific clusters (for concurrency).

Can you move this description to the PR summary instead?

common/thrift_router.h
```diff
@@ -377,9 +382,20 @@ class ThriftRouter {
...
-        const int shrink_target) {
+        const int shrink_target,
+        const std::string& specific_az) {
       std::vector<const Host*> v;
-      if (role == Role::ANY && !FLAGS_always_prefer_local_host) {
+      if (specific_az != "") {
+        // if az is specified by caller, return hosts belong only to that az.
+        for (const auto& hi : host_info) {
+          if (hi.first->az.compare(specific_az) == 0) {
+            v.push_back(hi.first);
```
jaricftw, line 392: what if there is no matched hosts for the AZ?

common/tests/thrift_router_test.cpp
```diff
@@ -496,6 +496,63 @@ TEST(ThriftRouterTest, LocalAzTest) {
...
+  EXPECT_EQ(v.size(), 3);
+  for (auto client : v) {
+    EXPECT_NO_THROW(client->future_ping().get());
+  }
+  for (const auto& h : handlers) {
+    EXPECT_EQ(h->nPings_.load(), 1);
+  }
+
+  // Get the client from local az
+  // All requests should hit the local az handler
```
jaricftw, line 532: can we add test cases for:
- no hosts in the AZ
- multiple matches

common/thrift_router.h
monilsanghavi: added the comments now

common/thrift_router.h
```diff
@@ -190,10 +192,11 @@ class ThriftRouter {
       const Role role,
       const Quantity quantity,
       std::map<ShardID, std::vector<std::shared_ptr<ClientType>>>*
-        shard_to_clients) {
+        shard_to_clients,
+      const std::string& specific_az = "") {
```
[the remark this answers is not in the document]
monilsanghavi: added both comments and the argument for the above function as well.

common/tests/thrift_router_test.cpp
monilsanghavi: added the test cases

common/thrift_router.h
monilsanghavi: if there are no matched hosts, no clients will be returned as user explicitly wants the clients of the az specified

common/thrift_router.cpp
```diff
@@ -51,6 +51,8 @@ bool parseHost(const std::string& str, common::detail::Host* host,
     return false;
   }
   auto group = (tokens.size() == 3 ? tokens[2] : "");
+  auto pos = group.find("_");
+  host->az = (pos != 0 && pos != std::string::npos) ? group.substr(0, pos) : group;
```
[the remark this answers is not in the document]
monilsanghavi: format is "az_pg". This I got from the shard map generated by rocksplicator for the hosts it manages. If no "_" found, I default the az to be group.

common/thrift_router.h
```diff
@@ -377,9 +382,20 @@ class ThriftRouter {
         const Role role,
         const unsigned rotation_counter,
         const std::string& segment,
-        const int shrink_target) {
+        const int shrink_target,
+        const std::string& specific_az) {
       std::vector<const Host*> v;
-      if (role == Role::ANY && !FLAGS_always_prefer_local_host) {
+      if (specific_az != "") {
```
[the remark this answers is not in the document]
monilsanghavi: I see your point. I will change the code to use the same logic as when we sort hosts in the local az.

monilsanghavi:
> > > Can you add a brief description on when this feature will be useful and how you're planning to use it?
> > 
> > 
> > The feature will enable the ability to select hosts based only on the az specified by the caller. In our service, we plan to use it to have expensive queries routed only to a specific leaf cluster so that the other replicas are not affected by them (for protection). In the future, we also plan to use this for routing to multiple specific clusters (for concurrency).
> 
> Can you move this description to the PR summary instead?

done

common/thrift_router.cpp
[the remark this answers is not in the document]
jaricftw: yeah, I was more referring to if we have defined the format somewhere else in the codebase --- this is to avoid hard-coding the delimiter, and potentially reusing the parsing scheme if it's already defined somewhere. 

Right now this feels very error-prone, if the format changes, the parsing here breaks

common/thrift_router.h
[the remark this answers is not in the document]
jaricftw: yeah, sounds good

common/tests/thrift_router_test.cpp
```diff
@@ -496,6 +496,70 @@ TEST(ThriftRouterTest, LocalAzTest) {
...
+  sleep(1);
+
+  EXPECT_EQ(router.getShardNumberFor("user_pins"), 3);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 0), 3);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 1), 3);
+  EXPECT_EQ(router.getHostNumberFor("user_pins", 2), 3);
+
+  // no hosts in the az specified
+  EXPECT_EQ(
+    router.getClientsFor("user_pins", Role::ANY, Quantity::ALL, 2, &v, "us-east-1f"),
```
jaricftw, line 522: can we get test coverage for the other `getClientsFor` with specific shardID? (because it's also modified as part of this PR)

common/thrift_router.cpp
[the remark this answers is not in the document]
monilsanghavi: I tried to find it in the codebase but was unable to do so. I got the format from the shard map files generated. I agree this can be error prone. however, to save us from the different formats, I am just substituting to use the whole group.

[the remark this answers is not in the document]
jaricftw: Looks like the format is defined in the Java code: https://github.com/pinterest/rocksplicator/blob/63c9c7d738dc868e59c552a43e2b305314092b0c/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java#L414 where the shard_map is generated, and documented here: https://github.com/pinterest/rocksplicator/blob/63c9c7d738dc868e59c552a43e2b305314092b0c/cluster_management/src/main/java/com/pinterest/rocksplicator/ConfigGenerator.java#L69

Can you add a comment here so readers know why it's parsed this way?

Ideally, we should extract the parsing scheme as a common utilization function, maybe in https://github.com/pinterest/rocksplicator/blob/63c9c7d738dc868e59c552a43e2b305314092b0c/common/availability_zone.h, but that can be a follow-up discussion.

common/thrift_router.h
```diff
@@ -377,9 +383,20 @@ class ThriftRouter {
         const Role role,
         const unsigned rotation_counter,
         const std::string& segment,
-        const int shrink_target) {
+        const int shrink_target,
+        const std::string& specific_az) {
```
jaricftw, line 387: sorry, just realize one more thing: if user specified a AZ, and in the meantime wants a role `master/leader`, the currently implementation will ignore the role completely, but that may be surprising. Should we take `role` into consideration when doing AZ aware routing?

common/thrift_router.cpp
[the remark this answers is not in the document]
monilsanghavi: I added the comment speciyfing this format

common/thrift_router.h
monilsanghavi: yes, I added the code which will take role into consideration

jaricftw: The comments of this function says we firstly filter hosts by role, then sort them by other preferences, which I think makes sense. However, the implementation of the AZ aware routing firstly filters by AZ, then sort them by role

monilsanghavi: fixed it

common/thrift_router.h
```diff
@@ -377,9 +387,30 @@ class ThriftRouter {
...
       std::vector<const Host*> v;
-      if (role == Role::ANY && !FLAGS_always_prefer_local_host) {
+      if (specific_az != "") {
+        v.reserve(host_info.size());
+        std::vector<const Host*> v_s;
+        // if az is specified by caller, return hosts belong only to that az.
+        for (const auto& hi : host_info) {
+          if (hi.first->az.compare(specific_az) == 0) {
+            if ((role == Role::ANY && hi.second == Role::MASTER) || role == hi.second) {
```
jaricftw, line 399: if `role == Role::SLAVE`, this means we'll push slave nodes to `v`, and master to `v_s`, and select both in the final results, is that intended?

common/thrift_router.h
monilsanghavi: instead of adding separate condition, I am checking of specific_az right before inserting into the host array.

common/thrift_router.h
```diff
@@ -361,31 +368,35 @@ class ThriftRouter {
...
         // prefer MASTER, then prefer groups with longer common prefix length
         std::vector<const Host*> v_s;
         for (const auto& hi : host_info) {
-          if (hi.second == Role::SLAVE) {
+          if (hi.second == Role::SLAVE &&
+              (specific_az == "" || hi.first->az.compare(specific_az) == 0)) {
             v_s.push_back(hi.first);
-          } else if (hi.second == Role::MASTER) {
+          } else if (hi.second == Role::MASTER
+                     && (specific_az == "" || hi.first->az.compare(specific_az) == 0)) {
```
jaricftw, line 399: no need to do the check multiple places, consider:
```
for (const auto& hi : host_info) {
    if (specific_az != "" && hi.first->az.compare(specific_az) != 0) {
            continue
    }
    if (hi.second == Role::SLAVE ) {
    } else ...
}
```

common/thrift_router.h
```diff
@@ -399,7 +410,8 @@ class ThriftRouter {
         // prefer local
         v.reserve(host_info.size());
         for (const auto& hi : host_info) {
-          if (role == Role::ANY || hi.second == role) {
+          if ((role == Role::ANY || hi.second == role) &&
+              (specific_az == "" || hi.first->az.compare(specific_az) == 0)) {
```
jaricftw, line 414: avoid boiler plate code, consider extracting a common function for the az check, e.g.

```
bool matchAZ(const string& az, const string& specific_az) {
   return specific_az != "" && az.compare(specific_za) == 0
}
```

jaricftw (commented):
The latest revision is in a better shape, though we'll have to add the AZ check in multiple places -- not quite desirable, ideally we'd only filter by AZ once for all checks in one place. I think that's doable, but would require some refactor that can be followed-up separately.

common/thrift_router.h
monilsanghavi: done

common/thrift_router.h
monilsanghavi: done

common/tests/thrift_router_test.cpp
```diff
@@ -496,6 +496,158 @@ TEST(ThriftRouterTest, LocalAzTest) {
...
+    v.clear();
+    EXPECT_EQ(
+        router.getClientsFor("user_pins", Role::SLAVE, Quantity::ONE, 2, &v, "us-east-1b"),
+        ReturnCode::OK);
+    EXPECT_EQ(v.size(), 1);
+    v[0]->future_ping().get();
+  }
+  EXPECT_EQ(handlers[3]->nPings_.load(), 100);
+
+  // Get the client from host which is master in the specified az
```
jaricftw, line 624: can you add test case for Role::ANY as well

jaricftw (approved):
Thanks for addressing the comments

common/tests/thrift_router_test.cpp
monilsanghavi: already added in line 603-611

common/thrift_router.cpp
```diff
@@ -51,6 +51,11 @@ bool parseHost(const std::string& str, common::detail::Host* host,
     return false;
   }
   auto group = (tokens.size() == 3 ? tokens[2] : "");
+  // Based on format of generated shard config which is
+  //    ${ip_address}:${thrift_service_port}:${az}_${pg}
+  // as specified in ConfigGenerator.java, we try to cache the az information.
```
rajathprasad, line 56: Nit, but I think it makes it more clear to move this section of comments above line 53. Or above 43, to be even more clear.

common/thrift_router.h
```diff
@@ -481,6 +496,10 @@ class ThriftRouter {
       }
     }
 
+    bool matchAZ(const std::string& host_az, const std::string& specific_az) {
+       return (specific_az == "" || host_az.compare(specific_az) == 0);
```
rajathprasad, line 500: Is there a reason we have the `specific_az == ""` check? The next check should take care of it, right?

rajathprasad (approved):
A couple more minor comments. But overall, it looks good!

common/thrift_router.h
jaricftw: the `specific_az == ""` check is so we ignore checking AZ when it's not specified, i.e. always returning `true` here. 

Maybe the function can be renamed to make it clearer, how about `matchAZIfNotEmpty`

common/thrift_router.cpp
monilsanghavi: done

common/thrift_router.h
monilsanghavi: compare returns a non zero value if specific_az is "". Hence we need the first condition. Basically, if specific_az is set to "", we do not have to do string compare.

common/thrift_router.cpp
monilsanghavi: done

PR #555, opened 2021-11-17
ConfigManager to manage global dataset level Config within rocksplicator

premkumr:
- The Value will be set from rockstore/rockstoreWC
-

### DIFF
--- a/common/config.h
+++ b/common/config.h
@@ -18,6 +18,7 @@
 
 #pragma once
 #include <cstdint>
+#include <memory>
 
 namespace common {
   struct Config {
@@ -35,4 +36,7 @@ namespace common {
       return c;
     }
   };
+
+  using ConfigPtr = std::shared_ptr<const Config>;
+
 } //  namespace common
--- a/common/dbconfig.cpp
+++ b/common/dbconfig.cpp
@@ -1,0 +1,107 @@
+// Copyright 2021-present Pinterest. All Rights Reserved.
+//
+// @author Premkumar (prem@pinterest.com)
+//
+
+#include "common/dbconfig.h"
+#include "common/segment_utils.h"
+#include <utility>
+
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+
+DECLARE_int32(replicator_replication_mode);
+
+namespace common {
+
+DBConfigManager* DBConfigManager::get() {
+  static DBConfigManager instance;
+  return &instance;
+}
+
+DBConfigManager::DBConfigManager() : dbConfig_(new DBConfig()), fDataLoaded_(false) {
+}
+
+bool DBConfigManager::loadJsonStr(const std::string& content) {
+  Json::Value root;
+  Json::Reader reader;
+  if (!reader.parse(content, root) || !root.isObject()) {
+    LOG(ERROR) << "Could not load config";
+    return false;
+  }
+
+  return loadJsonObject(root);
+}
+
+bool DBConfigManager::loadJsonObject(const Json::Value& root) {
+  try {
+    auto newConfig = parseConfig(root);
+    if (newConfig == nullptr) {
+      return false;
+    }
+
+    LOG(INFO) << "Successfully updated config";
+    std::atomic_exchange_explicit(&dbConfig_, newConfig, std::memory_order_release);
+    fDataLoaded_.store(true);
+    return true;
+  } catch (std::exception& ex) {
+    LOG(ERROR) << "Exception from parseConfig: " << ex.what();
+    return false;
+  }
+}
+
+DBConfigPtr DBConfigManager::parseConfig(const Json::Value& root) {
+  static const std::string ACK_MODE = "ack_mode";
+  static const std::string DATASET = "dataset";
+
+  DBConfig newDBConfig;
+
+  if (root.isMember(DATASET)) {
+    // parse dataset specific config
+    auto jsonDataSetConfigs = root[DATASET];
+
+    for (const auto& dataset : jsonDataSetConfigs.getMemberNames()) {
+      std::shared_ptr<Config> config = std::make_shared<Config>();
+      const auto& jsonDataSetConfig = jsonDataSetConfigs[dataset];
+      bool isSet = false; /* creating this for future multiple assigns */
+
+      if(jsonDataSetConfig.isMember(ACK_MODE)) {
+        config->replication_mode = jsonDataSetConfig[ACK_MODE].asUInt();
+        isSet = true;
+      }
+
+      if (isSet) {
+        newDBConfig.dataSetConfigMap[dataset] = config;
+      }
+    }
+  }
+  if (newDBConfig.dataSetConfigMap.size() == 0) {
+    // no data was loaded..
+    return nullptr;
+  }
+  return std::make_shared<const DBConfig>(std::move(newDBConfig));
+}
+
+/* conf getter functions */
+ConfigPtr DBConfigManager::getConfig(const std::string& dbName) const {
+  auto dsName = DbNameToSegment(dbName);
+  const auto dbconfigbak = std::atomic_load_explicit(&dbConfig_, std::memory_order_acquire);
+  const auto& dsConfigMap = dbconfigbak->dataSetConfigMap;
+  const auto dsIter = dsConfigMap.find(dsName);
+  if (dsIter == dsConfigMap.end()) {
+    return nullptr;
+  } else {
+    return dsIter->second;
+  }
+}
+
+uint DBConfigManager::getReplicationMode(const std::string& dbName, uint defValue) const {
+  const ConfigPtr config = getConfig(dbName);
+  if (config != nullptr) {
+    return config->replication_mode;
+  }
+
+  return defValue;
+}
+
+}  // namespace common
--- a/common/dbconfig.h
+++ b/common/dbconfig.h
@@ -1,0 +1,78 @@
+// Copyright 2021-present Pinterest. All Rights Reserved.
+//
+// @author Premkumar (prem@pinterest.com)
+//
+
+#pragma once
+
+#include <atomic>
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "common/config.h"
+#include "common/jsoncpp/include/json/json.h"
+
+/**
+ * {
+ *   "dataset" : {
+ *      // dataset level configs
+ *        "ad_query_join" : {
+ *              "ack_mode" : 1,
+ *       }
+ *    }
+ * }
+ */
+
+namespace common {
+
+struct DBConfig {
+  std::unordered_map<std::string, ConfigPtr> dataSetConfigMap;
+};
+
+using DBConfigPtr  = std::shared_ptr<const DBConfig>;
+
+/**
+ * DBConfigManager manages the per-dataset and service level
+ * configuration for Rocksplicator.
+ * 
+ * The actual config is managed by the upper layer service and
+ * is passed on via loadJsonObject or loadJsonStr
+ * 
+ * Thread safety is guaranteed by atomically loading and swapping 
+ * the underlying ptr of the shared_ptr
+ * 
+ * Dangling ptrs of internals of the swapped object is avoided by using 
+ * shared_ptrs all over.
+ * 
+ * This will be used via a static singleton instance
+ */
+class DBConfigManager {
+public:
+  static DBConfigManager* get();
+
+
+  /**
+   * Both the load functions are used to load data after instantiation
+   * During init there may not be data for it to load.
+   **/
+  bool hasValidData() const {
+    return fDataLoaded_.load();
+  };
+  /* load from a json object */
+  bool loadJsonObject(const Json::Value& root);
+  /* load from json string */
+  bool loadJsonStr(const std::string& data);
+
+  /* conf getter functions */
+  uint getReplicationMode(const std::string& dbName, uint defValue = 0) const;
+
+private:
+  /* No direct instantiation. Use the global static via ::get()*/
+  DBConfigManager();
+  DBConfigPtr dbConfig_;
+  ConfigPtr getConfig(const std::string& dbName) const;
+  DBConfigPtr parseConfig(const Json::Value& root);
+  std::atomic<bool> fDataLoaded_;
+};
+}  // namespace common
--- a/common/tests/dbconfig_test.cpp
+++ b/common/tests/dbconfig_test.cpp
@@ -1,0 +1,94 @@
+#include "common/dbconfig.h"
+
+#include <string>
+#include <glog/logging.h>
+#include "gtest/gtest.h"
+
+
+namespace common {
+  /**
+   *  use https://www.freeformatter.com/json-escape.html#ad-output
+   *  to escape & unescape
+   */
+
+TEST(DBConfigTest, Basic) {
+  auto cfgMgr = DBConfigManager::get();
+  std::string content = R"({
+   "dataset": {
+    "testdataset" : {
+        "ack_mode" : 1
+      }
+    } 
+  })";
+
+  EXPECT_FALSE(cfgMgr->hasValidData());
+  cfgMgr->loadJsonStr(content);
+  EXPECT_TRUE(cfgMgr->hasValidData());
+  
+  EXPECT_EQ(cfgMgr->getReplicationMode("dbtest00001"), 0);
+  EXPECT_EQ(cfgMgr->getReplicationMode("testdataset00001"), 1);
+  EXPECT_NE(cfgMgr->getReplicationMode("db1_00001"), 1);
+  EXPECT_NE(cfgMgr->getReplicationMode("testdataset"), 1);
+
+  EXPECT_EQ(cfgMgr->getReplicationMode("dbtest00001", 1), 1);
+}
+
+TEST(DBConfigTest, BasicFromObject) {
+  auto cfgMgr = DBConfigManager::get();
+  std::string content = R"({
+   "dataset": {
+    "testdataset" : {
+        "ack_mode" : 1
+      }
+    }
+  })";
+
+  Json::Value root;
+  Json::Reader reader;
+  if (!reader.parse(content, root) || !root.isObject()) {
+    LOG(ERROR) << "Could not parse json config :" << content;
+  }
+
+  cfgMgr->loadJsonObject(root);
+
+  EXPECT_EQ(cfgMgr->getReplicationMode("dbtest00001"), 0);
+  EXPECT_EQ(cfgMgr->getReplicationMode("testdataset00001"), 1);
+  EXPECT_NE(cfgMgr->getReplicationMode("db1_00001"), 1);
+  EXPECT_NE(cfgMgr->getReplicationMode("testdataset"), 1);
+
+  EXPECT_EQ(cfgMgr->getReplicationMode("dbtest00001", 1), 1);
+}
+
+TEST(DBConfigTest, InvalidContent) {
+  auto cfgMgr = DBConfigManager::get();
+  std::string content = R"({
+  })";
+
+  Json::Value root;
+  Json::Reader reader;
+  if (!reader.parse(content, root) || !root.isObject()) {
+    LOG(ERROR) << "Could not parse json config :" << content;
+  }
+
+  EXPECT_FALSE(cfgMgr->loadJsonObject(root));
+}
+
+TEST(DBConfigTest, EmptyContent) {
+  auto cfgMgr = DBConfigManager::get();
+  std::string content = R"()";
+
+  Json::Value root;
+  Json::Reader reader;
+  if (!reader.parse(content, root) || !root.isObject()) {
+    LOG(ERROR) << "Could not parse json config :" << content;
+  }
+
+  EXPECT_FALSE(cfgMgr->loadJsonObject(root));
+}
+
+}  // namespace common
+
+int main(int argc, char** argv) {
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -22,6 +22,7 @@
 #include <string>
 #include <vector>
 
+#include "common/dbconfig.h"
 #include "common/helix_client.h"
 #include "common/segment_utils.h"
 #include "folly/MoveWrapper.h"
@@ -78,7 +79,7 @@ namespace replicator {
 rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
     const rocksdb::WriteOptions& options,
     rocksdb::WriteBatch* updates,
-    const common::Config& config,
+    const common::Config& , // TODO(prem) : remove this arg soon
     rocksdb::SequenceNumber* seq_no) {
   if (role_ == DBRole::SLAVE) {
     throw ReturnCode::WRITE_TO_SLAVE;
@@ -94,10 +95,12 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
   auto end = GetCurrentTimeMs();
   logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
   
+  auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
+  // TODO(prem) : remove support for gflags soon
   // for now we have to support both till all clusters are migrated
-  auto replication_mode =  config.replication_mode > FLAGS_replicator_replication_mode ? 
-                            config.replication_mode :  FLAGS_replicator_replication_mode;
-
+  if (FLAGS_replicator_replication_mode > replication_mode) {
+    replication_mode = FLAGS_replicator_replication_mode;
+  }
   if (status.ok()) {
     cond_var_.notifyAll();
 
@@ -305,7 +308,7 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
 void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
     std::unique_ptr<CallbackType> callback,
     std::unique_ptr<ReplicateRequest> request,
-    const common::Config& config) {
+    const common::Config&) { // TODO(prem) : remove this arg soon
   CHECK(request->db_name == db_name_);
 
   auto db = shared_from_this();
@@ -318,7 +321,11 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
     logMetric(kReplicatorLeaderSequenceNumbersBehind, seq_no - leaderSeqNum, db ->db_name_);
   }
 
-  auto replication_mode =  config.replication_mode > FLAGS_replicator_replication_mode ? config.replication_mode :  FLAGS_replicator_replication_mode;
+  auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
+  // for now we have to support both till all clusters are migrated
+  if (FLAGS_replicator_replication_mode > replication_mode) {
+    replication_mode = FLAGS_replicator_replication_mode;
+  }
   if (replication_mode == 1 || replication_mode == 2) {
     // post the largest sequence number the Slave has committed
     max_seq_no_acked_.post(seq_no);


common/dbconfig.cpp
```diff
@@ -0,0 +1,103 @@
...
+DECLARE_int32(replicator_replication_mode);
+
+namespace common {
+
+DBConfigManager* DBConfigManager::get() {
+  static DBConfigManager instance;
+  return &instance;
+}
+
+DBConfigManager::DBConfigManager() : dbConfig(new DBConfig()) {
```
jaricftw, line 22: C++ question: `new DBConfig()`, or shall we use `std::make_shared<DBConfig>()`

common/dbconfig.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+bool DBConfigManager::loadJsonStr(const std::string& content) {
+  Json::Value root;
+  Json::Reader reader;
+  if (!reader.parse(content, root) || !root.isObject()) {
```
jaricftw, line 28: nit: check `!root.isObject()` first?

common/dbconfig.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+std::shared_ptr<const DBConfig> DBConfigManager::parseConfig(const Json::Value& root) {
+  static const std::string ACK_MODE = "ack_mode";
+  static const std::string DATASET = "dataset";
+
+  DBConfig newDBConfig;
+
+  for (const auto& rootMember : root.getMemberNames()) {
+    // parse dataset specific config
+    if (rootMember == DATASET) {
```
jaricftw, line 60: nit: reduce nesting
```
if (rootMember != DATASET) {
   continue
}
....
```

common/dbconfig.h
```diff
@@ -0,0 +1,47 @@
...
+namespace common {
+
+struct DBConfig {
+  std::unordered_map<std::string, Config> dataSet;
+};
+
+class DBConfigManager {
+public:
+  static DBConfigManager* get();
+  bool loadJsonObject(const Json::Value& root);
```
jaricftw, line 35: it seems `loadJsonObject` is only a util function used by `loadJsonStr`, do we need to make it public?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -78,7 +79,7 @@ namespace replicator {
 rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
     const rocksdb::WriteOptions& options,
     rocksdb::WriteBatch* updates,
-    const common::Config& config,
+    const common::Config& , // ignore this value
```
jaricftw, line 82: should we remove this after the new change is live? maybe a TODO

rocksdb_replicator/replicated_db.cpp
```diff
@@ -94,10 +95,11 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
   auto end = GetCurrentTimeMs();
   logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
   
+  auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
   // for now we have to support both till all clusters are migrated
-  auto replication_mode =  config.replication_mode > FLAGS_replicator_replication_mode ? 
-                            config.replication_mode :  FLAGS_replicator_replication_mode;
-
+  if (FLAGS_replicator_replication_mode > replication_mode) {
```
jaricftw, line 100: does this mean the global gflag will override the per-dataset config? is that indented?

common/dbconfig.h
kangnanli: +1. the two `load` APIs private should suffice

common/dbconfig.h
```diff
@@ -0,0 +1,47 @@
...
+ *              "ack_mode" : 1,
+ *       }
+ *    }
+ * }
+ */
+
+namespace common {
+
+struct DBConfig {
+  std::unordered_map<std::string, Config> dataSet;
```
kangnanli, line 29: should we rename to something like `segmentToConfigMap`, I feel like `dataSet` means a dataset struct.
e.g. at dbconfig.cpp L63, the `dataset` means `dataset name`. 

Now, `dataset` could cause confusion when read the code

common/dbconfig.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return nullptr;
+  } else {
+    return &(dsIter->second);
+  }
+}
+
+uint DBConfigManager::getReplicationMode(const std::string& dbName, uint defValue) const {
+  const auto* config = getConfig(dbName);
+  if (config != nullptr) {
+    return config->replication_mode;
```
kangnanli, line 97: I think this could cause data race. 
When we write/swap the dbconfig, we are using `std::atomic_exchange_explicit(&dbConfig, newConfig, std::memory_order_release);` which is release.
Here, when we read it, we did not enforce any memory order, e.g. `acquire`. 
- without memory order, btw L86 and L96, another write for `dbConfig` could happen, the read/write will have no synchoronization, then, causing data race.

common/dbconfig.h
```diff
@@ -0,0 +1,47 @@
...
+};
+
+class DBConfigManager {
+public:
+  static DBConfigManager* get();
```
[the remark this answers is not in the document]
premkumr: This is to make sure we have only one object alive , which will be globally shared within the process.

common/dbconfig.h
```diff
@@ -0,0 +1,47 @@
...
+  bool loadJsonObject(const Json::Value& root);
+  bool loadJsonStr(const std::string& data);
+
+  /* conf getter functions */
+  uint getReplicationMode(const std::string& dbName, uint defValue = 0) const;
+
+private:
+  DBConfigManager();
```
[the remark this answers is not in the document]
premkumr: This is to avoid instantiating separate objects & force anyone using the class to use the static instance

common/dbconfig.cpp
premkumr: `std::make_shared<DBConfig>()` would be typically cleaner and should be used in normal init. This being in the initializer list of ctor, `dbConfig(new DBConfig())` is shorter ..

common/dbconfig.cpp
premkumr: `isObject` refers  to check if the json in `{..}` (Json Object) format & not an array `[..]` or just bool like `name:prem` .. So root will be filled after parsing. Without parsing. root.isNull() will be true.

common/dbconfig.cpp
premkumr: Good point. I'll re-write this part

common/dbconfig.h
premkumr: `loadJsonStr` is for testing and for future use. `loadJsonObject` is what will be used from serviceconfig

rocksdb_replicator/replicated_db.cpp
premkumr: Yes We need to do that , but will have that as a separate change ...

rocksdb_replicator/replicated_db.cpp
premkumr: For now yes.. This is because rocksplicator is used by multiple services. Most of them use this flag. Till we migrate all of them , we need to give it priority. Once we migrate all services out of gflags, we will remove this check

common/dbconfig.h
premkumr: good point will fix the dataset variable..

common/dbconfig.cpp
premkumr: Will fix this .

common/dbconfig.h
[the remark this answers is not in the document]
jaricftw: oh, a singleton? Just searching for best practice here and ran into https://stackoverflow.com/questions/1008019/c-singleton-design-pattern, a couple of things as I noticed:
- do we wanna return a reference v.s. pointer?
- shall we ensure no copies are made (with those `delete` methods)

common/dbconfig.h
jaricftw: I don't see unit testing for `loadJsonObject`, we could either make this private for now (can always make it public when needed), or add unit testing for any public APIs.

jaricftw: Also, can you add comments for these APIs

rocksdb_replicator/replicated_db.cpp
jaricftw: I see..
this can be very confusing and error-prone, we'll need to distinguish between:
- default ack mode
- per-dataset override
- global gflag override
At least we need to document somewhere in the code on how they work together currently, and how we plan to migrate moving forward.

jaricftw: Separately, what do you think of always using `FLAGS_replicator_replication_mode` as a global default for a service, and per-dataset config always override that?

common/dbconfig.cpp
jaricftw: ah sorry, I confused myself

common/tests/dbconfig_test.cpp
```diff
@@ -0,0 +1,36 @@
...
+  })";
+  cfgMgr->loadJsonStr(content);
+  
+  EXPECT_EQ(cfgMgr->getReplicationMode("dbtest00001"), 0);
+  EXPECT_EQ(cfgMgr->getReplicationMode("testdataset00001"), 1);
+  EXPECT_NE(cfgMgr->getReplicationMode("db1_00001"), 1);
+  EXPECT_NE(cfgMgr->getReplicationMode("testdataset"), 1);
+
+  EXPECT_EQ(cfgMgr->getReplicationMode("dbtest00001", 1), 1);
+}
```
jaricftw, line 29: a couple edge test cases:
- empty content
- no `dataset` filed in json

common/dbconfig.cpp
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+  }
+  return std::make_shared<const DBConfig>(std::move(newDBConfig));
+}
+
+/* conf getter functions */
+const Config* DBConfigManager::getConfig(const std::string& dbName) const {
+  auto dsName = DbNameToSegment(dbName);
+  const auto& dsConfigMap = std::atomic_load_explicit(&dbConfig, std::memory_order_acquire)->dataSetConfigMap;
```
jaricftw, line 83: C++ questions:
it seems we just need to have read/write locks here (write lock on the write path, read lock on the read path), what's the best practice we do with C++? or why we chose `atomic_load_explicit` and `atomic_exchange_explicit`?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -94,10 +95,11 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
   auto end = GetCurrentTimeMs();
   logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
   
+  auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
```
jaricftw, line 98: how do we know/ensure `DBConfigManager` has loaded the config for usage here?

common/dbconfig.cpp
```diff
@@ -0,0 +1,101 @@
...
+    return nullptr;
+  } else {
+    return &(dsIter->second);
+  }
+}
+
+uint DBConfigManager::getReplicationMode(const std::string& dbName, uint defValue) const {
+  const auto* config = getConfig(dbName);
+  if (config != nullptr) {
+    return config->replication_mode;
```
jaricftw, line 95: is there still a race condition here, since `getConfig` returns a pointer here there is no locking in reading its `replication_mode` --- it could have been mutated after `getConfig` is retuned

common/dbconfig.cpp
kangnanli: I did some search, and find this: https://stackoverflow.com/a/15067071, and https://stackoverflow.com/a/18754519.

Lock is expensive; while, here: all we need is, read and operations after the read are ensured to see the written value. Thus, we need a synchronization between the write/read. 
refer to : https://www.modernescpp.com/index.php/acquire-release-fences

rocksdb_replicator/replicated_db.cpp
kangnanli: static variable inside a funct will be instantiated during the fist call. And, `static` is guaranteed to be instantiated only once, thus, no race condition.

kangnanli (approved):
I don't have any concern from my reading.

rocksdb_replicator/replicated_db.cpp
jaricftw: which `static` variable are you referring to @kangnanli ?

I am asking the underlying json config loaded & parsed by `DBConfigManager`, since that's not created as part of the construction/initialization, how do we ensure it's available when we read here?

kangnanli: O. I misunderstood the issue. Yes, you are right. 
`common::DBConfigManager::get()` can't guarantee the underlying json file is actually parsed and loaded into memory. That's why at the test above, it has to load it explicitly `cfgMgr->loadJsonStr(content);`. @premkumr any reason, why we want to load explicitly, instead of loading it at the DBConfigManager instantiation time at Line17-23 at dbconfig.cpp?

kangnanli (changes requested):

common/dbconfig.cpp
jaricftw: looking at https://www.cplusplus.com/reference/atomic/atomic_exchange_explicit/, it takes a `std::atomic<T>*`, but `&dbConfig` isn't? How does this work without a lock

rocksdb_replicator/replicated_db.cpp
premkumr: During instantiation , the json config my not be available , the config fields will be populated when the config file is present and is passed on from rockstore/wc

common/dbconfig.cpp
premkumr: The functions are for swapping shared_ptrs atomically ..

common/dbconfig.cpp
premkumr: True, We can leave that as opposed to using a lock/unlock by all threads ..

common/dbconfig.h
premkumr: Done ..

common/tests/dbconfig_test.cpp
premkumr: done

rocksdb_replicator/replicated_db.cpp
premkumr: just in case , I've added a variable to verify data load. Without data load, the functions will respond with default values

common/dbconfig.cpp
jaricftw: got it, thanks. https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

rocksdb_replicator/replicated_db.cpp
jaricftw: It's less about what we do when the config is not available, but rather can we avoid such race conditions at the first place? Typically a server should firstly load/initialize all its configs, before it start serving any runtime requests --- is that something we can achieve with rockstore/rocksplicator?

common/dbconfig.cpp
jaricftw: hmm, we just need read lock here, so no contention for concurrent reads between threads. Only lock contention will happen is when there is a concurrent write (i.e. when config gets updated), which should rarely happen. 

(correctness over performance)

rocksdb_replicator/replicated_db.cpp
premkumr: We can handle this at the service level, where we can wait for the config file to be available just like how we now want to wait for the shardmap to be available

common/dbconfig.cpp
```diff
@@ -0,0 +1,102 @@
...
+      if (isSet) {
+        newDBConfig.dataSetConfigMap[dataset] = config;
+      }
+    }
+  }
+  return std::make_shared<const DBConfig>(std::move(newDBConfig));
+}
+
+/* conf getter functions */
+const Config* DBConfigManager::getConfig(const std::string& dbName) const {
```
newpoo, line 82: The returned Config pointer is not guaranteed to be valid, as it may outlive the dbConfig object.

common/dbconfig.h
```diff
@@ -0,0 +1,60 @@
...
+  /* conf getter functions */
+  uint getReplicationMode(const std::string& dbName, uint defValue = 0) const;
+
+private:
+  /* No direct instantiation. Use the global static via ::get()*/
+  DBConfigManager();
+  std::shared_ptr<const DBConfig> dbConfig;
+  const Config* getConfig(const std::string& dbName) const;
+  std::shared_ptr<const DBConfig> parseConfig(const Json::Value& root);
+  bool fDataLoaded = false;
```
newpoo, line 58: can you append "_" to all data members to follow the naming convention?

rocksdb_replicator/replicated_db.cpp
newpoo: We are doing a lookup for every write. any way to optimize it out?

common/dbconfig.h
```diff
@@ -0,0 +1,60 @@
...
+ * }
+ */
+
+namespace common {
+
+struct DBConfig {
+  std::unordered_map<std::string, Config> dataSetConfigMap;
+};
+
+class DBConfigManager {
```
newpoo, line 32: Are all public interfaces thread-safe? Can you please add some comments for that?

newpoo (commented):
A high level question: can we please add some comments to document the features? such as if per write replication mode change is supported, how application is expected to integrate with it?

common/dbconfig.cpp
premkumr: Fixed this by making a copy of the shared pointer in getConfig and returning a shared_ptr of Config..

rocksdb_replicator/replicated_db.cpp
premkumr: The map size is very small & for now it is just 1 config.  Should not be an issue. if it becomes a concern, we can have hook that will update a local variable on config change.

jaricftw: chat offline with @premkumr about this --- we agree ideally we should avoid a global config and should instead pass down config via initialization, but that may require lots of refactors and plumbing config through rockstore to rocksplicator handlers, and makes dynamic config updates tricky.

rocksdb_replicator/replicated_db.cpp
jaricftw: @premkumr I still have the above questions, can you clarify the order of applying different overrides to the ack mode

rocksdb_replicator/replicated_db.cpp
jaricftw: can you add a TODO to make this clear?

common/dbconfig.h
jaricftw: +1

premkumr: Not all public interfaces are thread safe. The initialization happens via a single instance (from ServiceConfig) and atomically swapped. The shared_ptr after atomic load, takes care of thread safety automatically

rocksdb_replicator/replicated_db.cpp
premkumr: sure will do ..

rocksdb_replicator/replicated_db.cpp
premkumr: @jaricftw oops, I completely skipped this.. Sorry about that ..

> Separately, what do you think of always using FLAGS_replicator_replication_mode as a global default for a service, and per-dataset config always override that?

Technically you are correct. Right now there are only 1/2 datasets that actually use the setting. And we are planning to turn this setting OFF at the gflags and completely control via the config. the default for the gflag is 0  & the default for the config struct is 0 ... So it is going to be controlled by the passed in config...  As we are planning to deprecate this asap, and support the cluster where we are not passing the config file, it would be better to  pick the max for now.. I'll add a TODO.. what do you think ?

common/dbconfig.cpp
premkumr: Have Changed this to Shared_ptr

common/dbconfig.h
premkumr: Done

common/dbconfig.h
premkumr: updated with comments

common/dbconfig.h
```diff
@@ -0,0 +1,75 @@
...
+public:
+  static DBConfigManager* get();
+
+
+  /**
+   * Both the load functions are used to load data after instantiation
+   * During init there may not be data for it to load.
+   **/
+  bool hasValidData() const {
```
jaricftw, line 56: can we add unit test for `hasValidData`

common/dbconfig.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /* conf getter functions */
+  uint getReplicationMode(const std::string& dbName, uint defValue = 0) const;
+
+private:
+  /* No direct instantiation. Use the global static via ::get()*/
+  DBConfigManager();
+  DBConfigPtr dbConfig_;
+  ConfigPtr getConfig(const std::string& dbName) const;
+  DBConfigPtr parseConfig(const Json::Value& root);
```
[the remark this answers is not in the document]
premkumr: parseConfig actually updates the member variable. So it cannot be const.

common/dbconfig.h
premkumr: Will do

common/config.h
```diff
@@ -18,6 +18,7 @@
 
 #pragma once
 #include <cstdint>
+#include <memory>
```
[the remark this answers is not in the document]
premkumr: that is where the atomic stuff is defined.

common/dbconfig.h
```diff
@@ -0,0 +1,75 @@
...
+  bool fDataLoaded_ = false;
```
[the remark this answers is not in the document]
premkumr: for primitive bool, making atomic is not much value , esp here. no harm though..

common/config.h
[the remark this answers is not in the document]
jaricftw: oh, especially the `std::shared_ptr` in this file right

common/dbconfig.h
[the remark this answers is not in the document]
jaricftw: oh I misread, sorry

common/dbconfig.h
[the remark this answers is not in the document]
jaricftw: yeah, still seems safer: https://stackoverflow.com/questions/16320838/when-do-i-really-need-to-use-atomicbool-instead-of-bool/16321311

common/dbconfig.cpp
```diff
@@ -0,0 +1,103 @@
...
+    std::atomic_exchange_explicit(&dbConfig_, newConfig, std::memory_order_release);
+    fDataLoaded_ = true;
+    return true;
+  } catch (std::exception& ex) {
+    LOG(ERROR) << "Exception from parseConfig: " << ex.what();
+    return false;
+  }
+}
+
+DBConfigPtr DBConfigManager::parseConfig(const Json::Value& root) {
```
newpoo, line 53: semantically, this should return a unique_ptr.

common/dbconfig.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return false;
+  }
+
+  return loadJsonObject(root);
+}
+
+bool DBConfigManager::loadJsonObject(const Json::Value& root) {
+  try {
+    auto newConfig = parseConfig(root);
+    if (newConfig == nullptr) {
```
newpoo, line 39: this will never be true. Because parseConfig() always returns a non-nullptr.

How do we want to handle a misformated config?

common/dbconfig.h
premkumr: Done

common/dbconfig.h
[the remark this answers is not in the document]
newpoo: In practice, it's ok to not use atomic<bool> on the platform that our systems run on.

In theory, it's a data race to read/update a bool from multiple threads, which can cause undefined behavior. Given that it's trivial to add atomic<>, I'd suggest we do that to make our code correct both in practice and in theory.

common/dbconfig.h
```diff
@@ -0,0 +1,75 @@
...
+namespace common {
+
+struct DBConfig {
+  std::unordered_map<std::string, ConfigPtr> dataSetConfigMap;
+};
+
+using DBConfigPtr  = std::shared_ptr<const DBConfig>;
+
+/**
+ * DBConfigManager manages the per-dataset and service level
```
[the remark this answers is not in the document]
premkumr: Done .

common/dbconfig.cpp
premkumr: Good Catch . We have changed the code so many times now.  It was supposed to  return nullptr ..  Will add that in.

common/dbconfig.h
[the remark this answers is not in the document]
premkumr: Done

premkumr:
Unless there are no logic issues, I would like to close this out by Friday .. Thanks

common/dbconfig.cpp
jaricftw: how do we decide it's unique_ptr or a shared_ptr here, do we expect the parsed config here to be used in multiple places.

Also, since we need to atomically swap the `DBConfigPtr` upon config update, does that play a role here?

common/dbconfig.cpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+DBConfigPtr DBConfigManager::parseConfig(const Json::Value& root) {
+  static const std::string ACK_MODE = "ack_mode";
+  static const std::string DATASET = "dataset";
+
+  DBConfig newDBConfig;
+
+  if (root.isMember(DATASET)) {
```
newpoo, line 59: How about we reorganize the code like the following?

```
if (!root.isMember(DATASET)) {
 return nullptr;
}

// ...
```

common/dbconfig.cpp
newpoo: A shared_ptr can be constructed from a unique_ptr. So we can still use shared_ptr for dbConfig_.

Semantically parseConfig() is producing a single copy of object, which is not being shared by anything that parseConfig() is aware of. So it's up to parseConfig()'s client to determine if they want to using the returned object via a unique_ptr (using it directly) or a shared_ptr (assign it to a shared_ptr dbConfig_).

newpoo (commented):
left one more comment, otherwise lgtm.

Jia, feel free to take a final pass.

common/dbconfig.cpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  return loadJsonObject(root);
+}
+
+bool DBConfigManager::loadJsonObject(const Json::Value& root) {
+  try {
+    auto newConfig = parseConfig(root);
+    if (newConfig == nullptr) {
+      return false;
```
jaricftw, line 40: nit: when there is no config available (e.g. the config file is not available yet), this mean we'll return `false`. Can you clarify this in the API comments, otherwise it may be a little confusing to the reader, who wonder if there is difference between:
- config is not available, or empty
- config is failed to load/parse

jaricftw (approved):
overall LGTM, after some final comments, thanks!

common/dbconfig.cpp
premkumr: sounds good, Jia

common/dbconfig.cpp
newpoo: @premkumr can you please address this comment?

It makes the code easier to read and maintain.
Also, as in its current form, it can't differentiate between an invalid config and a valid config with empty config content (use the default settings). In another word, if someone wants to change to use the default settings by submitting an empty config, the new config will be rejected.

premkumr: My bad on missing this comment. Yes it cannot differentiate between an empty config and and invalid config. I wanted to ignore an empty config. missing or invalid will be caught by exception and ignored. .. If we want to accept an empty config, effectively '{}' , I can change it in PR:565 .  Let me know @newpoo ...

common/dbconfig.cpp
premkumr: I've made this change in PR 565 ..

PR #556, opened 2021-11-19
thrift_router: refactor and simplify hosts filtering and ranking

jaricftw:
Motivated by reviewing https://github.com/pinterest/rocksplicator/pull/554, where our hosts selection logic in `thrift_router` is bloated and causing lots of confusions, e.g. `RankHostsByGroupPrefixLengthAndShrinkTo` is called 3 times in a single function. https://github.com/pinterest/rocksplicator/pull/554 had to check AZ in multiple places -- error prone and make things even more un-readable. 

This PR attempts to simplify the host selection logic, by separating it into two phases in sequential: (1) filtering, and (2) sorting. 

There shouldn't be any perf impact because we still do two pass of the host list: one for filtering by role & AZ, and another for sorting the list. 

No production logic change, and passing existing test should suffice.

### DIFF
--- a/common/thrift_router.h
+++ b/common/thrift_router.h
@@ -419,7 +419,7 @@ class ThriftRouter {
       const Role role,
       const std::string& segment,
       const int shrink_target) {
-        auto comparator = [this, hostToRole, role, &segment, rotation_counter]
+        auto comparator = [this, &hostToRole, role, &segment, rotation_counter]
         (const Host* h1, const Host* h2) -> bool {
         // prefer master to slave
         if (role == Role::ANY && !FLAGS_always_prefer_local_host) {
--- a/common/thrift_router.h
+++ b/common/thrift_router.h
@@ -304,7 +304,7 @@ class ThriftRouter {
             FLAGS_thrift_router_max_num_hosts_to_consider < shrink_target) {
           shrink_target = FLAGS_thrift_router_max_num_hosts_to_consider;
         }
-        auto hosts_for_shard = selectHostsForShard(
+        auto hosts_for_shard = selectHosts(
           shard_to_hosts[shard], role, rotation_counter, segment, shrink_target, specific_az);
         if (hosts_for_shard.empty()) {
           LOG_EVERY_N(ERROR, FLAGS_thrift_router_log_frequency)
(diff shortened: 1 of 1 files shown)


common/thrift_router.h
```diff
@@ -368,69 +368,70 @@ class ThriftRouter {
...
-        const int shrink_target) {
-      auto comparator = [this, &segment, rotation_counter]
+    void sortAndShrinkHosts(
+      std::vector<const Host*>* v,
+      const std::unordered_map<const Host*, Role>& hostToRole,
+      const unsigned rotation_counter,
+      const Role role,
+      const std::string& segment,
+      const int shrink_target) {
+        auto comparator = [this, hostToRole, role, &segment, rotation_counter]
```
rajathprasad, line 422: Can we pass `hostToRole` by reference? Will avoid a copy.

rajathprasad (commented):
Reviewed this in detail, and it looks good. What is the test plan for this? Since this is a crucial piece of the the routing framework, we would want to ensure there are no bugs.

common/thrift_router.h
jaricftw: done

jaricftw:
> Reviewed this in detail, and it looks good. What is the test plan for this? Since this is a crucial piece of the the routing framework, we would want to ensure there are no bugs.

it's a pure refactor so existing tests should cover it (locally, I intentionally injected a few bugs to make sure unit test work as expected and they did). That said, I can double check to see if we miss any test cases. 

We could also test with a private build ---or we could also do that when we bump the rocksplicator hash in prod repo

rajathprasad:
> > Reviewed this in detail, and it looks good. What is the test plan for this? Since this is a crucial piece of the the routing framework, we would want to ensure there are no bugs.
> 
> it's a pure refactor so existing tests should cover it (locally, I intentionally injected a few bugs to make sure unit test work as expected and they did). That said, I can double check to see if we miss any test cases.
> 
> We could also test with a private build ---or we could also do that when we bump the rocksplicator hash in prod repo

Didn't realize we already have unit tests, gives more confidence :) 
If possible, lets still try to try it with a private build. Or be mindful when we roll it to prod.

rajathprasad (approved):

monilsanghavi (approved):

jaricftw:
> > > Reviewed this in detail, and it looks good. What is the test plan for this? Since this is a crucial piece of the the routing framework, we would want to ensure there are no bugs.
> > 
> > 
> > it's a pure refactor so existing tests should cover it (locally, I intentionally injected a few bugs to make sure unit test work as expected and they did). That said, I can double check to see if we miss any test cases.
> > We could also test with a private build ---or we could also do that when we bump the rocksplicator hash in prod repo
> 
> Didn't realize we already have unit tests, gives more confidence :) If possible, lets still try to try it with a private build. Or be mindful when we roll it to prod.

yeah,  also just deployed a test build based on the new hash (https://phabricator.pinadmin.com/D816082) into pininfoservice test stage, monitored all metrics (SR, latency, QPS to other service) and didn't observe any changes. 

Will also monitor carefully when we release this in production

PR #563, opened 2021-12-10
[common/thrift_client_pool] close connections that are in bad state t…

newpoo:
…o avoid accumulating CLOSE_WAIT

### DIFF
--- a/common/thrift_client_pool.h
+++ b/common/thrift_client_pool.h
@@ -307,24 +307,30 @@ class ThriftClientPool {
     }
 
     void cleanupStaleChannels(const folly::SocketAddress& addr) {
-      // cleanup stale entries if it hasn't been done for a period of time.
       auto now = time(nullptr);
-      if (last_cleanup_time_ + FLAGS_channel_cleanup_min_interval_seconds
-          < now) {
-        last_cleanup_time_ = now;
-        auto itor = channels_.find(addr);
-        int n = 0;
-        while (itor != channels_.end() &&
-               n++ < FLAGS_channel_max_checking_size) {
-          // We don't cleanup !good() live channels here. Otherwise we
-          // will need to upgrade it to a shared_ptr. We expect clients
-          // won't keep a !good() channels for a long period of time.
-          if (itor->second.first.use_count() == 0) {
-            itor = channels_.erase(itor);
-          } else {
-            ++itor;
-          }
-        }
+      // skip cleanup if it was done recently
+      if (now < last_cleanup_time_ + FLAGS_channel_cleanup_min_interval_seconds) {
+	return;
+      }
+
+      last_cleanup_time_ = now;
+      auto itor = channels_.find(addr);
+      int n = 0;
+      while (itor != channels_.end() &&
+	     n++ < FLAGS_channel_max_checking_size) {
+	auto c = itor->second.first.lock();
+	// the channel has been released
+	if (!c) {
+	  itor = channels_.erase(itor);
+	  continue;
+	}
+
+	// the channel is bad
+	if (!itor->second.second->is_good.load()) {
+	  c->closeNow(); // close the connection to avoid accumulating CLOSE_WAIT
+	}
+
+	++itor;
       }
     }
 


common/thrift_client_pool.h
```diff
@@ -307,24 +307,30 @@ class ThriftClientPool {
...
+      if (now < last_cleanup_time_ + FLAGS_channel_cleanup_min_interval_seconds) {
+	return;
+      }
+
+      last_cleanup_time_ = now;
+      auto itor = channels_.find(addr);
+      int n = 0;
+      while (itor != channels_.end() &&
+	     n++ < FLAGS_channel_max_checking_size) {
+	auto c = itor->second.first.lock();
```
rajathprasad, line 321: Is the lock here to ensure that the weak ptr is not released while it's status is checked in the next few lines?

rajathprasad: Oh lock() converts the weak pointer to a shared pointer: https://en.cppreference.com/w/cpp/memory/weak_ptr/lock
And we need to convert to a shared pointer to be able to access the underlying object.

common/thrift_client_pool.h
```diff
@@ -307,24 +307,30 @@ class ThriftClientPool {
...
+	// the channel has been released
+	if (!c) {
+	  itor = channels_.erase(itor);
+	  continue;
+	}
+
+	// the channel is bad
+	if (!itor->second.second->is_good.load()) {
+	  c->closeNow(); // close the connection to avoid accumulating CLOSE_WAIT
```
rajathprasad, line 330: Is it not necessary to remove the channel from the map in this case?

newpoo: I wanted to minimize behavior change to reduce the risk.

If we remove it here, then we will more aggressively establish new connections to the the destination.

When the connection was released (weak_ptr.lock() fails), we know that it's been a while since the connection was established.

When the connection is not released but !good() (here), we are not sure about that.

rajathprasad: Sounds good, thanks for the explanation.

rajathprasad (approved):

common/thrift_client_pool.h
jaricftw: this creates a shared_ptr every time, for `FLAGS_channel_max_checking_size` number of times, is that indented?

common/thrift_client_pool.h
```diff
@@ -307,24 +307,30 @@ class ThriftClientPool {
...
+	return;
+      }
+
+      last_cleanup_time_ = now;
+      auto itor = channels_.find(addr);
+      int n = 0;
+      while (itor != channels_.end() &&
+	     n++ < FLAGS_channel_max_checking_size) {
+	auto c = itor->second.first.lock();
+	// the channel has been released
```
jaricftw, line 322: Q: under what circumstances will the channel be released? (or put it another way, clarify what "release" means here, which is less clear than "closed")

jaricftw:
> …o avoid accumulating CLOSE_WAIT

initial summary seems truncated

jaricftw:
oh PR title is truncated as well

common/thrift_client_pool.h
newpoo: Yes, it's intended. itor will be advanced either via erase() or ++ below. So it creates a shared_ptr for different channel each time.

common/thrift_client_pool.h
newpoo: Released means all shared_ptrs that are associated with the weak_ptr has been destructed. So the underlying channel object has been destructed and its connection has been closed.

There are two layers: thrift_client_pool and thrift_router. 

In the initial design of thrift_client_pool, the assumption was that its client (thrift_router, etc) will release (destruct) the returned client shared_ptr object shortly after the connection becomes bad (such as closed by the remote peer). This assumption turned out to be not always true. thrift_router stores the shared_ptr client objects from thrift_client_pool in a thread local data structure. Though active threads will release (destruct) the shard_ptr objects shortly after a connection becomes bad, idle thread won't. (Some services have hundreds of worker threads, and they are scheduled in a LIFO way, i.e., the worker pool always tries to use the idle threads that were most recently used).

jaricftw: thanks for the detailed explanation!

common/thrift_client_pool.h
jaricftw: oh I misread -- I thought this function only cleans the channel for the specified `addr`. It seems it will clean-up up to `FLAGS_channel_max_checking_size` channels? in the common case I think that may be wasted work assuming connections are healthy?

Not introduce by this PR though, so just to understand the original motivation.

jaricftw (approved):

common/thrift_client_pool.h
newpoo: If a connection is health, the client (thrift_router, etc) of thrift_client_pool will cache the returned shared_ptr so that it can be reused.  In this case, the weak_ptr will be able to lock() into a shared_ptr, so this function won't do anything to it.

jaricftw: also, created https://github.com/pinterest/rocksplicator/pull/565 as an alternative to this approach, i.e. since we already know when a channel is bad, we can close it right before creating a new one for a destination address, wdyt?

newpoo: Feel free to try it out. I can abandon this one if #565 works.
One subtle difference is that #565 only close connections that are being requested. This PR will clean up bad connections to any destinations which may not be requested.

PR #564, opened 2021-12-10
Protect config around SharedMutex

premkumr:
There were too many concerns about avoiding rw-lock around the config
var. The primary reason it was avoided was performance. We will roll
this back if this hits perf issues. [folly/SharedMutex ](https://github.com/facebook/folly/blob/main/folly/SharedMutex.h)seems to be quite performant

### DIFF
--- a/common/dbconfig.cpp
+++ b/common/dbconfig.cpp
@@ -41,8 +41,12 @@ bool DBConfigManager::loadJsonObject(const Json::Value& root) {
     }
 
     LOG(INFO) << "Successfully updated config";
-    std::atomic_exchange_explicit(&dbConfig_, newConfig, std::memory_order_release);
-    fDataLoaded_.store(true);
+    {
+      folly::SharedMutex::WriteHolder holder(lock);
+      dbConfig_.reset();
+      dbConfig_.swap(newConfig);
+      fDataLoaded_.store(true);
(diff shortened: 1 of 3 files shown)


common/dbconfig.cpp
```diff
@@ -71,22 +75,22 @@ DBConfigPtr DBConfigManager::parseConfig(const Json::Value& root) {
...
     // no data was loaded..
     return nullptr;
   }
-  return std::make_shared<const DBConfig>(std::move(newDBConfig));
+  return newDBConfig;
 }
 
 /* conf getter functions */
 ConfigPtr DBConfigManager::getConfig(const std::string& dbName) const {
+  folly::SharedMutex::ReadHolder holder(lock);
```
kangnanli, line 91: this `ReadHolder` will only protect till we find the DBConfig for the `dbName`. Then, during the 102: `getReplicationMode` at `return config->replication_mode;`, it could still case the race condition. Should we consider return a copy of the DBConfig?

premkumr: ConfigPtr is a shared_ptr. Once it is returned, it will be valid , either it is from the previous or current config. It will be a valid data.

premkumr:
Abandoning this to split into different PRs.

PR #567, opened 2021-12-13
Differentiate between empty config and invalid config

premkumr:
- When an empty json is specified all setting become default
- If an invalid json is specified , it is ignored


common/dbconfig.cpp
```diff
@@ -76,9 +80,10 @@ DBConfigPtr DBConfigManager::parseConfig(const Json::Value& root) {
     }
   }
   if (newDBConfig.dataSetConfigMap.size() == 0) {
-    // no data was loaded..
-    return nullptr;
+    // if the json file is empty, then all settings
+    // are at default
   }
```
newpoo, line 85: remove this completely?

premkumr: Done ..

newpoo (approved):

PR #570, opened 2022-02-01
remove unused per-request config

jaricftw:
per-request config is no longer in use, as we have switched to use per-dataset config


premkumr:
Thanks for the cleaning this up Jia !! LGTM.

premkumr (approved):

PR #573, opened 2022-02-04
rocksdb_replicator: post max_seq_no_acked_ regardless of replication mode

jaricftw:
We now support dynamically modifying the replication mode for a db, and the way we use the `max_seq_no_acked_` may run into edge case issues:

**Background on how this `max_seq_no_acked_` is used:**
Currently we only updates `max_seq_no_acked_` in two places:
- replication mode = 1 or 2: when the leader receives a replication request from a follower with the latest sequence number it committed
- replication mode = 1: when the leader sent an rocskdb update with a specific sequence  out to the network (without waiting for follower ack)

On the `Write` Path, if replication_mode is 1 or 2, the leader which received the write will wait for `max_seq_no_acked_` to catch up to it's own local latest sequence, before returning the response to the client. It doesn't wait if replication_mode is 0.

**Possible issues**
As we can see, `max_seq_no_acked_` is not updated when replication_mode == 0. This mean the following can happen:
- db is operating with replication_mode == 2
- a write request arrives and wait for `max_seq_no_acked_` to catch up
- db switched to replication_mode == 0, and as a result, `max_seq_no_acked_` is not updated even when followers ACKed the leader that they have received the write.
- the write request times out after **5s** , during which all other requests may be blocked and may cause cascading failures/timeouts. 

**Proposed solution**
This PR proposes a simple fix: always update `max_seq_no_acked_` regardless of the replication mode, i.e. even when replication_mode == 0, `max_seq_no_acked_` will have the largest sequence number ACKed by followers. This fixes the above race condition when switching from replication mode 2 to 0.  There is no impact to stable states:
- in replication_mode 1 & 2, there is no logic change with this PR
- in replication_mode 0, only change is we update the `max_seq_no_acked_` upon follower ACK, but since we don't wait for `max_seq_no_acked_` anyway in the Write path, it's a no-op.

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -319,15 +319,15 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
     logMetric(kReplicatorLeaderSequenceNumbersBehind, seq_no - leaderSeqNum, db ->db_name_);
   }
 
+  // post the largest sequence number the Slave has committed
+  max_seq_no_acked_.post(seq_no);
+
   auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
   // for now we have to support both till all clusters are migrated
   if (FLAGS_replicator_replication_mode > replication_mode) {
     replication_mode = FLAGS_replicator_replication_mode;
   }
-  if (replication_mode == 1 || replication_mode == 2) {
-    // post the largest sequence number the Slave has committed
-    max_seq_no_acked_.post(seq_no);
-  }
+
   auto timeout = request->max_wait_ms;
 
   cond_var_.runIfConditionOrWaitForNotify(


kangnanli:
How do we inform the service that "db switched to replication_mode == 0,"? 
- if we inform all service at the same time, then, another option is, we force all service to drop all the wait threads since we are going to switch to "repliation_mode 0" anyway, we won't care of dropping those waits. 
- but, if rollout this "db switched to replication_mode == 0,", host by host, ie. deploy a cluster, then, we will run into the case you described above: dbA is still running replication_mode(1/2) and wait for `max_seq_no_acked_`'s updating, which dbB is newly running replication_mode(0), which does not send back update to dbA. Then, dbA will have to wait till timeout and block others.

jaricftw:
> How do we inform the service that "db switched to replication_mode == 0,"?
> 
> * if we inform all service at the same time, then, another option is, we force all service to drop all the wait threads since we are going to switch to "repliation_mode 0" anyway, we won't care of dropping those waits.
> * but, if rollout this "db switched to replication_mode == 0,", host by host, ie. deploy a cluster, then, we will run into the case you described above: dbA is still running replication_mode(1/2) and wait for `max_seq_no_acked_`'s updating, which dbB is newly running replication_mode(0), which does not send back update to dbA. Then, dbA will have to wait till timeout and block others.

Configs are dynamically updated via DBConfigManager https://github.com/pinterest/rocksplicator/blob/master/common/dbconfig.h#L50, and there is no way we'll inform all service instances all at the same time (in our case it'll be via managed data). 

The main issue here is that we hav two paths:
- the `Write` path where a write will wait for `max_seq_no_acked_` to be updated by follower replication request 
- the `handleReplicateRequest` path which updates `max_seq_no_acked_`. 
During the wait in the `Write` path, there could be service config change that switched replication mode from 2 to 0, and hence `max_seq_no_acked_` won't be updated.

kangnanli (approved):
At "replication mode 0", no threads will be waiting as described in the PR summary, thus, this post will wakeup "nobody", thus, won't add any overhead. 

lgtm

PR #574, opened 2022-02-04
cluster_management/LeaderFollowerStateModelFactory: refactor replica rebuild logic

jaricftw:
the rebuild condition checks expanded over the time, see
- initial check only based on WAL TTL and replica liveness: https://github.com/pinterest/rocksplicator/pull/225
- added check for sequence number: https://github.com/pinterest/rocksplicator/pull/236
- expansion to check DB meta: https://github.com/pinterest/rocksplicator/pull/484
And now it becomes much complicated and hard to reason about. 

Refactor the whole logic into a separate function `needRebuildDB` that hopefully should make the logic clearer.

No production impact, pure refactor.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -588,9 +588,8 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
 
     // rebuild the replica if there is at least one live replica to rebuild from AND it meets either of TWO conditions:
     // Case 1. DB local meta is different from upstream
-    // Case 2. DB local meta is the same as upstream, but:
-    //   - latest local data update is too old (WAL TTL expired)
-    //   - sequence number is not the same as upstream (otherwise no need to rebuild)
+    // Case 2. DB local meta is the same as upstream, but latest local data update is too old (WAL TTL expired).
+    //         However, skip rebuild if sequence number is the same as upstream (nothing to rebuild from).
     private boolean needRebuildDB(String dbName, CheckDBResponse upstreamStatus, CheckDBResponse localStatus, Map<String, String> liveHostAndRole, String upstreamHost, int upstreamPort) {
       if (liveHostAndRole.isEmpty()) {
         LOG.error("No other live replicas, skip rebuild " + dbName);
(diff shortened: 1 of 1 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -607,6 +586,46 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
       }
     }
 
+    // rebuild the replica if there is at least one live replica to rebuild from AND it meets either of TWO conditions:
+    // Case 1. DB local meta is different from upstream
+    // Case 2. DB local meta is the same as upstream, but:
+    //   - latest local data update is too old (WAL TTL expired)
+    //   - sequence number is not the same as upstream (otherwise no need to rebuild)
```
kangnanli, line 593: In case2, it is actually not very accurate to say "what condition we will rebuild", it is actually "what condition we will skip rebuild". 
- for those conditions, we will skip, but. Logically, this equals to "only rebuild when both are true". But, it is actually quite hard to understand why we only rebuild when both are true. 

I think the wording of "case2" are confusing.

jaricftw: updated, let me know if it's better

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -607,6 +586,45 @@ public void onBecomeFollowerFromOffline(Message message, NotificationContext con
       }
     }
 
+    // rebuild the replica if there is at least one live replica to rebuild from AND it meets either of TWO conditions:
+    // Case 1. DB local meta is different from upstream
+    // Case 2. DB local meta is the same as upstream, but latest local data update is too old (WAL TTL expired).
+    //         However, skip rebuild if sequence number is the same as upstream (nothing to rebuild from).
```
kangnanli, line 592: could we re-phrase to (I think, use AND or OR is not very easy in our case since we are mixing "AND OR" in our conditions):
We have following conditions to "must rebuild" and "can skip":
"must rebuild" (if either is true):
- local meta is different from upstream
"can skip"
- WAL's latest update timestamp is still fairly new, it could still possible get all the latest data from upstream's WAL via replication
- local seqNo is same as upstream

jaricftw: my intention here is not to mix AND/OR, but to ONLY state when a rebuild is necessary. Hence the two cases. The second case is a little nuanced but hopefully should be clear enough with the update. 

I got confused on your latest comment, since "must rebuild" and "can skip" means a reader will still need to figure out when a rebuild is needed. we should avoid vague messages like "can skip", imho.

kangnanli: I think the "however" part is still confusing, since it is another exception inside the *Case2*. 
I can't think of a better rephrase now, and current change is at least better than before, thus, r+.

kangnanli (approved):

PR #575, opened 2022-02-09
rocksdb_replicator: add Write instrumentation 

jaricftw:
This PR adds the following metrics for instrumenting the `Write` path:

Success & failure counters:
- `kReplicatorWriteSuccess`
- `kReplicatorWriteLeaderFailure`
- `kReplicatorWriteWaitTimedOut`

Latency metrics:
- `kReplicatorWriteToLeaderMs` (just a rename from existing `kReplicatorTimedOut`)
- `kReplicatorWriteSuccessResponseTime`
- `kReplicatorWriteFailureResponseTime`

This will help us track the internal write behavior in rocksplicator (instead of just relying on client side metrics)

Pure metrics addition/change (and small refactor as part of that), no other production logic change

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -85,49 +85,61 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
     throw ReturnCode::WRITE_TO_SLAVE;
   }
 
+  auto write_begin = GetCurrentTimeMs();
+
   incCounter(kReplicatorWriteBytes, updates->GetDataSize(), db_name_);
 
   auto ms = GetCurrentTimeMs();
   updates->PutLogData(rocksdb::Slice(reinterpret_cast<const char*>(&ms),
                                      sizeof(ms)));
-  auto start = GetCurrentTimeMs();
+  auto write_leader_begin = GetCurrentTimeMs();
   auto status = db_wrapper_->WriteToLeader(options, updates);
-  auto end = GetCurrentTimeMs();
-  logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
-  
+  auto write_leader_end = GetCurrentTimeMs();
+  auto write_leader_time = write_leader_begin < write_leader_end ? write_leader_end - write_leader_begin : 0;
+  logMetric(kReplicatorWriteToLeaderMs, write_leader_time, db_name_);
+
+  if (!status.ok()) {
+    incCounter(kReplicatorWriteLeaderFailure, 1, db_name_);
+    auto write_failure_end = GetCurrentTimeMs();
+    logMetric(kReplicatorWriteFailureResponseTime, write_begin < write_failure_end? write_failure_end - write_begin:0, db_name_);
+    return status;
+  }
+
   auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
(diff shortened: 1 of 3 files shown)


rocksdb_replicator/replicated_db.cpp
```diff
@@ -85,49 +85,61 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
...
-  logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
-  
+  auto write_leader_end = GetCurrentTimeMs();
+  auto write_leader_time = write_leader_begin < write_leader_end ? write_leader_end - write_leader_begin : 0;
+  logMetric(kReplicatorWriteToLeaderMs, write_leader_time, db_name_);
+
+  if (!status.ok()) {
+    incCounter(kReplicatorWriteLeaderFailure, 1, db_name_);
+    auto write_failure_end = GetCurrentTimeMs();
+    logMetric(kReplicatorWriteFailureResponseTime, write_begin < write_failure_end? write_failure_end - write_begin:0, db_name_);
```
rajathprasad, line 104: Nit: space before and after the colon. 
`write_begin : 0`

rocksdb_replicator/replicated_db.cpp
```diff
@@ -85,49 +85,61 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
...
+      LOG(ERROR) << "Failed to receive ack from follower, timing out for " << db_name_;
+      return rocksdb::Status::TimedOut("Failed to receive ack from follower");
     }
+    break;
+  default:
+    CHECK(replication_mode == 0)
+      << "Invalid replicaton mode " << replication_mode;
   }
+  auto write_success_end = GetCurrentTimeMs();
+  logMetric(kReplicatorWriteSuccessResponseTime, write_begin < write_success_end? write_success_end - write_begin:0, db_name_);
```
rajathprasad, line 141: Same nit here.

rocksdb_replicator/replicated_db.cpp
```diff
@@ -85,49 +85,61 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
...
   auto status = db_wrapper_->WriteToLeader(options, updates);
-  auto end = GetCurrentTimeMs();
-  logMetric(kReplicatorWriteMs, start < end ? end - start : 0, db_name_);
-  
+  auto write_leader_end = GetCurrentTimeMs();
+  auto write_leader_time = write_leader_begin < write_leader_end ? write_leader_end - write_leader_begin : 0;
+  logMetric(kReplicatorWriteToLeaderMs, write_leader_time, db_name_);
+
+  if (!status.ok()) {
+    incCounter(kReplicatorWriteLeaderFailure, 1, db_name_);
```
rajathprasad, line 102: I'm fine if you want to keep this, but you can always use the `count` of a metric to get the counter. You don't need a counter _and_ a metric.

rajathprasad (approved):

rocksdb_replicator/replicated_db.cpp
jaricftw: yeah, it took me a while to realize this when looking at our rockstore metrics. I still prefer separate counters to be explicit, and consistent with other success/failure counters. Otherwise the naming of the metrics could confuse people. Plus there is little perf concern here given this is in the failure path.

PR #578, opened 2022-02-09
stats: render an index page with all registered endpoints

jaricftw:
This is so we can see all available endpoints in the index page.

Tested via a private build, and see this in `/` page:
```
registered endpoints: 
 /db_info.txt
 /dump_heap
 /gflags.txt
 /pprof/cmdline
 /pprof/heap
 /pprof/profile
 /rocksdb_info.txt
 /server_info.txt
 /service_config
 /stats.txt
```


rajathprasad (approved):
Neat!!

PR #579, opened 2022-02-11
Add introspection for rocksdb replicator 

jaricftw:
Initial PR to support introspection into RocksDB replication states (e.g. DBRole, upstream_address, sequence number, etc), and expose it all the way to `AdminHandler::IntrospectDB` so application can easily hook this up, and render an admin page that shows the replication state for all local shards.

E.g. tested via a private build, and see this:
```
ApplicationDBManager:
audience_targeting-_-visitor_v2_DEV00740:
 ReplicatedDB:
  name: audience_targeting-_-visitor_v2_DEV00740
  DBRole: MASTER
  upstream_addr: unknown_addr
  cur_seq_no: 21368

audience_targeting-_-visitor_v2_DEV00396:
 ReplicatedDB:
  name: audience_targeting-_-visitor_v2_DEV00396
  DBRole: MASTER
  upstream_addr: unknown_addr
  cur_seq_no: 21061

online_migration_e2e_test-_-f200067:
 ReplicatedDB:
  name: online_migration_e2e_test-_-f200067
  DBRole: MASTER
  upstream_addr: unknown_addr
  cur_seq_no: 0

audience_targeting-_-visitor_v2_DEV00076:
 ReplicatedDB:
  name: audience_targeting-_-visitor_v2_DEV00076
  DBRole: SLAVE
  upstream_addr: 10.3.42.199
  cur_seq_no: 20715
```

### DIFF
--- a/common/network_util.cpp
+++ b/common/network_util.cpp
@@ -51,4 +51,14 @@ const std::string& getLocalIPAddress() {
   return local_ip;
 }
 
+std::string getNetworkAddressStr(const folly::SocketAddress& addr) noexcept {
+  std::string add_str = "unknown_addr";
+  try {
+    add_str = addr.getAddressStr();
+  } catch (const std::exception& e) {
+    LOG(ERROR) << "cannot get upstream address: " << e.what();
+  }
+  return add_str;
+}
+
 }  // namespace common
--- a/common/network_util.h
+++ b/common/network_util.h
@@ -16,11 +16,20 @@
 
 #include <string>
 
+#include "folly/SocketAddress.h"
+
+
 namespace common {
 
 /*
  * Get the local IP address from eth0.
  */
 const std::string& getLocalIPAddress();
+
+/**
+ * Get a string representation if the provide address is IPv4 or IPv6,
+ * otherwise "unknown_addr" is returned.
+ */
+std::string getNetworkAddressStr(const folly::SocketAddress&) noexcept;
 
 }  // namespace common
--- a/common/network_util.h
+++ b/common/network_util.h
@@ -27,7 +27,7 @@ namespace common {
 const std::string& getLocalIPAddress();
 
 /**
- * Get a string representation if the provide address is IPv4 or IPv6,
+ * Get a string representation of the provide socket address if it is IPv4 or IPv6,
  * otherwise "unknown_addr" is returned.
  */
 std::string getNetworkAddressStr(const folly::SocketAddress&) noexcept;
--- a/rocksdb_admin/tests/application_db_manager_test.cpp
+++ b/rocksdb_admin/tests/application_db_manager_test.cpp
@@ -41,17 +41,20 @@ TEST(ApplicationDBManagerTest, Basics) {
   admin::ApplicationDBManager db_manager;
   std::string error_message;
   auto ret = db_manager.addDB("test_db", std::move(test_db),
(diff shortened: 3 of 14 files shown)


common/network_util.cpp
```diff
@@ -51,4 +51,14 @@ const std::string& getLocalIPAddress() {
   return local_ip;
 }
 
+std::string getNetworkAddressStr(const folly::SocketAddress& addr) noexcept {
+  std::string add_str = "unknown_addr";
```
rajathprasad, line 55: In your description, I saw a few entries without an upstream address. When would that happen?

rocksdb_admin/tests/application_db_manager_test.cpp
```diff
@@ -41,18 +41,28 @@ TEST(ApplicationDBManagerTest, Basics) {
   admin::ApplicationDBManager db_manager;
   std::string error_message;
   auto ret = db_manager.addDB("test_db", std::move(test_db),
-    replicator::DBRole::SLAVE, &error_message);
+    replicator::DBRole::MASTER, &error_message);
```
jaricftw, line 44: change to `MASTER` because that allows us to test both MASTER and SLAVE DB

common/network_util.cpp
jaricftw: yeah, I noticed that, I it's all for MASTER/Leader replica, the upstream should really be localhost, but i think we don't set that --- I'll look into that separately.

rocksdb_replicator/replicated_db.cpp
```diff
@@ -144,6 +145,28 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
   return status;
 }
 
+std::string RocksDBReplicator::ReplicatedDB::Introspect() {
+  // TODO(jz): extract a common function for DBRole -> string
+  std::string role_string = "__unknown_role__";
+  if (role_ == DBRole::MASTER) {
+    role_string = "MASTER";
+  } else if (role_ == DBRole::SLAVE) {
```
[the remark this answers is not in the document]
jaricftw: sure, will update this, we probably also need to do a change for all other places

common/network_util.h
```diff
@@ -16,11 +16,20 @@
...
 /*
  * Get the local IP address from eth0.
  */
 const std::string& getLocalIPAddress();
 
+/**
+ * Get a string representation of the provide socket address if it is IPv4 or IPv6,
+ * otherwise "unknown_addr" is returned.
+ */
+std::string getNetworkAddressStr(const folly::SocketAddress&) noexcept;
```
kangnanli, line 33: Two questions:
1. could you provide some education links for this c++ syntax, that we ignore the variable name, and only indicate the types. 
2. does the socketAddress get updated somewhere else, which make you return a copy instead of a reference?

jaricftw: 1. it should be perfectly fine to not include a variable name in the function declaration to indicate the types. Just a random search: http://www.cplusplus.com/forum/general/28012/. It's even legal to do this in the actual 
2. this is a utilization function to transform any input SocketAddress into a string, so the original SocketAddress isn't stored anywhere and will be destroyed once the function returns. If we were to return a reference, we would need to  allocate the SocketAddress in the heap (probably with `static`), something we should probably for a temporary variable I think

PR #587, opened 2022-03-02
rocksdb_replicator: reject replicate request when not leader

jaricftw:
Observed in production that we may have followers not receiving updates because it has the wrong leader/upstream address  set. See the following real production example (see https://jira.pinadmin.com/browse/KV-5289 for details):

We have 3 hosts for a shard: hostA, hostB, hostC. 

Helix view:
- hostA: leader
- hostB: follower
- hostC: follower

However, each individual replica thinks differently:
- hostA: leader is itself
- hostB: leader is hostC
- hostC: leader is hostB

As a result, updates are still sent to hostA correctly as the true leader. However, those updates are never replicated to hostB or hostC, because they treat each other as leader, and send replication requests to each other (no-op because they have the same replication sequence number). This is very bad as the followers gets stuck forever (causing inconsistency read, and write timeout when 2-ack mode is on)

A follower only `resetUpstream` when the replication request returns with certain exceptions. 

Therefore, this PR proposes a simple fix: return an exception `NOT_LEADER` for the replication request if the receiving replica is NOT leader. 

For safer rollout/rollback, add a GFLAG to control this behavior. 
Also adding metrics to catch this edge case when it happens.

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -66,7 +66,9 @@ DEFINE_int32(replication_error_reset_upstream_percentage, 10,
 DEFINE_bool(reset_upstream_on_std_exception, false, 
             "Flag to control whether to reset the upstream address for a generic exception in replication");
 DECLARE_int32(rocksdb_replicator_port);
-
+DEFINE_bool(reject_replicate_request_when_not_leader,
+            false,
+            "Flag to control whether to reject a replication request when the receiving replica is not a leader.");
 
 namespace {
 
@@ -287,6 +289,11 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
               incCounter(kReplicatorRemoteApplicationExceptionsNotFound, 1, db->db_name_);
               db->resetUpstream();
             }
+            if (ex.code == ErrorCode::NOT_LEADER) {
+              // request was sent to the wrong leader, reset it
+              incCounter(kReplicatorRemoteApplicationExceptionsNotLeader, 1, db->db_name_);
+              db->resetUpstream();
+            }
           } catch (const std::exception& ex) {
             LOG(ERROR) << "std::exception: " << ex.what();
             incCounter(kReplicatorConnectionErrors, 1, db->db_name_);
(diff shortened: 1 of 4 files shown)


kangnanli:
What conditions might cause these two followers have the wrong `upstream`, causing 
```
hostB: leader is hostC
hostC: leader is hostB
```
?

rajathprasad (changes requested):
I have a feeling this won't work for the following case:
1. When a leader goes down, helix randomly picks another follower to promote to a leader.
2. If this follower is lagging behind the other follower, it will ask for updates from the other follower to catch up before becoming the leader.
3. This new code will block the follower becoming a leader because it is unable to get updates from another follower.

Can you verify if this is the case?

jaricftw:
> What conditions might cause these two followers have the wrong `upstream`, causing
> 
> ```
> hostB: leader is hostC
> hostC: leader is hostB
> ```
> 
> ?

I don't know yet, but this is something I am trying to figure out currently

jaricftw:
> I have a feeling this won't work for the following case:
> 
> 1. When a leader goes down, helix randomly picks another follower to promote to a leader.
> 2. If this follower is lagging behind the other follower, it will ask for updates from the other follower to catch up before becoming the leader.
> 3. This new code will block the follower becoming a leader because it is unable to get updates from another follower.
> 
> Can you verify if this is the case?

I am not sure I understand this:
> If this follower is lagging behind the other follower, it will ask for updates from the other follower to catch up before becoming the leader.

I thought a replica will only ask update from a "leader" (i.e. [pullfromUpstream](https://github.com/pinterest/rocksplicator/blob/dca834f5b0f335a9e0c55b670cd48fd9b284eef1/rocksdb_replicator/replicated_db.cpp#L250)). If it ever asks a follower, it's probably a mistake, not by design?

rajathprasad:
I was referring to the code here: https://github.com/pinterest/rocksplicator/blob/master/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java#L261

> > I have a feeling this won't work for the following case:
> > 
> > 1. When a leader goes down, helix randomly picks another follower to promote to a leader.
> > 2. If this follower is lagging behind the other follower, it will ask for updates from the other follower to catch up before becoming the leader.
> > 3. This new code will block the follower becoming a leader because it is unable to get updates from another follower.
> > 
> > Can you verify if this is the case?
> 
> I am not sure I understand this:
> 
> > If this follower is lagging behind the other follower, it will ask for updates from the other follower to catch up before becoming the leader.
> 
> I thought a replica will only ask update from a "leader" (i.e. [pullfromUpstream](https://github.com/pinterest/rocksplicator/blob/dca834f5b0f335a9e0c55b670cd48fd9b284eef1/rocksdb_replicator/replicated_db.cpp#L250)). If it ever asks a follower, it's probably a mistake, not by design?

jaricftw:
> I was referring to the code here: https://github.com/pinterest/rocksplicator/blob/master/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java#L261

I see, thanks. That looks a little strange, we are changing replica roles that conflicts with Helix's view. Probably that's also why we got into this state at the first place --- divergency of Helix view and replica view. 

In the meantime, I will take a closer look at that.

jaricftw:
replaced by https://github.com/pinterest/rocksplicator/pull/591

PR #591, opened 2022-03-07
rocksdb_replicator: reset upstream when fetching from a non-leader that has no updates

jaricftw:
We see in production where writes can timeout after 2-ack mode is enabled, which usually is caused by followers talking to the wrong upstream (typically a follower), hence never ACKing the write. See this internal [doc](https://docs.google.com/document/d/1xMX3Vkf3fne-usSowSaW9L2deKDDuMEGHLg1IXkxNTo/edit#heading=h.pik5pkd0ubqe) for detailed investigation and plans on fixing those issues. 

This PR serves as the initial step to mitigate the issue: trigger a upstream reset when both of the two conditions below are met:
- no updates from upstream, for a consecutive number of times (controlled by a GFLAG, in case it's transient)
- upstream informs the follower that the upstream is NOT a leader (avoids unnecessary reset when there is no traffic). We introduced the `ReplicaRole` in https://github.com/pinterest/rocksplicator/pull/592

This new behavior is disabled by default (via a GFLAG `reset_upstream_on_empty_updates_from_non_leader`) for a safe rollout, but can be reversed once this feature goes out well in production.

Added unit tests for the two edge cases we see in prod (followers setting each other as upstream; and follower setting itself as upstream).

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -31,7 +31,6 @@
 #include "folly/Random.h"
 #include "rocksdb_replicator/replicator_stats.h"
 #include "rocksdb_replicator/rocksdb_replicator.h"
-#include "rocksdb_replicator/utils.h"
 
 DEFINE_int32(replicator_max_server_wait_time_ms, 10 * 1000,
              "Max wait time before an empty response is returned");
@@ -47,12 +46,13 @@ DEFINE_int32(replicator_pull_delay_on_error_ms, 5 * 1000,
 
 // When there are no updates, it roughly takes
 // replicator_pull_delay_on_error_ms * replicator_max_consecutive_no_updates_before_upstream_reset
-// to trigger a upstream reset. (e.g. with 5s pull delay and 10 consecutive pulls, that's 50s wait in between upstreamReset)
-// Additionally resetUpstream may be skipped as controlled by replication_error_reset_upstream_percentage.
+// to trigger a upstream reset. (e.g. with 5s pull delay and 5 consecutive pulls, that's 25s wait before an upstreamReset)
+// Additionally resetUpstream may be skipped as controlled by replication_error_reset_upstream_percentage (default to 10%),
+// So it may take up to 250s to reset the upstream.
 // This is mainly a protection against checking the upstream too often (currently adds load to ZK).
-// Once we move to checking upstream based on a local shardmap file, 
+// Once we move to checking upstream based on a local shardmap file,
 // we can check more aggressively.
-DEFINE_int32(replicator_max_consecutive_no_updates_before_upstream_reset, 10,
+DEFINE_int32(replicator_max_consecutive_no_updates_before_upstream_reset, 5,
              "The max number of consecutive replication responses with no updates, before a upstream reset is triggered");
 
 DEFINE_int32(replicator_replication_mode, 0,
@@ -336,7 +336,7 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
           } else {
             incCounter(kReplicatorPullRequestsNoUpdates, 1, db->db_name_);
             // no updates consecutively, and the upstream says it's NOT a leader.
-            // Therefore we reset upstream. 
+            // Therefore we reset upstream.
             db->pullFromUpstreamNoUpdates_++;
             if (response.role != ReplicaRole::LEADER
                 && FLAGS_reset_upstream_on_empty_updates_from_non_leader
@@ -443,7 +443,7 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
         }
         if (use_cached_iter || status.ok() || status.IsNotFound()) {
           ReplicateResponse response;
-          response.role = db->role_thrift_;
+          response.role = db->role_;
           uint64_t read_bytes = 0;
           for (int32_t i = 0;
                i < (*request)->max_updates && iter && iter->Valid();
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -31,6 +31,7 @@
 #include "folly/Random.h"
 #include "rocksdb_replicator/replicator_stats.h"
 #include "rocksdb_replicator/rocksdb_replicator.h"
+#include "rocksdb_replicator/utils.h"
 
 DEFINE_int32(replicator_max_server_wait_time_ms, 10 * 1000,
              "Max wait time before an empty response is returned");
@@ -43,6 +44,16 @@ DEFINE_int32(replicator_max_updates_per_response, 50,
 
 DEFINE_int32(replicator_pull_delay_on_error_ms, 5 * 1000,
              "How long to wait before sending the next pull request on error");
+
+// When there are no updates, it roughly takes
+// replicator_pull_delay_on_error_ms * replicator_max_consecutive_no_updates_before_upstream_reset
+// to trigger a upstream reset. (e.g. with 5s pull delay and 10 consecutive pulls, that's 50s wait in between upstreamReset)
+// Additionally resetUpstream may be skipped as controlled by replication_error_reset_upstream_percentage.
+// This is mainly a protection against checking the upstream too often (currently adds load to ZK).
+// Once we move to checking upstream based on a local shardmap file, 
+// we can check more aggressively.
+DEFINE_int32(replicator_max_consecutive_no_updates_before_upstream_reset, 10,
+             "The max number of consecutive replication responses with no updates, before a upstream reset is triggered");
 
 DEFINE_int32(replicator_replication_mode, 0,
              "Replication mode. "
@@ -65,6 +76,8 @@ DEFINE_int32(replication_error_reset_upstream_percentage, 10,
              "what percentage of replication errors should query helix for the latest leader");
 DEFINE_bool(reset_upstream_on_std_exception, false, 
             "Flag to control whether to reset the upstream address for a generic exception in replication");
+DEFINE_bool(reset_upstream_on_empty_updates_from_non_leader, false,
+            "Flag to control whether to reset the upstream address when empty updates are provided from a non-leader upstream");
 DECLARE_int32(rocksdb_replicator_port);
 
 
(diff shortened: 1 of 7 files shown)


rocksdb_replicator/rocksdb_replicator.h
```diff
@@ -136,7 +136,9 @@ class RocksDBReplicator {
     std::shared_ptr<replicator::DbWrapper> db_wrapper_;
     folly::Executor* const executor_;
     const DBRole role_;
+    const ReplicaRole role_thrift_;
```
kangnanli, line 139: Why do we need two *roles*? Won't DBRole is actually the replication role?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -317,7 +331,23 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
...
           if (!response.updates.empty()) {
+            db->pullFromUpstreamNoUpdates_ = 0;
             db->cond_var_.notifyAll();
+          } else {
+            incCounter(kReplicatorPullRequestsNoUpdates, 1, db->db_name_);
+            // no updates consecutively, and the upstream says it's NOT a leader.
+            // Therefore we reset upstream. 
+            db->pullFromUpstreamNoUpdates_++;
+            if (response.role != ReplicaRole::LEADER
+                && FLAGS_reset_upstream_on_empty_updates_from_non_leader
```
kangnanli, line 342: could you elaborate why we need `reset_upstream_on_empty_updates_from_non_leader`? 

when replicated from a nonleader && empty response, shouldn't we reset upstream immediately?

rocksdb_replicator/rocksdb_replicator.h
jaricftw: This is an optimization: we shouldn't need to pass string over the wire on every replication response which can be expensive

rocksdb_replicator/replicated_db.cpp
jaricftw: This FLAG is for a safe rollout so we don't introduce this change immediately when upgrading each cluster. Once we verified this works for a few clusters, perhaps we can enable this feature by default

> when replicated from a nonleader && empty response, shouldn't we reset upstream immediately?
yep, that's the goal state. Today we want to be a little conservative to avoid overloading Helix/ZK. The plan is to reset immediately (PRs will be coming to achieve this goal, and if you are interested, check the doc I linked in the description).

rocksdb_replicator/rocksdb_replicator.h
kangnanli: `DBRole` is *enum*, right? not a string: https://github.com/pinterest/rocksplicator/blob/master/rocksdb_replicator/rocksdb_replicator.h#L69

jaricftw: Sorry, should have made it clearer: we cannot pass `DBRole` directly into the thrift replication response, so we would need to define a thrift enum instead of the `DBRole` enum as you pointed above. 

I thought about using the thrift enum everywhere and get rid of the existing `DBRole` enum, but that will cause a large blast radius of refactor: https://github.com/pinterest/rocksplicator/search?p=3&q=DBRole. Something I can follow up. 

There is also another difference I forgot to callout: `DBRole` has `NOOP` as well which is a validate replication role (e.g. in [BOOTSTRAP](https://github.com/pinterest/rocksplicator/blob/63c9c7d738dc868e59c552a43e2b305314092b0c/cluster_management/src/main/java/com/pinterest/rocksplicator/BootstrapStateModelFactory.java#L129) mode I think), but for the replication flow, only LEADER/FOLLOWER is concerned, so these two enums are not exactly the same.

kangnanli: I think we need unify them, it will cause confusion. 

There is really only 1 role for the db, which is defined by *Replication*. 
The *Noop* just tell the replicationDB that it does not replicate from anyone. https://github.com/pinterest/rocksplicator/blob/63c9c7d738dc868e59c552a43e2b305314092b0c/rocksdb_admin%2Fadmin_handler.cpp#L616-L617

kangnanli: On top of the cpp, at the cluster management layer, we have another *role*, which Helix partition role, such as Helix partition Leader, Folllower, etct, which defined by statemodel.

rocksdb_replicator/thrift/replicator.thrift
```diff
@@ -53,10 +53,18 @@ struct Update {
   3: optional i64 seq_no,
 }
 
+enum ReplicaRole {
+  OTHER = 0,
```
kangnanli, line 57: I think, this should be *NoOP*.

rocksdb_replicator/thrift/replicator.thrift
```diff
@@ -53,10 +53,18 @@ struct Update {
   3: optional i64 seq_no,
 }
 
+enum ReplicaRole {
```
kangnanli, line 56: maybe considering renaming to *ReplicatedRole* to make it explicit.

rocksdb_replicator/rocksdb_replicator.h
kangnanli: Should we consider other names than `role_thrift_`? Maybe, `replica_role_` if the class is `ReplicaRole`

jaricftw: Makes sense, In that case, does my above proposal works for you?
> I thought about using the thrift enum everywhere and get rid of the existing DBRole enum, but that will cause a large blast radius of refactor: https://github.com/pinterest/rocksplicator/search?p=3&q=DBRole. Something I can follow up.

Or do you prefer we do that first before this change. I am okay either way

jaricftw: I will try put up a PR anyway

jaricftw: See https://github.com/pinterest/rocksplicator/pull/592, can you review that first

jaricftw:
@kangnanli or @rajathprasad this is ready for another look

rocksdb_replicator/replicated_db.cpp
```diff
@@ -317,7 +331,23 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
...
 
           if (!response.updates.empty()) {
+            db->pullFromUpstreamNoUpdates_ = 0;
             db->cond_var_.notifyAll();
+          } else {
+            incCounter(kReplicatorPullRequestsNoUpdates, 1, db->db_name_);
+            // no updates consecutively, and the upstream says it's NOT a leader.
+            // Therefore we reset upstream.
+            db->pullFromUpstreamNoUpdates_++;
+            if (response.role != ReplicaRole::LEADER
```
kangnanli, line 341: Do you want to add a counter to record pull from Non-leader for replication?

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
```diff
@@ -348,6 +361,126 @@ TEST(RocksDBReplicatorTest, 1_master_2_slaves_chain) {
   EXPECT_EQ(db_slave_2->GetLatestSequenceNumber(), 2 * n_keys);
 }
 
+TEST(RocksDBReplicatorTest, 1_master_1_slave_upstream_itself) {
```
kangnanli, line 364: I think, for the test name, we still keep the camel case, not `1_master_1_slave_upstream_itself`

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
```diff
@@ -348,6 +361,126 @@ TEST(RocksDBReplicatorTest, 1_master_2_slaves_chain) {
...
+    WriteBatch updates;
+    auto str = to_string(i);
+    updates.Put(str + "key", str + "value");
+    updates.Put(str + "key2", str + "value2");
+    EXPECT_EQ(master.replicator_->write("shard1", options, &updates),
+              ReturnCode::OK);
+    EXPECT_EQ(db_master->GetLatestSequenceNumber(), i * 2 + 2);
+  }
+
+  // expect follower reset upstream to be triggered
```
kangnanli, line 407: when the follower trigger the `resetUpstream`, it seem like the `resetUpstream` will request zookeeper?
Now, at this unit test, there is no zk setup ,right? Then, what is the side effect of that `resetUpstream` call?

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
kangnanli, line 364: Could you add a few lines of comments to explain:
- what is the purpose of test, and how would you plan to test it?
( I can get these info by reading the code, but, I think it might beneficial to briefly explain it)

kangnanli (approved):
I don't have any block reviews. 
Fix before land

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
jaricftw: I am using this style for consistency with all existing tests in this file. If we decide to use CamelCase, then we can change it for all

jaricftw: added

rocksdb_replicator/replicated_db.cpp
jaricftw: good idea, added `kReplicatorPullFromNonLeader`

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
jaricftw: yeah, it'll just be a no-op: https://github.com/pinterest/rocksplicator/blob/bec3645b49c996e3321faac5ab52b9fe04df1a51/rocksdb_replicator/replicated_db.cpp#L231, I have a comment/test below to verify the `resetUpstreamAttempts_` is incremented, but no sequence number change (because reset won't succeed due to this)

jaricftw:
Landing, this won't have any production impact, unless explicitly enabled via the GFLAG.

PR #595, opened 2022-03-10
Replicator: add ReplicaRoleString utils

jaricftw:
Address a TODO to convert ReplicaRole enum to string. Use it in introspection & logging.

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -337,7 +337,7 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
             // no updates consecutively, and the upstream says it's NOT a leader.
             // Therefore we reset upstream.
             db->pullFromUpstreamNoUpdates_++;
-            if (response.role != ReplicaRole::LEADER
+            if (response.__isset.role && response.role != ReplicaRole::LEADER
                 && FLAGS_reset_upstream_on_empty_updates_from_non_leader
                 && db->pullFromUpstreamNoUpdates_ >= FLAGS_replicator_max_consecutive_no_updates_before_upstream_reset) {
               LOG(ERROR) << "No updates when fetching from a non-leader (" + std::string(ReplicaRoleString(response.role))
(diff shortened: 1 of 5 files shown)


rocksdb_replicator/replicated_db.cpp
```diff
@@ -342,10 +337,11 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
             // no updates consecutively, and the upstream says it's NOT a leader.
             // Therefore we reset upstream.
             db->pullFromUpstreamNoUpdates_++;
-            if (response.role != ReplicaRole::LEADER
+            if (response.__isset.role && response.role != ReplicaRole::LEADER
```
kangnanli, line 340: thanks for fixing this. otherwise, it will throw NPE

rocksdb_replicator/rocksdb_replicator.h
```diff
@@ -130,6 +130,7 @@ class RocksDBReplicator {
     std::shared_ptr<replicator::DbWrapper> db_wrapper_;
     folly::Executor* const executor_;
     const ReplicaRole role_;
+    const char* role_str_;
```
kangnanli, line 133: I am not sure, whether this is good practice to include a string version to the class property. 
Since it does not really provide any other info to the class.

rocksdb_replicator/replicated_db.cpp
jaricftw: I don't think it will throw NPE, it'll just be un-initialized, in which case it'll be `NOOP`

rocksdb_replicator/rocksdb_replicator.h
jaricftw: it's pretty common to cache frequent accessed variables, rather than compute on the fly each time its accessed (in this case it's the admin endpoint). And the memory overload of caching a simple short string is very low.

rocksdb_replicator/replicated_db.cpp
kangnanli: OK, it will `un-initialized`, not NPE in c++. if the logic is in Java, if will be NPE.

kangnanli (approved):

PR #597, opened 2022-03-14
rocksdb_replicator: enable leader behind follower stats

jaricftw:
Noticed in prod that we may see leader behind follower sequence, so I wonder if we have metrics for it. Turns out we do, but never enabled it. 

Removing the GFLAG since I don't think it's necessary, it's useful to always have this visibility to catch issues: an upstream/leader received a replication request, but turns out its own local sequence number is lower than requested. 

Test Plan: deployed a private build and don't see any such metrics emitted, as expected. I don't think this case is expected to happen in production, but if they do, it'll be a good chance to debug why

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -68,9 +68,6 @@ DEFINE_uint64(replicator_timeout_ms, 5 * 1000,
               " waiting forever");
 
 DECLARE_int32(replicator_idle_iter_timeout_ms);
-DEFINE_bool(emit_stat_for_leader_behind,
-            false,
-            "Flag to control whether to emit a stat when the leader is behind the follower during a sync request.");
 DEFINE_string(replicator_zk_cluster, "", "Zookeeper cluster");
 DEFINE_string(replicator_helix_cluster, "", "Helix cluster");
 DEFINE_int32(replication_error_reset_upstream_percentage, 10,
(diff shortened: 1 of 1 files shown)


rocksdb_replicator/replicated_db.cpp
```diff
@@ -386,7 +383,7 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
 
   // Inverse of predicate below: if requested sequence number is HIGHER than latest sequence number on leader, emit a stat)
   auto leaderSeqNum = db->db_wrapper_->LatestSequenceNumber();
-  if (FLAGS_emit_stat_for_leader_behind && leaderSeqNum < seq_no) {
```
kangnanli, line 389: should we completely remove the gflag? Is it a good practice to remove a gflag define?

jaricftw: the GFLAG is only necessary when we want to flexibly control some behavior without changing the code/binary. In this case, we are using it to control a metrics emission, which I think is an overkill because:
- This is a rare case which shouldn't probably happen in production, so we need this metrics by default to catch the issue, in case it ever happens. 
- The overhead of enabling this metrics is cheap

Given the above reasons, if we keep the GFLAG, it should default to `true` instead.  Then additionally, if it should always be true, that indicates the GFLAG isn't necessary at all. Removing unnecessary configs makes code less complicated and more maintainable.

rajathprasad: There might be some environments using this gflag. The process doesn't start up if there are unrecognized arguments, so would be good to keep an eye out during deployments or do a sweep of the existing configurations.

rajathprasad (approved):

rocksdb_replicator/replicated_db.cpp
jaricftw: yep, I did a check of all rockstore & rockstorewidecolumn clusters/stages and don't see any usage of that, but will double check.

jaricftw: https://phabricator.pinadmin.com/P11969

PR #598, opened 2022-03-14
cluster_management: add simple retry for helix external view

jaricftw:
Noticed in prod where state transitions could fail (thus leading to ERROR shards) due to failure to obtain Helix external view, hence getting NPE when accessing the `null` view. 

It's not clear whether the failure was transient, but I think it's a good practice to add some retry when talking to external dependency. (currently configured to have 3 retries, with 1s, 2s, and 3s sleep in between). 

This change affects both `LeaderFollower` and `MasterSlave` state models

I plan to have a a follow-up change on dealing with the `null` view if retry still fails. 

ref: https://jira.pinadmin.com/browse/KV-5161

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -40,6 +40,8 @@ import com.pinterest.rocksdb_admin.thrift.SetDBOptionsRequest;
 import org.apache.curator.RetryLoop;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.helix.model.Message;
+import org.apache.helix.HelixAdmin;
+import org.apache.helix.model.ExternalView;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TBinaryProtocol;
 import org.apache.thrift.transport.TSocket;
@@ -52,6 +54,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.TimeUnit;
 
 public class Utils {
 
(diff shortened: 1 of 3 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -60,6 +63,40 @@
...
+    // retry for maxRetries times with sleep in between
+    final int maxRetries = 3;
+    final int retryWaitSec = 1;
+    try {
+      for (int retryCount = 1; retryCount <= maxRetries; retryCount++) {
+        TimeUnit.SECONDS.sleep(retryWaitSec * retryCount);
+        view = admin.getResourceExternalView(clusterName, resourceName);
+        if (view != null) {
+          return view;
+        }
```
rajathprasad, line 89: Do you want to log that there was an attempt made here to get the external view?

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -60,6 +63,40 @@
...
+      }
+    } catch (InterruptedException e) {
+      LOG.error("Interrupted when getting external view for resource " + resourceName, e);
+      return null;
+    }
+
+    LOG.error("failed to obtain external view for resource {} in cluster {} after {} retries", resourceName, clusterName, maxRetries);
```
rajathprasad, line 96: Super nit: Upper case f in failed

rajathprasad (commented):
@jaricftw can you make the same changes to MasterSlaveStateModelFactory too?

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
jaricftw: good idea, added

jaricftw:
> @jaricftw can you make the same changes to MasterSlaveStateModelFactory too?

was planning on a follow-up, but sure, added in this PR

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -60,6 +63,42 @@
...
+    }
+
+    LOG.error("Failed to get external view for " + resourceName + ", retry ...");
+
+    // retry for maxRetries times with sleep in between
+    final int maxRetries = 3;
+    final int retryWaitSec = 1;
+    try {
+      for (int retryCount = 1; retryCount <= maxRetries; retryCount++) {
+        TimeUnit.SECONDS.sleep(retryWaitSec * retryCount);
```
kangnanli, line 87: Should we do exponential backoff retry?

jaricftw: Updated, though I think it doesn't matter in this case, exponentially backoff is typically only used for retrying for a long period (to avoid retrying too frequently and causing issues)

rajathprasad (approved):
Ship it!

PR #601, opened 2022-03-18
Add new replicator role to distinguish follower vs. observer

iprentic:
For 2-ack mode, we need to differentiate between a follower and observer. Add a new replicator role for this.

### DIFF
--- a/rocksdb_replicator/thrift/replicator.thrift
+++ b/rocksdb_replicator/thrift/replicator.thrift
@@ -36,6 +36,9 @@ struct ReplicateRequest {
   # uppper limit set by client side.
   # A value of 0 means no limit
   4: required i32 max_updates,
+
+  # role is the replica role of the downstream host requesting the updates
+  5: optional ReplicaRole role;
 }
 
 typedef binary (cpp.type = "folly::IOBuf") IOBuf
@@ -56,7 +59,8 @@ struct Update {
 enum ReplicaRole {
   NOOP = 0, // no replication needed
   FOLLOWER = 1,
-  LEADER = 2
+  LEADER = 2,
+  OBSERVER = 3 // replicates but should not be counted as a follower ack
 }
 
 struct ReplicateResponse {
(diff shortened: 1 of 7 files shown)


jaricftw:
I think we also need to change here: https://github.com/pinterest/rocksplicator/blob/bd3842591ece082ca8502bf307b28e277bee5eba/rocksdb_replicator/rocksdb_replicator.cpp#L126 otherwise the observer won't perform any replication request. 

Can you test this in a private build?

rocksdb_replicator/thrift/replicator.thrift
```diff
@@ -56,7 +59,8 @@ struct Update {
 enum ReplicaRole {
   NOOP = 0, // no replication needed
   FOLLOWER = 1,
-  LEADER = 2
+  LEADER = 2,
+  OBSERVER = 3 // replicates but should not be counted as a follower ack
```
jaricftw, line 63: can you update https://github.com/pinterest/rocksplicator/blob/master/rocksdb_replicator/utils.h#L21 as well

iprentic:
> I think we also need to change here:
> 
> https://github.com/pinterest/rocksplicator/blob/bd3842591ece082ca8502bf307b28e277bee5eba/rocksdb_replicator/rocksdb_replicator.cpp#L126
> 
> otherwise the observer won't perform any replication request.
> Can you test this in a private build?

Tested on rocksobserver cluster only: Worked
Tested on rocksobserver cluster and rockstorewidcolumn clusters: worked
Tested on rockstorewidecolumn cluster only: worked

rocksdb_replicator/utils.h
```diff
@@ -24,6 +24,8 @@ const char* ReplicaRoleString(ReplicaRole role) {
         return "NOOP";
     case ReplicaRole::FOLLOWER:
         return "FOLLOWER";
+    case ReplicaRole::OBSERVER:
```
jaricftw, line 27: can you add unit tests for this -- I think we already have a test for this function

jaricftw (approved):
LGTM

PR #602, opened 2022-03-23
Adds more info to S3Util::listObjectsHelper error message

mradwan:
This adds more information to listObjectsHelper while I believe the original task asked for more data (like request id) it is not supported in the current version of AWS API we are using so I tried to extract most information from what we have as exception name and whether it is a retryable error or not.

I tested this by making sure the code path is called and supplying a wrong bucket name to aws to trigger an error.

The error is logged inside admin_handler.cpp: restoreDbFromS3
```
# E0323 17:52:11.712312 21926 admin_handler.cpp:1146] Error happened when fetching files in checkpoint from S3: ListObjectRequest failed with exception_name: NoSuchBucket, error_message: The specified bucket does not exist, should_retry: false under path: meariby_test/rockstor..._data_dummy/1/
```

### DIFF
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -32,6 +32,8 @@
 #include <aws/s3/model/Object.h>
 #include <aws/s3/model/PutObjectRequest.h>
 
+#include <folly/Format.h>
+
 #include <algorithm>
 #include <cstdlib>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


common/s3util.cpp
```diff
@@ -252,7 +254,10 @@ void S3Util::listObjectsHelper(const string& prefix, const string& delimiter,
     }
   } else {
     if (error_message != nullptr) {
-      *error_message = listObjectResult.GetError().GetMessage();
+      *error_message = folly::sformat("ListObjectsRequest failed with ExceptionName: {}, ErrorMessage: {}, ShouldRetry: {}.", 
+        listObjectResult.GetError().GetExceptionName(), 
+        listObjectResult.GetError().GetMessage(),
+        listObjectResult.GetError().ShouldRetry());
```
jaricftw, line 260: Can we get the response code: https://sdk.amazonaws.com/cpp/api/LATEST/class_aws_1_1_client_1_1_a_w_s_error.html#adaf553bee7a092a3fb8a1446c1670622

mradwan: Sounds good Just added it

rajathprasad: Can we also add the request ID? That's what the AWS support team asks for to debug failures.

rajathprasad: Oh I'm sorry, just read the description. You can ignore my comment :)

rajathprasad (approved):

PR #604, opened 2022-03-24
S3 util improve

mradwan:
Two changes in this branch
- Add logging for different errors encountered when using listObjects
- Add support for using ListObjectsV2 instead and gate it behind gflag

### DIFF
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -63,7 +63,7 @@ DEFINE_int32(direct_io_buffer_n_pages, 1,
              "Number of pages we need to set to direct io buffer");
 DEFINE_bool(disable_s3_download_stream_buffer, false,
             "disable the stream buffer used by s3 downloading");
-DEFINE_bool(use_list_objects_v2, false, "use ListObjectsV2 instead of ListObjects in S3Client.");
+DEFINE_bool(use_s3_list_objects_v2, false, "use ListObjectsV2 instead of ListObjects in S3Client.");
 
 
 namespace {
@@ -298,6 +298,7 @@ void S3Util::listObjectsHelper(const string& prefix, const string& delimiter,
         objects->push_back(object.GetKey());
       }
     }
+
     if (listObjectResult.GetResult().GetIsTruncated() &&
             next_marker != nullptr) {
       if (listObjectResult.GetResult().GetNextMarker().empty()) {
@@ -356,7 +357,7 @@ ListObjectsResponseV2 S3Util::listAllObjects(const string& prefix, const string&
   string marker;
   string next_marker;
   do {
-    if(FLAGS_use_list_objects_v2) {
+    if(FLAGS_use_s3_list_objects_v2) {
       listObjectsV2Helper(prefix, delimiter, marker, &objects, &next_marker, &error_message);
     } else {
       listObjectsHelper(prefix, delimiter, marker, &objects, &next_marker, &error_message);
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -290,7 +290,6 @@ void S3Util::listObjectsHelper(const string& prefix, const string& delimiter,
       for (const auto& object : contents) {
         objects->push_back(object.GetPrefix());
       }
-
     } else {
       Aws::Vector<Aws::S3::Model::Object> contents =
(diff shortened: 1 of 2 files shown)


mradwan:
Test plan:
For testing the additional logging I deployed private build on a cluster and observed the stats as I triggered some errors

Screen Shot 2022-03-25 at 10 55 31 AM

For testing the s3 util change I gated it behind flag and enabled it on test cluster, will observe how it goes.
I also did a local test to ensure continuation works same way by enforcing 100 object limit on the request
```
#include "third_party/rocksplicator/rocksplicator/common/s3util.h"
#include <gflags/gflags.h>
#include <iostream>


int main(int argc, char** argv){
    gflags::ParseCommandLineFlags(&argc, &argv, true);
    auto util = common::S3Util::BuildS3Util(50, "pinterest-jackson");
    auto resp = util->listAllObjects("meariby_test/rockstore_data/1/");
    std::cout << resp.Body().objects.size() << std::endl;
    return 0;
}
```

```
root@dev-meariby:/home/meariby/code/cosmos# ./bazel-bin/sandbox/test_main --use_list_objects_v2
Using list objects v2 (new)
Read 100 objects
Using list objects v2 (new)
Read 100 objects
Using list objects v2 (new)
Read 100 objects
Using list objects v2 (new)
Read 100 objects
Using list objects v2 (new)
Read 100 objects
Using list objects v2 (new)
Read 30 objects
530
root@dev-meariby:/home/meariby/code/cosmos# ./bazel-bin/sandbox/test_main
Using List Objects old
Read 100 objects
Using List Objects old
Read 100 objects
Using List Objects old
Read 100 objects
Using List Objects old
Read 100 objects
Using List Objects old
Read 100 objects
Using List Objects old
Read 30 objects
530
```

mradwan:
Note: I did consider doing a refactor to avoid duplicating the entire function but given that this code is only needed temporarily I found it not worth investing in. but rather focus on rolling out and just deleting the old function.

common/s3util.cpp
```diff
@@ -254,6 +310,10 @@ void S3Util::listObjectsHelper(const string& prefix, const string& delimiter,
     }
   } else {
     if (error_message != nullptr) {
+      Stats::get()->Incr(folly::sformat("s3_list_objects_helper_error response_code={} exception_name={} should_retry={}", 
+        static_cast<int32_t>(listObjectResult.GetError().GetResponseCode()),
+        listObjectResult.GetError().GetExceptionName(), 
+        listObjectResult.GetError().ShouldRetry()));
```
kangnanli, line 316: Do you have a sample error log after this change?

kangnanli (approved):
lgtm. @jaricftw, can you also take a look in case I missed anything. 

Fix the cmt before land.

common/s3util.cpp
```diff
@@ -214,6 +217,59 @@ SdkGetObjectResponse S3Util::sdkGetObject(const string& key,
   return getObjectResult;
 }
 
+
+void S3Util::listObjectsV2Helper(const string& prefix, const string& delimiter,
```
jaricftw, line 221: `listObjectsV2Helper` looks like very much like `listObjectsHelper`, is there anything we can do to DRY this up a bit to reuse common logic?

jaricftw:
> Note: I did consider doing a refactor to avoid duplicating the entire function but given that this code is only needed temporarily I found it not worth investing in. but rather focus on rolling out and just deleting the old function.

oh didn't see this when I made the comment. Yeah, I think this is fair, in this case: can we add a comment on ListObjectHelper to indicate this is being deprecated and will be deleted once we move to ListObjectV2?

jaricftw (approved):

common/s3util.cpp
mradwan: @kangnanli this was in my previous pull request https://github.com/pinterest/rocksplicator/pull/602

mradwan: Screen Shot 2022-03-25 at 10 55 31 AM

common/s3util.cpp
mradwan: To be honest the refactor is quite doable with templatizing the request object and the actual list objects call but  but I prefer keeping old function as is and just doing switch over.

PR #605, opened 2022-03-28
rocksdb_replicator: log db & upstream address upon replication exception

jaricftw:
When replication happen due to network errors happen (e.g. upstream is down), we get logs like:
```
E0328 07:18:51.642714 14803 replicated_db.cpp:297] std::exception: Channel is !good()
```
Adding db name and upstream address would help us pinpoint where it failed

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -283,7 +283,7 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
             t.exception().throwException();
 #endif
           } catch (const ReplicateException& ex) {
-            LOG(ERROR) << "ReplicateException: upstream = " << db->upstream_addr_.getAddressStr() << ", code = " << static_cast<int>(ex.code)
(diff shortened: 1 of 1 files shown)


mradwan:
LGTM,

This might raise an exception if the address is neither IPv4 or IPv6 address but seems like we use it extensively already.
https://github.com/facebook/folly/blob/main/folly/SocketAddress.h#L417
Do you think its possible for address is invalid?

jaricftw:
> LGTM,
> 
> This might raise an exception if the address is neither IPv4 or IPv6 address but seems like we use it extensively already. https://github.com/facebook/folly/blob/main/folly/SocketAddress.h#L417 Do you think its possible for address is invalid?

Good question, yeah, I thought about that, and we don't expect that to be invalid (if it's so, it's probably some bugs introduce by us somewhere). 

However, it doesn't hurt to be extra safe, especially I've already worked on this before: https://github.com/pinterest/rocksplicator/blob/master/common/network_util.cpp#L54, let me try to use that

mradwan (approved):
LGTM thanks for addressing it!

PR #607, opened 2022-03-31
cluster_management: include dates in logs for other components

jaricftw:
Similar to https://github.com/pinterest/rocksplicator/pull/589, let's add date in logs for other helix/java components

tested on a private build, e.g. for shardmap-agent:
```
06 Apr 2022 07:56:55.943 [main] ERROR ksplicator.ClientShardMapAgent - ShardMapAgent running
```

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/ClientShardMapAgent.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/ClientShardMapAgent.java
@@ -136,7 +136,7 @@ public class ClientShardMapAgent {
   public static void main(String[] args) throws Exception {
     org.apache.log4j.Logger.getRootLogger().setLevel(Level.INFO);
     BasicConfigurator.configure(new ConsoleAppender(
-        new PatternLayout("%d{HH:mm:ss.SSS} [%t] %-5p %30.30c - %m%n")
+        new PatternLayout("%d{dd MMM yyyy HH:mm:ss.SSS} [%t] %-5p %30.30c - %m%n")
     ));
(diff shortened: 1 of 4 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/ClientShardMapAgent.java
```diff
@@ -136,7 +136,7 @@ private static CommandLine processCommandLineArgs(String[] cliArgs) throws Parse
   public static void main(String[] args) throws Exception {
     org.apache.log4j.Logger.getRootLogger().setLevel(Level.INFO);
     BasicConfigurator.configure(new ConsoleAppender(
-        new PatternLayout("%d{HH:mm:ss.SSS} [%t] %-5p %30.30c - %m%n")
+        new PatternLayout("%d{dd MMM yyyy HH:mm:ss.SSS} [%t] %-5p %30.30c - %m%n")
```
kangnanli, line 139: could you add a sample log in the PR description?

kangnanli (approved):
please fix before land

cluster_management/src/main/java/com/pinterest/rocksplicator/ClientShardMapAgent.java
jaricftw: see https://github.com/pinterest/rocksplicator/pull/589, it would be the same, but I can try to see if we have test stage setup for spectators, etc

jaricftw: tested a private build and updated the summary fro this.

PR #608, opened 2022-04-08
rocksdb_replicator: add a util function ReturnCodeString

jaricftw:
This will be used on the caller side to log/print the return code as string

### DIFF
--- a/rocksdb_replicator/utils.h
+++ b/rocksdb_replicator/utils.h
@@ -15,6 +15,7 @@
 #pragma once
 
 #include "rocksdb_replicator/thrift/gen-cpp2/Replicator.h"
+#include "rocksdb_replicator/rocksdb_replicator.h"
 
 namespace replicator {  // namespace replicator
 
@@ -29,6 +30,26 @@ const char* ReplicaRoleString(ReplicaRole role) {
 
     default:
         return "__unknown_role__";
+  }
+}
+
+const char* ReturnCodeString(ReturnCode code) {
+  switch(code) {
+    case ReturnCode::OK:
+        return "OK";
+    case ReturnCode::DB_NOT_FOUND:
+        return "DB_NOT_FOUND";
+    case ReturnCode::DB_PRE_EXIST:
+        return "DB_PRE_EXIST" ;
+    case ReturnCode::WRITE_TO_SLAVE:
+        return "WRITE_TO_SLAVE" ;
+    case ReturnCode::WRITE_ERROR:
+        return "WRITE_ERROR" ;
+    case ReturnCode::WAIT_SLAVE_TIMEOUT:
+        return "WAIT_SLAVE_TIMEOUT" ;
+
+    default:
+        return "__unknown_code__";
   }
 }
 
--- a/rocksdb_replicator/tests/utils_test.cpp
+++ b/rocksdb_replicator/tests/utils_test.cpp
@@ -27,6 +27,20 @@ TEST(ReplicatorUtilsTest, ReplicaRoleString) {
   }
 }
 
+TEST(ReplicatorUtilsTest, ReturnCodeString) {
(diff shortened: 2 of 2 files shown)


rocksdb_replicator/utils.h
```diff
@@ -32,4 +33,24 @@ const char* ReplicaRoleString(ReplicaRole role) {
   }
 }
 
+const char* ReturnCodeString(ReturnCode code) {
```
kangnanli, line 36: In thrift, they generate a map with following code in Thrift.h
```
template <typename EnumTypeT>
struct TEnumMapFactory {
  using EnumType = EnumTypeT;
  using ValuesToNamesMapType = std::map<EnumType, const char*>;
  using NamesToValuesMapType = std::map<const char*, EnumType, ltstr>;
  using Traits = TEnumTraits<EnumType>;

  static ValuesToNamesMapType makeValuesToNamesMap() {
    ValuesToNamesMapType _return;
    for (size_t i = 0; i < Traits::size; ++i) {
      _return.emplace(EnumType(Traits::values[i]), Traits::names[i].data());
    }
    return _return;
  }
  static NamesToValuesMapType makeNamesToValuesMap() {
    NamesToValuesMapType _return;
    for (size_t i = 0; i < Traits::size; ++i) {
      _return.emplace(Traits::names[i].data(), EnumType(Traits::values[i]));
    }
    return _return;
  }
};
```

jaricftw: Yeah, I looked at many other possible implementations for converting enums to strings: https://stackoverflow.com/questions/207976/how-to-easily-map-c-enums-to-strings, but I don't think any of them is **simple & performant enough.** Ideally what I want is a tool to auto-generate the code that does the translation so we don't have to write any code. We don't want any runtime reflection for perf reasons. Do you have suggestion for one that's available for C++? (e.g. in Go this would be very simple: https://pkg.go.dev/golang.org/x/tools/cmd/stringer)

I am not sure how the above code is used in thrift, is that just something they use to auto-gen an enum to string map?

kangnanli: for generation part, I don't know how thrift get it to work, will have to dive into thrift code to get it. 

but, for the data structure part, do you think we should consider store it a map? It will be easier to visualize and sort of consistent to thrift

mradwan: I am okay with this implementation, but the way I would go about implementing this is using a bit of meta programming where you would have a typetrait for each enum value.

I did a bit of exploration to see if we can do compile time checks but unfortunately all of them would rely on knowing the enum end marker.

another thing you could think about is using macros, these will probably work but they are not the most fun thing to read..

mradwan (approved):

rocksdb_replicator/utils.h
kangnanli: @mradwan could you elaborate on "have a typetrait for each enum value"?

kangnanli: I though we only need `NameTrait<CURRENT>::value`, not for the `Lookup`. 
overall, this seem to be over complicated....

mradwan: Thing is NameTrait<X> means X needs to be known at compile time.

However size_t key or     TestEnum test = TestEnum::A; is a variable in memory thus only known at run time.

So what this code does it will generate code for different structs

Lookup<0>, Lookup<1>, Lookup<2> and then choosing which one to return NameTrait from happen during runtime based on the value in memory

mradwan: @kangnanli I spent sometime thinking how to improve it, and I think the way to go would be extracting everything into library function that can be reused for all enums thus complexity will be there once.

I did use C++20 tho, because it makes error message much better

Here is how you would implement a good enum
```
enum TestEnum {
  A,
  B,
  END_MARKER,
};

```

PR #609, opened 2022-04-11
cluster_management: clean up zk state check logic in leader transition

jaricftw:
Fixing a `FIXME`, where we have dead code in the leader transition code.

The original intention of the logic is to block leader transition when the local sequence number `localSeq` is 
lower than `lastState.seqNum * 0.5`. But the actual implementation has a bug when makes the 
blocking happen only when `localSeq <= 0`. 

We could fix the bug based on the original intention, but that's not necessary: we already have logic above that allows the 
leader to catch up with a follower if it's lagging behind, and if it failed to do so in 10min, it'll thrown an exception. 

Therefore, by the time we check zk state, the leader should already be caught up (if there is live replica to catch up from), hence we can simplify the logic to not having the `tolerance` concept for now. This simplifies the whole logic a bit, so it should be a good starting point before we further tune it. 

No production logic change in this PR. Only code clean up.


rajathprasad (approved):
Makes sense, thanks for the write up.

PR #610, opened 2022-04-12
rocksdb_replicator: adaptive replication timeout for 2-ACK mode

jaricftw:
When 2-ACK Mode is enabled, writes to the leader will need to wait for at least one follower ACK, before returning to the client. The wait timeout is controlled by a GFLAG `replicator_timeout_ms`, which can be generous to accommodate for the p99/p999 network delay (e.g. we typically set it to be 1s in production). However, in the cases when there is replication issues for a db, the leader may not be receiving followers' ACK for a long time, and as a result, all writes to this db would be blocked for 1s, which may end up blocking all worker threads and impacting all other dbs in the process. In production, we see this happens quite often, which leads to load shedding errors in the thrift handler and other cascading failures (e.g. state transitions may fail due to this).

As a mitigation, we introduce adaptive replication timeout, i.e.:
-  We keep the current `GFLAG_replication_timeout_ms` as the normal timeout (e.g. 1s)
- We introduce a new  `GFLAG_replication_timeout_degraded_ms` which can be very small (e.g. 10ms), based on stats: https://statsboard.pinadmin.com/share/j7qev, p99 is usually around 1ms, so requests should typically still succeed in that case.
- For a DB, if we see N (e.g. N=100) **consecutive** timeouts, we know that it's likely there is replication issue and the leader cannot receive follower ACK, so we fall back to the degraded mode, and use `GFLAG_replication_timeout_degraded_ms` instead. The goal is to **fail-fast** so that we can yield resource for other DBs. 
- In degradation mode, as soon as there is **one** successful request (i.e. follower ACK returned with `GFLAG_replication_timeout_degraded_ms`), we immediately **recover** to normal mode with the regular `GFLAG_replication_timeout_ms`.

Also, while we are here, updating the default `replicator_timeout_ms` GFLAG to be 2s based on our production stats that should be sufficient for the worst-case. 

========= Testing Plan ============
Added new unit tests.

Also, a private build with this PR is deployed to a cluster.

[**Degradation**]
Stop 2 replicas for a DB, we see timeouts of requests (about 1s apart). Then the DB enters **degradation** mode:

```
$cat rockstorewidecolumn.log | grep 'use new timeout' -A5 -B5
E0412 04:14:30.128701  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:31.133306  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:32.137323  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:33.141415  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:34.145385  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:34.145534  5267 replicated_db.cpp:162] Enter degradation mode for db jzhan_rw_test00050 after 100 write timeouts, use new timeout to fail fast: 10ms
E0412 04:14:34.160810  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:34.174832  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:34.188071  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:34.201182  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
E0412 04:14:34.214716  5267 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
```

it took 100 consecutive request timeouts to enter degradation mode
```
jzhan@infra-rockstorewidecolumndev-rw-wc-beta-test-dev-0a083df5:/data/xvdb/rockstorewidecolumn$ cat rockstorewidecolumn.log | grep 'use new timeout to fail fast: 10ms' -B 1000 | grep 'Failed to receive ack from follower' | wc -l
100
```

In degradation mode, timeout is 10ms, and we can see from below, average request latency is 13ms (the `adhoc.py` scripts simply issues 1000 requests in sequential)
```
root@dev-jzhan:/mnt/pinboard/tools/rocksplicator/scripts# python adhoc.py
KVStoreGetRowsResponse(rowResponses=[KVStoreGetSingleRowResponse(status=Status(statusCode=0, statusMsg=None), row=Row(key=b'1', items=[Item(column='col', data=None, timestamp=None, itemStatus=Status(statusCode=0, statusMsg=None), values=[CellValue(data=b'11', timestamp=1649737293925)])], columnContinuation=None)), KVStoreGetSingleRowResponse(status=Status(statusCode=0, statusMsg=None), row=Row(key=b'2', items=[Item(column='col', data=None, timestamp=None, itemStatus=Status(statusCode=0, statusMsg=None), values=[CellValue(data=b'2222', timestamp=1649737293926)])], columnContinuation=None))])
13288
```

[**Recovery**] Bring back one replica, and we can see the DB exits the degradation mode:
```
E0412 04:21:33.936885  5265 replicated_db.cpp:152] Failed to receive ack from follower, timing out for jzhan_rw_test00050
I0412 04:22:20.511140  5265 replicated_db.cpp:172] 2-ACK Write succeeded, exit degradation mode and switch to use normal replicator timeout: 1000ms
E0412 04:22:22.457592  5122 file_watcher.cpp:166] File removed or unmounted: /data/xvdb/rocksplicator/zkshardmap/rockstorewidecolumndev-rw-wc-beta-test
```

Now request all go through quickly (4ms on average)
```
root@dev-jzhan:/mnt/pinboard/tools/rocksplicator/scripts# python adhoc.py
KVStoreGetRowsResponse(rowResponses=[KVStoreGetSingleRowResponse(status=Status(statusCode=0, statusMsg=None), row=Row(key=b'1', items=[Item(column='col', data=None, timestamp=None, itemStatus=Status(statusCode=0, statusMsg=None), values=[CellValue(data=b'11', timestamp=1649737877360)])], columnContinuation=None)), KVStoreGetSingleRowResponse(status=Status(statusCode=0, statusMsg=None), row=Row(key=b'2', items=[Item(column='col', data=None, timestamp=None, itemStatus=Status(statusCode=0, statusMsg=None), values=[CellValue(data=b'2222', timestamp=1649737877361)])], columnContinuation=None))])
4539
```

Also see metrics that are added in this PR: https://statsboard.pinadmin.com/share/2sqdr

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -63,15 +63,19 @@ DEFINE_int32(replicator_replication_mode, 0,
              "stack for the connection to one of the Slave; "
              "2: ack client once committed to Master and one of the Slaves.");
 
-DEFINE_uint64(replicator_timeout_ms, 5 * 1000,
+DEFINE_uint64(replicator_timeout_ms, 2 * 1000,
               "How long to wait for Follower ACK before timeout a client write, 0 means"
               " waiting forever");
 
 DEFINE_uint64(replicator_timeout_degraded_ms, 10,
               "In Degradation mode, i.e. when the writes keep timing out waiting for follower ACK, "
               "we would use this timeout to fail fast. "
-              "0 means this mode is disabled. "
-              "This needs to be smaller than replicator_timeout_ms for the config to be valid and used.");
+              "Note that for this config to be valid and used, it needs to be "
+              "smaller than replicator_timeout_ms, but not smaller than 1 (set by kMinReplTimeoutMs)");
+
+DEFINE_uint64(replicator_consecutive_ack_timeout_before_degradation, 100,
+             "The max number of consecutive replication timeout, before a DB enters degradation mode "
+             "and use the above replicator_timeout_degraded_ms");
 
 DECLARE_int32(replicator_idle_iter_timeout_ms);
 DEFINE_string(replicator_zk_cluster, "", "Zookeeper cluster");
@@ -146,32 +150,35 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
     // TODO(bol): This potentially could block all worker threads. We may
     // consider having a dedicated set of worker threads for admin requests,
     // and/or provide async write API when this turns out to be a problem.
-    if (!max_seq_no_acked_.wait(cur_seq_no, current_replicator_timeout_ms_)) {
+    if (!max_seq_no_acked_.wait(cur_seq_no, current_replicator_timeout_ms_.load())) {
       incCounter(kReplicatorWriteWaitTimedOut, 1, db_name_);
       LOG(ERROR) << "Failed to receive ack from follower, timing out for " << db_name_;
       numConsecutiveReplTimeout_ ++;
 
       // enter degradation mode if there has been consecutive timeouts waiting for the follower ACK.
       // This allows us to use a smaller wait timeout to fail fast, instead of blocking for long.
-      if (numConsecutiveReplTimeout_ == kNumReplTimeoutsBeforeDegradation
-          && current_replicator_timeout_ms_ == FLAGS_replicator_timeout_ms
+      if (numConsecutiveReplTimeout_.load() == FLAGS_replicator_consecutive_ack_timeout_before_degradation
+          && current_replicator_timeout_ms_.load() == FLAGS_replicator_timeout_ms
           && FLAGS_replicator_timeout_degraded_ms < FLAGS_replicator_timeout_ms
           && FLAGS_replicator_timeout_degraded_ms >= kMinReplTimeoutMs) {
-        current_replicator_timeout_ms_ = FLAGS_replicator_timeout_degraded_ms;
+        current_replicator_timeout_ms_.store(FLAGS_replicator_timeout_degraded_ms);
         LOG(ERROR) << "Enter degradation mode for db " << db_name_
-                   << " after " << numConsecutiveReplTimeout_ << " write timeouts,"
-                   << " use new timeout to fail fast: " << current_replicator_timeout_ms_ << "ms";
+                   << " after " << numConsecutiveReplTimeout_.load() << " write timeouts,"
+                   << " use new timeout to fail fast: " << current_replicator_timeout_ms_.load() << "ms";
         incCounter(kReplicatorWriteTwoAckDegraded, 1, db_name_);
       }
       return rocksdb::Status::TimedOut("Failed to receive ack from follower");
     }
-    numConsecutiveReplTimeout_ = 0;
-    if (current_replicator_timeout_ms_ != FLAGS_replicator_timeout_ms) {
-      current_replicator_timeout_ms_ = FLAGS_replicator_timeout_ms;
+
+    numConsecutiveReplTimeout_.store(0);
+    // TODO(jz): consider use CAS
+    if (current_replicator_timeout_ms_.load() != FLAGS_replicator_timeout_ms) {
+      current_replicator_timeout_ms_.store(FLAGS_replicator_timeout_ms);
       LOG(ERROR) << "2-ACK Write succeeded, exit degradation mode and switch to use normal replicator timeout: "
-                << current_replicator_timeout_ms_ << "ms";
+                << current_replicator_timeout_ms_.load() << "ms";
       incCounter(kReplicatorWriteTwoAckRecovered, 1, db_name_);
     }
+
     break;
   default:
     CHECK(replication_mode == 0)
@@ -195,6 +202,7 @@ std::string RocksDBReplicator::ReplicatedDB::Introspect() {
   ss << "  ReplicaRole: " << role_str_ << std::endl;
   ss << "  upstream_addr: " << upstream_addr_str << std::endl;
   ss << "  cur_seq_no: " << cur_seq_no << std::endl;
+  ss << "  current_replicator_timeout_ms_: " << current_replicator_timeout_ms_.load() << std::endl;
   // TODO(jz): add max_seq_no_acked_
   return ss.str();
 }
@@ -236,8 +244,9 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
     replicator_helix_cluster_ = FLAGS_replicator_helix_cluster;
   }
 
+  // ensure we don't use a timeout smaller than kMinReplTimeoutMs
   if (FLAGS_replicator_timeout_ms > kMinReplTimeoutMs) {
-    current_replicator_timeout_ms_ = FLAGS_replicator_timeout_ms;
+    current_replicator_timeout_ms_.store(FLAGS_replicator_timeout_ms);
   }
 
   rpc_options_.setTimeout(
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -64,8 +64,14 @@ DEFINE_int32(replicator_replication_mode, 0,
              "2: ack client once committed to Master and one of the Slaves.");
 
 DEFINE_uint64(replicator_timeout_ms, 5 * 1000,
-              "How long to wait for Slave before timeout a client write, 0 means"
+              "How long to wait for Follower ACK before timeout a client write, 0 means"
               " waiting forever");
+
+DEFINE_uint64(replicator_timeout_degraded_ms, 10,
+              "In Degradation mode, i.e. when the writes keep timeout out waiting for follower ACK, "
+              "we would use this small timeout to fail fast. "
+              "0 means this mode is disabled. "
+              "This needs to be smaller than replicator_timeout_ms for the config to be valid and used.");
 
 DECLARE_int32(replicator_idle_iter_timeout_ms);
 DEFINE_string(replicator_zk_cluster, "", "Zookeeper cluster");
@@ -80,6 +86,8 @@ DECLARE_int32(rocksdb_replicator_port);
 
 
 namespace {
+
+const uint32_t kNumReplTimeoutsBeforeDegradation = 100;
 
 uint64_t GetCurrentTimeMs() {
   return std::chrono::duration_cast<std::chrono::milliseconds>(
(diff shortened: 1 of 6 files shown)


rocksdb_replicator/replicated_db.cpp
```diff
@@ -139,11 +146,32 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
...
       LOG(ERROR) << "Failed to receive ack from follower, timing out for " << db_name_;
+      numConsecutiveReplTimeout_ ++;
+
+      // enter degradation mode if there has been consecutive timeouts waiting for the follower ACK.
+      // This allows us to use a smaller wait timeout to fail fast, instead of blocking for long.
+      if (current_replication_timeout_ms_ == FLAGS_replicator_timeout_ms
+          && numConsecutiveReplTimeout_ == kNumReplTimeoutsBeforeDegradation
+          && FLAGS_replicator_timeout_degraded_ms < FLAGS_replicator_timeout_ms
+          && FLAGS_replicator_timeout_degraded_ms >= kMinReplTimeoutMs) {
+        current_replication_timeout_ms_ = FLAGS_replicator_timeout_degraded_ms;
```
rajathprasad, line 160: To clarify, this degraded state is per db (shard), right? If one shard is degraded, the other shards continue to operate as usual without a change in timeouts. Is that accurate?

rocksdb_replicator/rocksdb_replicator.h
```diff
@@ -52,6 +52,9 @@ namespace wangle {
 
 namespace replicator {
 
+const uint32_t kMinReplTimeoutMs = 1;
+const uint32_t kNumReplTimeoutsBeforeDegradation = 100;
```
rajathprasad, line 56: Should we make this configurable using a gflag? For high traffic clusters, I'm wondering if 100 is too high as a threshold?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -208,6 +236,10 @@ RocksDBReplicator::ReplicatedDB::ReplicatedDB(
     replicator_helix_cluster_ = FLAGS_replicator_helix_cluster;
   }
 
+  if (FLAGS_replicator_timeout_ms > kMinReplTimeoutMs) {
+    current_replication_timeout_ms_ = FLAGS_replicator_timeout_ms;
+  }
+
```
rajathprasad, line 242: Why is this code snippet required? (Can you also add a comment explaining the same)?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -64,9 +64,15 @@ DEFINE_int32(replicator_replication_mode, 0,
...
 
 DEFINE_uint64(replicator_timeout_ms, 5 * 1000,
-              "How long to wait for Slave before timeout a client write, 0 means"
+              "How long to wait for Follower ACK before timeout a client write, 0 means"
               " waiting forever");
 
+DEFINE_uint64(replicator_timeout_degraded_ms, 10,
+              "In Degradation mode, i.e. when the writes keep timeout out waiting for follower ACK, "
+              "we would use this small timeout to fail fast. "
+              "0 means this mode is disabled. "
```
rajathprasad, line 73: This is true only when `kMinReplTimeoutMs` is 1, right?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -64,9 +64,15 @@ DEFINE_int32(replicator_replication_mode, 0,
              "2: ack client once committed to Master and one of the Slaves.");
 
 DEFINE_uint64(replicator_timeout_ms, 5 * 1000,
```
rajathprasad, line 66: Should we also reduce the default timeout? 5s feels too high.

rocksdb_replicator/replicated_db.cpp
```diff
@@ -139,11 +146,32 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
     // TODO(bol): This potentially could block all worker threads. We may
     // consider having a dedicated set of worker threads for admin requests,
     // and/or provide async write API when this turns out to be a problem.
-    if (!max_seq_no_acked_.wait(cur_seq_no, FLAGS_replicator_timeout_ms)) {
+    if (!max_seq_no_acked_.wait(cur_seq_no, current_replication_timeout_ms_)) {
       incCounter(kReplicatorWriteWaitTimedOut, 1, db_name_);
       LOG(ERROR) << "Failed to receive ack from follower, timing out for " << db_name_;
+      numConsecutiveReplTimeout_ ++;
```
rajathprasad, line 152: Do these operations need to be thread safe? Multiple threads could be writing to the same DB simultaneously, right?

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
```diff
@@ -485,6 +489,101 @@ TEST(RocksDBReplicatorTest, 1_leader_2_followers_deadlock) {
...
+  Host master(master_port);
+  Host slave(slave_port);
+
+  auto db_master = cleanAndOpenDB("/tmp/db_master");
+  auto db_slave = cleanAndOpenDB("/tmp/db_slave");
+
+  RocksDBReplicator::ReplicatedDB* replicated_db_master;
+  RocksDBReplicator::ReplicatedDB* replicated_db_slave;
+
+  EXPECT_EQ(master.replicator_->addDB("shard1", db_master, ReplicaRole::LEADER, folly::SocketAddress(),
```
rajathprasad, line 511: Would it be possible to add checks to make sure another shard `shard2` remains unaffected?

jaricftw: Good idea, added tests on that.

rocksdb_replicator/replicated_db.cpp
jaricftw: yeah, the timeout `current_replication_timeout_ms_ ` is stored per db, also added tests for this

rocksdb_replicator/replicated_db.cpp
jaricftw: added a comment to explain: basically this is an extra safety check to make sure we don't set a timeout that is too-short

rocksdb_replicator/replicated_db.cpp
jaricftw: Also verified in the introspection page:

image

rocksdb_replicator/replicated_db.cpp
```diff
@@ -63,10 +63,20 @@ DEFINE_int32(replicator_replication_mode, 0,
...
-              "How long to wait for Slave before timeout a client write, 0 means"
+DEFINE_uint64(replicator_timeout_ms, 2 * 1000,
+              "How long to wait for Follower ACK before timeout a client write, 0 means"
               " waiting forever");
 
+DEFINE_uint64(replicator_timeout_degraded_ms, 10,
+              "In Degradation mode, i.e. when the writes keep timing out waiting for follower ACK, "
+              "we would use this timeout to fail fast. "
+              "Note that for this config to be valid and used, it needs to be "
+              "smaller than replicator_timeout_ms, but not smaller than 1 (set by kMinReplTimeoutMs)");
```
rajathprasad, line 74: I'm still a little confused.
If value < 0 then this feature is disabled.
if 0 < value < kMinReplTimeoutMs, then also this feature is disabled, right?
Can we make it simpler by saying that this value should be greater than kMinReplTimeoutMs but lesser than replicator_timeout_ms?

rocksdb_replicator/replicated_db.cpp
```diff
@@ -139,11 +150,35 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
     // TODO(bol): This potentially could block all worker threads. We may
     // consider having a dedicated set of worker threads for admin requests,
     // and/or provide async write API when this turns out to be a problem.
-    if (!max_seq_no_acked_.wait(cur_seq_no, FLAGS_replicator_timeout_ms)) {
+    if (!max_seq_no_acked_.wait(cur_seq_no, current_replicator_timeout_ms_.load())) {
       incCounter(kReplicatorWriteWaitTimedOut, 1, db_name_);
       LOG(ERROR) << "Failed to receive ack from follower, timing out for " << db_name_;
+      numConsecutiveReplTimeout_ ++;
```
rajathprasad, line 156: Nit remove space before ++

rocksdb_replicator/replicated_db.cpp
```diff
@@ -139,11 +150,35 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
...
+
+      // enter degradation mode if there has been consecutive timeouts waiting for the follower ACK.
+      // This allows us to use a smaller wait timeout to fail fast, instead of blocking for long.
+      if (numConsecutiveReplTimeout_.load() == FLAGS_replicator_consecutive_ack_timeout_before_degradation
```
rajathprasad, line 160: Should we make this check to be greater than? Since this is multithreaded, there's a possibility it goes from 99 -> 101 before the check completes

rocksdb_replicator/replicated_db.cpp
```diff
@@ -139,11 +150,35 @@ rocksdb::Status RocksDBReplicator::ReplicatedDB::Write(
...
+          && current_replicator_timeout_ms_.load() == FLAGS_replicator_timeout_ms
+          && FLAGS_replicator_timeout_degraded_ms < FLAGS_replicator_timeout_ms
+          && FLAGS_replicator_timeout_degraded_ms >= kMinReplTimeoutMs) {
+        current_replicator_timeout_ms_.store(FLAGS_replicator_timeout_degraded_ms);
```
rajathprasad, line 164: These statements could be executed multiple times for a db when it is entering the degraded mode (from multiple threads) since you are not enclosing this in a mutex. I thought about it, and I think it should be safe. Just wanted to call it out in case you had other ideas.

rocksdb_replicator/replicated_db.cpp
jaricftw: it's an unsigned integer. reworded a bit, hopefully it's clearer now

rocksdb_replicator/replicated_db.cpp
jaricftw: Good point, updated

rocksdb_replicator/replicated_db.cpp
jaricftw: yeah, thanks for calling this out.  I was reluctant to use a mutex lock mainly for perf reasons. I also think it's okay, if two threads check these conditions at exactly the same time, they may decide to put the db in degradation mode by storing:
```
current_replicator_timeout_ms_.store(FLAGS_replicator_timeout_degraded_ms)
```
And that's idempotent. 
I can add a note for this.

Similarly for the below `current_replicator_timeout_ms_` check, I actually have a TODO for the same reason, but I thought that's just an optimization.

rajathprasad (approved):
Ship it!

jaricftw:
> Ship it!

Thank you. 

FYI, after you approved it, I made a change to move the logic inside the switch statement into a separate helper function for better readability, but no logic change.

PR #613, opened 2022-05-03
LeaderFollowerStateModelFactory: fix NPE when accessing stateMap

jaricftw:
During failure testing (when 3 nodes of a shard is down briefly), we saw error shards with NPE:
```
Caused by: java.lang.NullPointerException
--
at com.pinterest.rocksplicator.LeaderFollowerStateModelFactory$LeaderFollowerStateModel.getLiveHostAndRole(LeaderFollowerStateModelFactory.java:407)
at com.pinterest.rocksplicator.LeaderFollowerStateModelFactory$LeaderFollowerStateModel.onBecomeFollowerFromOffline(LeaderFollowerStateModelFactory.java:449)
```
This is due to state map being `null` in helix external view, and our state transition logic doesn't check for null. 
Fixing this by wrapping the helix `getStateMap` to return an empty map (instead of `null`) when the partition state is empty. 

Note that this also introduced some behavior changes in some cases, e.g. in FOLLOWER->LEADER transition, what previously caused ERROR shard due to NPE, will now let the transition go through for availability purposes. Some comments are added during the review process of the PR. 

ref: https://jira.pinadmin.com/browse/KV-5445

Test Plan:
unit tests still pass with `mvn test`

Deployed a private build and verified:
- cluster works fine from stats (SR, latency, etc)
- failure testing: restarting 3 hosts for a shard at the same time, which previously caused the NPE bug and ERROR shard, but no more issue with this fix

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -212,7 +212,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         // sanity check no existing Leader for up to 59 seconds
         for (int i = 0; i < 60; ++i) {
           view = admin.getResourceExternalView(cluster, resourceName);
-          stateMap = view.getStateMap(partitionName);
+          stateMap = Utils.getStateMap(view, partitionName);
 
           if (!stateMap.containsValue("LEADER")) {
             break;
@@ -313,7 +313,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         LOG.error("[" + dbName + "] Getting external view");
         view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
         LOG.error("[" + dbName + "] Got external view");
-        stateMap = view.getStateMap(partitionName);
+        stateMap = Utils.getStateMap(view, partitionName);
         // changeDBRoleAndUpStream(all_other_followers_or_offlines, "Follower", "my_ip_port")
         for (Map.Entry<String, String> instanceNameAndRole : stateMap.entrySet()) {
           String[] hostPort = instanceNameAndRole.getKey().split("_");
@@ -386,7 +386,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
     public String getLeaderInstance(NotificationContext context) {
       HelixAdmin admin = context.getManager().getClusterManagmentTool();
       ExternalView view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
-      Map<String, String> stateMap = view.getStateMap(partitionName);
+      Map<String, String> stateMap = Utils.getStateMap(view, partitionName);
       for (Map.Entry<String, String> instanceNameAndRole : stateMap.entrySet()) {
         if (instanceNameAndRole.getValue().equalsIgnoreCase("LEADER")) {
           return instanceNameAndRole.getKey();
@@ -398,7 +398,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
     public Map<String, String> getLiveHostAndRole(NotificationContext context, String dbName) {
       HelixAdmin admin = context.getManager().getClusterManagmentTool();
       ExternalView view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
-      Map<String, String> stateMap = view.getStateMap(partitionName);
+      Map<String, String> stateMap = Utils.getStateMap(view, partitionName);
 
       // find live replicas
       Map<String, String> liveHostAndRole = new HashMap<>();
@@ -662,7 +662,7 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         // "live_leader_or_follower")
         HelixAdmin admin = context.getManager().getClusterManagmentTool();
         ExternalView view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
-        Map<String, String> stateMap = view.getStateMap(partitionName);
+        Map<String, String> stateMap = Utils.getStateMap(view, partitionName);
 
         // find upstream which is not me, and prefer leader
         String upstream = null;
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -110,6 +110,10 @@ public class Utils {
     if (partitionName == null || partitionName.isEmpty()) {
       return Collections.emptyMap();
     }
+    
+    if (view == null) {
+      return Collections.emptyMap();
+    }
 
     Map<String, String> result = view.getStateMap(partitionName);
     if (result == null) {
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -50,6 +50,7 @@ import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -212,7 +212,7 @@ public void onBecomeLeaderFromFollower(Message message, NotificationContext cont
         // sanity check no existing Leader for up to 59 seconds
         for (int i = 0; i < 60; ++i) {
           view = admin.getResourceExternalView(cluster, resourceName);
-          stateMap = view.getStateMap(partitionName);
+          stateMap = Utils.getStateMap(view, partitionName);
```
kangnanli, line 215: For this `onBecomeLeaderFromFollower`, when `stateMap` is empty, it will `break` the for loop and treat it as no Leader. Then, it will skip a few section of code later on and eventually promote itself as Leader, is this expected?

(can you add similar explanation what would happen when `{}` returned compared to null, in other places of the code change)

jaricftw: Good question. For this case, this logic is to wait for up to 1min to ensure there is **no existing leader** in the shard, so we won't have two leaders for a shard at the same time. 

If somehow the shard is empty (no leader or followers), it could indicate the shard is temporarily down, and helix decide to promote a leader immediately, but then the shard doesn't have live instances at that point. This is a edge case, but when it happens, we ave a few choices:
- just crash and become a ERROR shard, which is the current behavior without the null check
- continue to become a leader

I think the later seems more like the intended behavior: a state transition is triggered for a shard to go from follower to leader, and the current shard is empty, so we should let it go through. 
Thoughts?

jaricftw: Also added similar explanation in other parts

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -313,7 +313,7 @@ public void onBecomeLeaderFromFollower(Message message, NotificationContext cont
         LOG.error("[" + dbName + "] Getting external view");
         view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
         LOG.error("[" + dbName + "] Got external view");
-        stateMap = view.getStateMap(partitionName);
+        stateMap = Utils.getStateMap(view, partitionName);
```
jaricftw, line 316: behavior change:

this is also part of follower->leader transition, where the leader tries to inform all non-leaders about it's leadership, so the followers can set upstream correctly. It find out the non-leaders by contacting the stateMap.

Currently the behavior is a NPE and cause the leader transition to fail (ERROR shard). 

The new behavior would be to let the leader transition go through. I think this is preferred, we need the leader for the shard to up, and informing other non-leaders should be a best-effort, especially in this case there is none available. (we also have logic to reset upstream in runtime even if a wrong upstream is set)

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -386,7 +386,7 @@ public void onBecomeFollowerFromLeader(Message message, NotificationContext cont
     public String getLeaderInstance(NotificationContext context) {
       HelixAdmin admin = context.getManager().getClusterManagmentTool();
       ExternalView view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
-      Map<String, String> stateMap = view.getStateMap(partitionName);
+      Map<String, String> stateMap = Utils.getStateMap(view, partitionName);
```
jaricftw, line 389: this is inside the `getLeaderInstance` helper function, so this change is straighforward --- if stateMap is empty, it means "no leader", instead of an exception and cause a ERROR shard. 

Currently this function is only called by OFFLINE->FOLLOWER transition

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -398,7 +398,7 @@ public String getLeaderInstance(NotificationContext context) {
     public Map<String, String> getLiveHostAndRole(NotificationContext context, String dbName) {
       HelixAdmin admin = context.getManager().getClusterManagmentTool();
       ExternalView view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
-      Map<String, String> stateMap = view.getStateMap(partitionName);
+      Map<String, String> stateMap = Utils.getStateMap(view, partitionName);
```
jaricftw, line 401: This is within `getLiveHostAndRole`, and is the one where I observed the NPE in production during failure testing, which motivated this PR. 

Here we are trying to find out the "live" instances in the stateMap for a shard. Therefore, if there is no instances in the stateMap, we should return empty result, instead of crashing and ERROR out. This is a straighforward change imo

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
```diff
@@ -662,7 +662,7 @@ public void onBecomeOfflineFromFollower(Message message, NotificationContext con
         // "live_leader_or_follower")
         HelixAdmin admin = context.getManager().getClusterManagmentTool();
         ExternalView view = Utils.getHelixExternalViewWithFixedRetry(admin, cluster, resourceName);
-        Map<String, String> stateMap = view.getStateMap(partitionName);
+        Map<String, String> stateMap = Utils.getStateMap(view, partitionName);
```
jaricftw, line 665: This happens in the FOLLOWER->OFFLINE state transition. 

What it does is to retrieve the upstream for the shard, and tell all other follower/offline replicas to use that upstream. If no upstream found in the stateMap, then no-op.

Therefore this PR change is straightforward, if the stateMap for the shard is empty, it means no upstream, so it's a no-op, and there is no need to make it a ERROR shard.

jaricftw (commented):
@kangnanli replied your comments, with explanation on each case. let me know what you think. 

I also hesitated when making a blanket change for all the places where `stateMap` is used and is subjected to NPE. Hope the explanation helps. 

That said, I am also okay to limit the scope, and only make the change for places where it's absolutely safe.

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -99,6 +100,25 @@ public static ExternalView getHelixExternalViewWithFixedRetry(HelixAdmin admin,
...
+   * @param view Helix external view for a resource
+   * @param partitionName partition name to look up
+   * @return StateMap, which will be empty if the partition is not found, or there is no instance in the partition
+   */
+  public static Map<String, String> getStateMap(ExternalView view, String partitionName) {
+    if (partitionName == null || partitionName.isEmpty()) {
+      return Collections.emptyMap();
+    }
+
+    Map<String, String> result = view.getStateMap(partitionName);
```
kangnanli, line 114: Is it possible that the `view` is also Null? (will that Helix funct ever return Null?)

cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
kangnanli: If the actual zk data is `shard is empty (no leader or followers)`, then, I agree with you, we should let helix promote this current replica to Leader immediately to at least recover the availability.

kangnanli (approved):
lgtm

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
jaricftw: yeah this could be `null` as well, e.g. when it failed to contact Helix.  I have previously added [retry](https://github.com/pinterest/rocksplicator/pull/598) on getting helix external view so it's less likely, but theoretically it can still happen. I've update the PR to check for `view` as well for additional safety.

PR #614, opened 2022-05-08
LeaderFollowerStateModelFactory: set rebuild WAL lag upper bound

jaricftw:
During FOLLOWER->LEADER state transition, follower may decide not to "rebuild" (restore from upstream snapshot) local db if its latest local db update is within the WAL TTL (default to 1h), i.e. that means newer updates have not been expired/deleted in the upstream, so it's able to catch up from replication RPC. 

However, WAL TTL can be arbitrary set, while we default it to 1h, application may increase it to be longer (e.g. 4h, or even a day).  In that case, we should probably trigger a rebuild if the follower is too lagging behind, since log catchup takes time. 

This PR proposes a simple change: regardless of the WAL TTL, we'll trigger rebuild if the local WAL is stale for more than one hour. We chose 1h as a starting point, because that's what the default WAL TTL is (so no production behavior change to existing applications that use the default WAL TTL), we may tune it in the future (e.g. further shrink it if our backup&restore path becomes more reliable)

ref: https://jira.pinadmin.com/browse/KV-5337

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -46,6 +46,7 @@ import org.slf4j.LoggerFactory;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.lang.Math;
 
 /**
  * The LeaderFollower state machine has 5 possible states. There are 7 possible state transitions
@@ -633,8 +634,12 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
         return true;
       } 
 
+      // do not trigger rebuild if the latest db update is within min(wal_ttl_seconds, 1h).
+      // the 1h upper bound is to ensure we are not too lagging behind when catching up from upstream via replication,
+      // since wal_ttl_seconds can be large.
+      long maxLogCatchupTimeSec = Math.min(localStatus.wal_ttl_seconds, 60*60 /* 1h */);
       if (System.currentTimeMillis() <
-          localStatus.last_update_timestamp_ms + localStatus.wal_ttl_seconds * 1000) {
+          localStatus.last_update_timestamp_ms + maxLogCatchupTimeSec * 1000) {
         LOG.error("Replication lag is within the range, skip rebuild " + dbName);
         LOG.error("Last update timestamp in ms: " + String
             .valueOf(localStatus.last_update_timestamp_ms));
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/LeaderFollowerStateModelFactory.java
@@ -637,6 +637,9 @@ public class LeaderFollowerStateModelFactory extends StateModelFactory<StateMode
       // do not trigger rebuild if the latest db update is within min(wal_ttl_seconds, 1h).
       // the 1h upper bound is to ensure we are not too lagging behind when catching up from upstream via replication,
       // since wal_ttl_seconds can be large.
+      // TODO: we hard-code the 1h upper bound because we have been using the default wal_ttl_seconds as 1h
+      // for years in production without ever having to change it, but we should evaluate it in the future
+      // based on the restore speed, and how much lagging behind we can tolerate when a follower goes online.
       long maxLogCatchupTimeSec = Math.min(localStatus.wal_ttl_seconds, 60*60 /* 1h */);
       if (System.currentTimeMillis() <
           localStatus.last_update_timestamp_ms + maxLogCatchupTimeSec * 1000) {


kangnanli:
> However, WAL TTL can be arbitrary set, while we default it to 1h, application may increase it to be longer (e.g. 4h, or even a day). In that case, we should probably trigger a rebuild if the follower is too lagging behind, since log catchup takes time.

So, we think the host-to-host RPC replication is too expensive is the lag is >1h? Do we have any stats to suggest this? Or, do we have any stats to suggest whether 30min, 1h, 1.5h, or 4h is too much lag? What is the limit?

After this change, it seem like, the original design of increase the WAL ttl to allow longer-lagged follower to catchup via RPC is not going to work anymore (ie. for those lagged over 1h). Then:
- the gflag for dynamically set the WAL TTL will seem to useless anymore, since, it is capped at 1h now.
- Few edge cases might cause issues: 
- - if the rebuild process take >1h, then, upon the Restore completed, it will detect it is lagging behind more than 1h again, then, it will rebuild again? right?

jaricftw:
> > However, WAL TTL can be arbitrary set, while we default it to 1h, application may increase it to be longer (e.g. 4h, or even a day). In that case, we should probably trigger a rebuild if the follower is too lagging behind, since log catchup takes time.
> 
> So, we think the host-to-host RPC replication is too expensive is the lag is >1h? Do we have any stats to suggest this? Or, do we have any stats to suggest whether 30min, 1h, 1.5h, or 4h is too much lag? What is the limit?
> 
Good question, yes I added stats for local testing, and will have a PR for that. For now, please refer to the linked ticket in summary, especially in the [end](https://jira.pinadmin.com/browse/KV-5337?focusedCommentId=2076205&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2076205) I have some preliminary stats. It surely depends on how much updates/QPS we have for a shard, but based on one production dataset we tested, it was relatively quick (took <7min to catch up 2.5million updates after lagging behind by 7.5 hours). 

"What is the limit": there is really no limit, it's a matter of tradeoff, depending on how much lag we can tolerate when we bring up a replica online. I think ideally the snapshot recovery  (aka backup & restore) should be trigger as frequent as possible, but in reality that's a heavy operation so we don't wanna do that too aggressively. 

> After this change, it seem like, the original design of increase the WAL ttl to allow longer-lagged follower to catchup via RPC is not going to work anymore (ie. for those lagged over 1h). Then:
> 
> * the gflag for dynamically set the WAL TTL will seem to useless anymore, since, it is capped at 1h now.

Not really, the GFLAG should be mainly used for configuring the WAL TTL, which is **mainly for durability purposes**. We also use it in deciding whether to have snapshot recovery or not during follower->leader transition, but that should only be an indicator. Hence this PR change. 

In fact, the plan is we will use that GFLAG to slowly increase the WAL TTL for clusters that needs higher durability/consistency guarantees. Blocked by this PR. 

> * Few edge cases might cause issues:
> * * if the rebuild process take >1h, then, upon the Restore completed, it will detect it is lagging behind more than 1h again, then, it will rebuild again? right?

This is theoretically true. Do we have stats to suggest so? My observation is we have always been using the default 1h WAL TTL (for years since beginning). So if what you described happens, it should happen regardless of this change. 

Sure we can increase the cap, but I'd prefer we keep it as is (therefore no production behavior change due to this PR), and only relax that as needed. wdty?

kangnanli:
Thanks for the detailed analysis in the JIRA ticket and the benchmarking. regarding to:
> (took <7min to catch up 2.5million updates after lagging behind by 7.5 hours)
this do provide us some confidence, catching up via End2End replication is fairly fast. 
- one thing to notice: the benchmarking might also be impacted by the key/value size. 
- From previous experience/reading, sometimes, the replication cause the Follower suffer much higher latency and lots of compaction and even high CPU on the Follower side. I think this might due to key/value size, which cause the Follower has too many L0 files and trigger minor compaction. In short, the compaction can't catch up with the replication speed, thus, causing the Follower into this endless compaction and slow down the replication. 


regarding to:
> WAL TTL, which is mainly for durability purposes.

I previously thought the WAL is mainly for the replication purpose. If WAL is only for durability purpose, I would think we don't even need 1h. 

> So if what you described happens, it should happen regardless of this change.

Not really, if the backup&restore is slow, e.g. took 2h. Then, by increasing the WAL to 3h, we could complete the backup&restore, and the follower can then catch up the 2h lag by replication. 
- but, after current change, it will be trapped into this backup&restore process, right?

jaricftw:
> Thanks for the detailed analysis in the JIRA ticket and the benchmarking. regarding to:
> 
> > (took <7min to catch up 2.5million updates after lagging behind by 7.5 hours)
> > this do provide us some confidence, catching up via End2End replication is fairly fast.
> 
> * one thing to notice: the benchmarking might also be impacted by the key/value size.
> * From previous experience/reading, sometimes, the replication cause the Follower suffer much higher latency and lots of compaction and even high CPU on the Follower side. I think this might due to key/value size, which cause the Follower has too many L0 files and trigger minor compaction. In short, the compaction can't catch up with the replication speed, thus, causing the Follower into this endless compaction and slow down the replication.
> 
Yeah, thanks for sharing your experience! it's hard to measure the replication speed without a comprehensive benchmarking. I think adding stats is the first step to capture such issues if it ever happens. That by itself worth a separate discussion. 

> regarding to:
> 
> > WAL TTL, which is mainly for durability purposes.
> 
> I previously thought the WAL is mainly for the replication purpose. If WAL is only for durability purpose, I would think we don't even need 1h.

Sorry I meant "WAL TTL". WAL is definitely for both purposes: durability and replication. However, the TTL mainly affects the durability/consistency, because even TTL expired, replication still goes on and ignore the expired logs, but that means a follower may be missing updates, and that's a durability risk.  I am trying to reduce the risk by increasing the TTL in the future.

> 
> > So if what you described happens, it should happen regardless of this change.
> 
> Not really, if the backup&restore is slow, e.g. took 2h. Then, by increasing the WAL to 3h, we could complete the backup&restore, and the follower can then catch up the 2h lag by replication.
> 
> * but, after current change, it will be trapped into this backup&restore process, right?

I meant that, we never change the default 1h setting for years (correct me if I am wrong, but I checked all clusters). So from that perspective, there shouldn't be any impact in practice. 

The only change is, if the corner case every happens is:
- in the old way, we may choose to increase the WAL TTL via gflag, and restart the cluster to mitigate
- in the new way, we may need to change the code, produce a new build, and deploy the cluster to mitigate. 
It's gonna be extra work -- but I doubt it's really necessary, since we are speculating about a failure that never happened. (correct me if I am wrong), so it's worth optimizing. 

Last but not least, I would like to rephrase the purpose the PR as: **decouple the decision of snapshot recovery from WAL TTL** . WAL TTL is one indicator, but shouldn't be used directly. E.g. in the future we may even increase the WAL TTL to be 24h (as this [example](https://github.com/facebook/rocksdb/blob/2e09a54c4fb82e88bcaa3e7cfa8ccbbbbf3635d5/docs/_posts/2014-03-27-how-to-persist-in-memory-rocksdb-database.markdown#L19) suggested), but we certainly don't want to catch up 24h of updates via replication. 

Open for suggestions though, @kangnanli is there anything else in your mind?

kangnanli:
I mainly concern about:
> in the new way, we may need to change the code, produce a new build, and deploy the cluster to mitigate.

otherwise, the PR looks good to me. I would like to the 1h to be another Gflag (which seem overskilled) instead of hard code. I am concerned, in case of those issues, make another build is too slow than just change the gflag and restart the cluster.

kangnanli (approved):
I will r+ this PR since I think there is no need to block this PR. 
Operator should rollout w/t caution and monitor whether this PR impact those *edge cases* we mentioned in the PR discussion.

jaricftw:
> I mainly concern about:
> 
> > in the new way, we may need to change the code, produce a new build, and deploy the cluster to mitigate.
> 
> otherwise, the PR looks good to me. I would like to the 1h to be another Gflag (which seem overskilled) instead of hard code. I am concerned, in case of those issues, make another build is too slow than just change the gflag and restart the cluster.

yeah, I wish we also have stats to measure the speed of rebuild. That way we can have more confidence. 
I'll also capture some of the discussion in the comments to explain the 1h setting and possible TODOs. Thanks.

PR #615, opened 2022-05-10
cluster_management: support an option to close admin client connection

jaricftw:
The current `getAdminClient` creates a Admin thrift client for various usage during state transition or 
even during runtime (e.g. in `PartitionStateUpdater`). Under the hood it opens a TCP socket which the
thrift client uses to establish connections. 

The socket is not closed after usage, which could be problematic, if there are a large number of admin clients
used. This is the case for the usage in `PartitionStateUpdater`, which schedules all leader partitions in a process
to retrieve local db state (via admin client) every few minutes. Eventually, we see process crash or run into failing state where it exhausted the socket count. 

This PR proposes a simple fix: provide a new `getAdminClientOnSocket` to asks the caller to provide the socket
and hence let the caller know they should be responsible for closing the socket. 

Test plan:
mvn test still work
tested in a private build, among many other tests, please see an [internal doc](https://docs.google.com/document/d/11-ybbeTkZsbw3QH4RIfpH9seHcP5rkbWzbAaguGAy2o/edit#) for detailed investigation and validation.

### DIFF
--- a/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
+++ b/cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
@@ -135,6 +135,9 @@ public class Utils {
 
   /**
    * Build a thrift client to host:adminPort
+   * Note this opens a socket underneath and doesn't close it. This would become a problem if 
+   * many admin clients are created on demand, causing connection leaking. 
+   * To properly close the connection, use getAdminClientOnSocket instead. 
    * @param host
    * @param adminPort
    * @return a client object
@@ -142,6 +145,18 @@ public class Utils {
    */
   public static Admin.Client getAdminClient(String host, int adminPort) throws TTransportException {
     TSocket sock = new TSocket(host, adminPort);
+    sock.open();
+    return new Admin.Client(new TBinaryProtocol(sock));
+  }
+
+   /**
+   * Open the provided TSocket and build a Admin thrift client on top of it for application usage.
+   * It's caller's responsibility to close the socket after usage. 
+   * @param sock user constructed TSocket
+   * @return an Admin client object
+   * @throws TTransportException 
+   */
+  public static Admin.Client getAdminClientOnSocket(TSocket sock) throws TTransportException {
     sock.open();
     return new Admin.Client(new TBinaryProtocol(sock));
   }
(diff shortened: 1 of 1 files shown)


cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -146,6 +149,18 @@ public static Admin.Client getAdminClient(String host, int adminPort) throws TTr
     return new Admin.Client(new TBinaryProtocol(sock));
```
kangnanli, line 149: So, any call to this api will have socket leak then? do we need to fix them?

jaricftw: yeah, we need to and I plan to do a follow up PR. Note that other places are not causing problems currently because they are using the admin client in StateTransition, so it's usually one off, unlike `PartitionStateUpdater` which happens continuously.

cluster_management/src/main/java/com/pinterest/rocksplicator/Utils.java
```diff
@@ -313,12 +328,15 @@ public static long getLocalLatestSequenceNumber(String dbName, int adminPort)
   public static long getLatestSequenceNumber(String dbName, String host, int adminPort) {
     LOG.error("Get seq number from " + host + " for " + dbName);
     try {
-      Admin.Client client = getAdminClient(host, adminPort);
+      TSocket sock = new TSocket(host, adminPort);
```
rajathprasad, line 331: I'm wondering if it's possible to use the RAII approach here. If there is an exception in the following lines, the socket will be leaked again. Should we close the socket in the `finally` section?

jaricftw: I also thought about that, and open to make that change. The reason I do it here is because: when there is any exception in creating TSocket or open the the socket, that means the socket isn't in `open` state, so there is no need to close it. 

However, agree that's not very straightward, so I'll try to use `finally`

jaricftw: updated. I ended up not using `finally` only for the actual request (since it's no needer TSocket construction and open as I explained above). SO it's more like the example 4 here https://www.programcreek.com/java-api-examples/?api=org.apache.thrift.transport.TSocket. Let me know if you have other thoughts

kangnanli (approved):
lgtm on my side

jaricftw:
The current getAdminClient creates a Admin thrift client for various usage during state transition or
even during runtime (e.g. in PartitionStateUpdater). Under the hood it opens a TCP socket which the
thrift client uses to establish connections.

The socket is not closed after usage, which could be problematic, if there are a large number of admin clients
used. This is the case for the usage in PartitionStateUpdater, which schedules all leader partitions in a process
to retrieve local db state (via admin client) every few minutes. Eventually, we see process crash or run into failing state where it exhausted the socket count.

This PR proposes a simple fix: provide a new getAdminClientOnSocket to asks the caller to provide the socket
and hence let the caller know they should be responsible for closing the socket.

Test plan:
mvn test still work
tested in a private build, among many other tests, please see an [internal doc](https://docs.google.com/document/d/11-ybbeTkZsbw3QH4RIfpH9seHcP5rkbWzbAaguGAy2o/edit#) for detailed investigation and validation.

PR #618, opened 2022-05-19
rocksdb_replicator: do not count observer replication ACK

jaricftw:
On replication mode TWO (2-ack mode), upon writes, the leader will need to wait for at least 
one follower to receive the update, before returning successfully to to the client. Currently it does so by tracking the `max_seq_no_acked_` which is the maximum ACK-ed sequence number from **any** replication requests. That's problematic, because the replication requests could come from either `FOLLOWER` or `OBSERVER`.  See https://github.com/pinterest/rocksplicator/pull/601 where we introduced the `OBSERVER` role.

This PR corrects that behavior, specifically, on the FOLLOWER/OBSERVER side, set the `role` in each replication request, and on the leader side, do not count OBSERVER sequence number when updating `max_seq_no_acked_`. Log a counter for visibility in that case. 

-- test plan -- 
Added unit test to verify the behavior. Will also try to deploy a test build to confirm.

### DIFF
--- a/rocksdb_replicator/replicated_db.cpp
+++ b/rocksdb_replicator/replicated_db.cpp
@@ -318,6 +318,7 @@ void RocksDBReplicator::ReplicatedDB::pullFromUpstream() {
   req.db_name = db_name_;
   req.max_wait_ms = FLAGS_replicator_max_server_wait_time_ms;
   req.max_updates = FLAGS_replicator_max_updates_per_response;
+  req.set_role(role_);
 
   incCounter(kReplicatorPullRequests, 1, db_name_);
 
@@ -446,8 +447,13 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
     logMetric(kReplicatorLeaderSequenceNumbersBehind, seq_no - leaderSeqNum, db ->db_name_);
   }
 
-  // post the largest sequence number the Slave has committed
-  max_seq_no_acked_.post(seq_no);
+  // Post the largest sequence number the Slave has committed.
+  // If the request comes from an Observer, ignore the sequence number update.
+  if (request->__isset.role && request->role == ReplicaRole::OBSERVER) {
+    incCounter(kReplicatorHandleObserverRequests, 1, db->db_name_);
+  } else {
+    max_seq_no_acked_.post(seq_no);
+  }
 
   auto replication_mode =  common::DBConfigManager::get()->getReplicationMode(db_name_);
   // for now we have to support both till all clusters are migrated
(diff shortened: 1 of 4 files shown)


rocksdb_replicator/tests/rocksdb_replicator_test.cpp
```diff
@@ -623,6 +623,120 @@ TEST(RocksDBReplicatorTest, 1_master_1_slave_replication_mode_2) {
...
+  const int16_t master_port = 9092;
+  const int16_t slave_port = 9093;
+  const int16_t observer_port = 9094;
+  Host master(master_port);
+  Host slave(slave_port);
+  Host observer(observer_port);
+
+  auto db_master = cleanAndOpenDB("/tmp/db_master");
+  auto db_slave = cleanAndOpenDB("/tmp/db_slave");
+  auto db_observer = cleanAndOpenDB("/tmp/db_observer");
```
mradwan, line 643: shall we use leader/follower here to be consistent with the naming for ReplicaRole?

mradwan (approved):
LGTM, It is good we have such utilities to write unit tests

rocksdb_replicator/tests/rocksdb_replicator_test.cpp
jaricftw: yeah, I plan to change all places at the same time in a follow-up PR. This file still uses master/slave in other tests so it's okay to be consistent.

jaricftw:
landing this the PR has an approver and has been open for quite a while. 

@iprentic would still appreciate a review when you get a chance, since this touches rocksobserver

rocksdb_replicator/replicated_db.cpp
```diff
@@ -446,8 +447,13 @@ void RocksDBReplicator::ReplicatedDB::handleReplicateRequest(
...
   }
 
-  // post the largest sequence number the Slave has committed
-  max_seq_no_acked_.post(seq_no);
+  // Post the largest sequence number the Slave has committed.
+  // If the request comes from an Observer, ignore the sequence number update.
+  if (request->__isset.role && request->role == ReplicaRole::OBSERVER) {
+    incCounter(kReplicatorHandleObserverRequests, 1, db->db_name_);
+  } else {
+    max_seq_no_acked_.post(seq_no);
```
iprentic, line 455: Do we want a stat for this case as well?

jaricftw: We could, though this is the common/default case, so not as important/useful as the observer requests where we need more visibility

PR #620, opened 2022-05-23
Deprecates support for Ubuntu 14

zhanyong-wan:
[rocksplicator](https://github.com/pinterest/rocksplicator) has conditional code that depends on `__GNUC__`, the major version of GCC used for compiling the code.  This was introduced for the Ubuntu 14 => Ubuntu 18 migration, as 14 and 18 use different GCC versions.

Per discussion with Bo Liu, the U14 code path is no longer used and can be cleaned up.

Why now: Pinterest is evaluating Clang as an alternative compiler.  The latest Clang (v14) defines `__GNUC__` as 4, since it's compatible with GCC 4.  Therefore, when rocksplicator is compiled with Clang, the U14 code path is incorrectly picked, which doesn't compile in the U18 environment.  I discussed this with Bo Liu and we decided that it's time to remove the U14 code path.

### DIFF
--- a/common/aws_s3_rate_limiter.h
+++ b/common/aws_s3_rate_limiter.h
@@ -20,11 +20,7 @@
 
 #include "aws/core/utils/ratelimiter/RateLimiterInterface.h"
 
-#if __GNUC__ >= 8
 #include "folly/synchronization/AtomicStruct.h"
-#else
-#include "folly/AtomicStruct.h"
-#endif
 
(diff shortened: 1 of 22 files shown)


zhanyong-wan:
Please withhold the review/merge.  Per chat with Bo Liu, we need to first upgrade the Dockerfile in rocksplicator to U18 and build and publish an image built from the new U18 Dockerfile.

PR #622, opened 2022-06-16
rocksplicator Adds stats for compact request in rocksplicator

mradwan:
In this diff I add few positive and negative logging for compaction api calls and results.

Test plan:
Build and deploy binary to canary host.
Verify logs are printed and metrics show up.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -177,6 +177,7 @@ const std::string kHDFSRestoreMs = "hdfs_restore_ms";
 const std::string kS3BackupMs = "s3_backup_ms";
 const std::string kS3RestoreMs = "s3_restore_ms";
 const std::string kDeleteDBFailure = "delete_db_failure";
+const std::string kCompactDb = "compact_db";
 
 int64_t GetMessageTimestampSecs(const RdKafka::Message& message) {
   const auto ts = message.timestamp();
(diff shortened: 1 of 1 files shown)


jaricftw (approved):

mradwan:
Forget to share the link: https://statsboard.pinadmin.com/share/vw2hd

rocksdb_admin/admin_handler.cpp
```diff
@@ -2134,8 +2135,11 @@ void AdminHandler::async_tm_compactDB(
       CompactDBResponse>>> callback,
     std::unique_ptr<CompactDBRequest> request) {
   ::admin::AdminException e;
+  const std::string statName = kCompactDb + " db=" + request->db_name;
```
jaricftw, line 2138: taking a second look, I think this may have metrics cardinality concerns, since we tag on a per-db/shard level. Would it be sufficient to emit metrics on a per-dataset level?

jaricftw (commented):
Accepted earlier, but just realized something.

rocksdb_admin/admin_handler.cpp
mradwan: @jaricftw understandable given rocksplicator doesn't follow this pattern. However we already have per db disk usage for example and many other metrics in rockstorewidecolumn. these metrics are only bumped on compaction request which happens once per day. so IMO it should be fine.

WDYT?

jaricftw: the per-db disk usage stats are partially blocked by Goku already (only 24h for 1m resolution): https://statsboard.pinadmin.com/share/2msys. Need rollup for longer lookback. 

Previously I tried to explore enable per-db usage for other replication metrics: https://jira.pinadmin.com/browse/KV-5608, but the conclusion was that Goku won't be able to support that. 

That said, I agree the compaction doesn't happen often, so we don't have that many metrics series to generate, so at least it won't be an issue from a capacity perspective. On the query side, as long as we don't expand host & db tag at the same time, we are probably okay. 

I am okay with giving it a try.

PR #623, opened 2022-06-22
Add support for passing db name to rocksdb options generator

mradwan:
This diff adds support for passing db name to rocksdb options generator.
Since this change spans across multiple projects in cosmos I added a second constructor to make it less intrusive.
I can merge this change in cosmos and make another change to update everybody's rocksdb options to accept db name.
Reason db_name is useful is for logging shard specific stats in compaction filters for improved debugging experience.

For testing I ran

```
docker run -v /home/meariby/code/rocksplicator:/rocksplicator -v $HOME/docker-root:/root -ti gopalrajpurohit/rocksplicator-build:librdkafka_1_4_0 bash
cd /rocksplicator && mkdir -p build && cd build && cmake .. && make -j
cd /rocksplicator && mkdir -p build && cd build && cmake .. && make -j && make test
```

I will also make sure cosmos build with this change as-is as part of test plan.

Cosmos diff:
https://phabricator.pinadmin.com/D901271

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -245,7 +245,7 @@ folly::Future<std::unique_ptr<rocksdb::DB>> GetRocksdbFuture(
 
 
 std::unique_ptr<::admin::ApplicationDBManager> CreateDBBasedOnConfig(
-    const admin::RocksDBOptionsGeneratorType& rocksdb_options) {
+    const admin::RocksDBOptionsGenerator& rocksdb_options) {
   auto db_manager = std::make_unique<::admin::ApplicationDBManager>();
   std::string content;
   CHECK(folly::readFile(FLAGS_shard_config_path.c_str(), content));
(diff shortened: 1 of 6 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -437,9 +437,17 @@ void deleteTmpDBs() {
 
 namespace admin {
 
+AdminHandler::AdminHandler(
+  std::unique_ptr<ApplicationDBManager> db_manager,
+  RocksDBOptionsGeneratorType rocksdb_options): AdminHandler(
```
mradwan, line 442: This is the old constructor signature it will just wrap the old function type and pass it as new function type.

mradwan (commented):
Just highlighting something

mradwan:
> How do you plan to use this, do you need to pinpoint a specific shard in the log?

@jaricftw one example is to know which db is [running version trimming compaction](https://phabricator.pinadmin.com/diffusion/CM/browse/master/rockstorewidecolumn/rocksdb_options.cpp;1a6fe1429f1e05449f4d3a54304c153cfb8244c5$233-234) filter to print out how many records deleted stats in the logs and potentially to statsboard as well

I understand this looks like a lot of plumbing. however I will send a change to cosmos after to use the new signature only and nuke the old constructor hence the code will be just as is but with better functionality.

jaricftw (approved):

PR #624, opened 2022-06-22
Add Gflag to enable/disable incremental backup feature

qiyanghe1998:
This diff adds a gflag to enable/disable incremental backup feature.
This is the first step for incremental backup.

For testing, I log some info to show the gflag is enabled.


mradwan:
@qiyanghe1998 Thanks for making the change it looks great so far. 

One comment I have is I see we already are adding thread / stop mechanism however we havent done a design review on this part yet.

For this, I suggest we hold off any changes apart from the flag and maybe logging that the feature is enabled. then after we do our design/brain storming session we can put in the rest of changes when we are ready to use them.

What do you think?

PR #625, opened 2022-06-23
rocksplicator EASY clean up s3 tmp upon startup

mradwan:
Right now we s3_tmp to store data downloaded from s3, data being uploaded to s3 and data being added to RO clusters.

This change will clean up s3_tmp upon startup

This way any obsolete backup/failures are deleted since they are not going to be used anymore.

I debated whether we shall do this in separate thread or not but deleting data usually takes less than few seconds and it's speed is proportional to amount of data we have so in 99.9% of cases it should be blazing fast.

Note: this doesn't affect path of FLAGS_enable_checkpoint_backup enabled here is the breakdown of code behavior

```
backup: (copy from master/rocksdb snap)
  auto local_path = folly::stringPrintf("%ss3_tmp/%s%d/", FLAGS_rocksdb_dir.c_str(), request->db_name.c_str(), ts);
restore: (recover from s3)
    if FLAGS_enable_checkpoint_backup:
        FLAGS_rocksdb_dir + request->db_name
    else
        folly::stringPrintf("%ss3_tmp/%s%d/", FLAGS_rocksdb_dir.c_str(), request->db_name.c_str(), ts)
add db: (Uploads)
  auto local_path = FLAGS_rocksdb_dir + "s3_tmp/" + request->db_name + "/";
```

Test plan:
Deployed to dev cluster
Verified that s3_tmp is cleaned up after deploy
NOTE: THIS WILL ONLY CLEAN UP if user is prod otherwise it will cause an error.

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -435,6 +435,10 @@ void deleteTmpDBs() {
   }
 }
 
+inline std::string getS3TmpPath() {
+  return FLAGS_rocksdb_dir + "s3_tmp/"; 
+}
+
 }  // anonymous namespace
 
 namespace admin {
@@ -503,6 +507,14 @@ AdminHandler::AdminHandler(
   // Initialize the atomic int variables
   num_current_s3_sst_downloadings_.store(0);
   num_current_s3_sst_uploadings_.store(0);
+  // All interactions with admin handler that involve s3_tmp folder will
+  // be called on the admin handler instance, these are
+  // 1) backupDbToS3
+  // 2) addS3SstFilesToDB
+  // 3) restoreDbFromS3
+  // thus no way they can be called before the handler instance 
+  // finishes construction.
+  initS3Tmp();
 }
 
 AdminHandler::~AdminHandler() {
@@ -969,7 +981,7 @@ void AdminHandler::async_tm_backupDBToS3(
   common::Timer timer(kS3BackupMs);
   LOG(INFO) << "S3 Backup " << request->db_name << " to " << request->s3_backup_dir;
   auto ts = common::timeutil::GetCurrentTimestamp();
-  auto local_path = folly::stringPrintf("%ss3_tmp/%s%d/", FLAGS_rocksdb_dir.c_str(), request->db_name.c_str(), ts);
+  auto local_path = folly::stringPrintf("%s%s%d/", getS3TmpPath(), request->db_name.c_str(), ts);
   boost::system::error_code remove_err;
   boost::system::error_code create_err;
   boost::filesystem::remove_all(local_path, remove_err);
@@ -1163,7 +1175,7 @@ void AdminHandler::async_tm_restoreDBFromS3(
 
   auto ts = common::timeutil::GetCurrentTimestamp();
   auto local_path = FLAGS_enable_checkpoint_backup ? FLAGS_rocksdb_dir + request->db_name :
-                    folly::stringPrintf("%ss3_tmp/%s%d/", FLAGS_rocksdb_dir.c_str(), request->db_name.c_str(), ts);
+                    folly::stringPrintf("%s%s%d/", getS3TmpPath(), request->db_name.c_str(), ts);
   boost::system::error_code remove_err;
   boost::system::error_code create_err;
   boost::filesystem::remove_all(local_path, remove_err);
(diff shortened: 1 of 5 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -2182,4 +2183,18 @@ std::vector<std::string> AdminHandler::getAllDBNames() {
     return db_manager_->getAllDBNames();
 }
 
+void AdminHandler::initS3Tmp() {
+  auto s3_path = folly::stringPrintf("%ss3_tmp/", FLAGS_rocksdb_dir.c_str());
```
jaricftw, line 2187: This `s3_path` will also be used elsewhere right? Shall we construct the path for once and use it throughout the program, would also be less error-prone

rocksdb_admin/admin_handler.cpp
```diff
@@ -503,6 +503,7 @@ AdminHandler::AdminHandler(
   // Initialize the atomic int variables
   num_current_s3_sst_downloadings_.store(0);
   num_current_s3_sst_uploadings_.store(0);
+  initS3Tmp();
```
jaricftw, line 506: just making sure: does any of the above logic could race against modifying the same directory?  Would be good to add a commend on why we the s3_tmp initialization in the end of `AdminHandler` initialization.

rocksdb_admin/admin_handler.cpp
```diff
@@ -2182,4 +2183,18 @@ std::vector<std::string> AdminHandler::getAllDBNames() {
     return db_manager_->getAllDBNames();
 }
 
+void AdminHandler::initS3Tmp() {
```
jaricftw, line 2186: can we unit test this. 

We could also extract the delete-and-create logic into a separate function, if that makes  testing easier.

jaricftw (commented):
I think this can be a simple start, but it will only clean up the stale files upon process start-up. I.e. the in case of the backup&restore failure, some shards are in ERROR state but I think the process is still up and running, so stale files will be left on disk until we restart the process.

mradwan:
@jaricftw I added unit test and a comment explaining why there shouldnt be any races.
Please take another look.

rocksdb_admin/admin_handler.cpp
jaricftw: thanks for adding the comment, would it make it safer to do the init at the start of the AdminHandler initialization? I worry about background threads that may be created before this

rocksdb_admin/tests/admin_handler_test.cpp
```diff
@@ -882,6 +882,23 @@ TEST(AdminHandlerTest, MetaData) {
   EXPECT_EQ(meta.db_name, db_name);
 }
 
+
+TEST(AdminHandlerTest, InitS3Tmp) {
+  gflags::FlagSaver saver;
+  FLAGS_rocksdb_dir = "/tmp/admin_handler_test/";
+  auto s3_path = FLAGS_rocksdb_dir + "s3_tmp/"; 
+  auto test_directory = s3_path + "test_directory";
+  clearAndCreateDir(FLAGS_rocksdb_dir);
```
jaricftw, line 891: didn't realize we have `clearAndCreateDir` already, should use move that into a production utilization function, and use that in the s3_tmp directory initialization?

rocksdb_admin/admin_handler.cpp
mradwan: @jaricftw how would background threads going to call an instance function on AdminHandler before it's constructed?

I grepped for s3_tmp this is the only place where this exists.

For thrift server initialization we would do the following
- Build the handler
- Build the server using the handler
- Server start accepting connections

rocksdb_admin/tests/admin_handler_test.cpp
jaricftw: wdyt?

mradwan:
@jaricftw I addressed your comment. Do you mind giving this another look?

rocksdb_admin/utils.h
```diff
@@ -41,4 +41,9 @@ bool EncodeThriftStruct(const T& obj, std::string* data) {
   return !ex;
 }
 
+// Will remove and create a directory
+// returns true if any of these operations failed.
```
jaricftw, line 45: should we return `false` upon failure? `true` usually indicates success

jaricftw: alternatively, how about we just return the error message, and when it succeeds, it'll be empty.

jaricftw (approved):
LGTM after comments

PR #627, opened 2022-06-23
Add retries to s3 operations in rocksplicator

qiyanghe1998:
This diff adds retry logic to the restore operation from S3. 
Because there is some dependency between different steps in restoreDBFromS3, so I have not implemented fine-grained retry logic so far.

### DIFF
--- a/common/s3util.cpp
+++ b/common/s3util.cpp
@@ -82,6 +82,8 @@ namespace {
   const std::string kS3GetObjectCallable = "s3_getobject_callable";
   const std::string kS3CopyObject = "s3_copyobject";
   const std::string kS3DeleteObject = "s3_deleteobject";
+  const int kS3RetryLimit = 3;
+  const int kS3RetrySec = 1;
 }
 
 namespace common {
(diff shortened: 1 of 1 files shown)


qiyanghe1998:
The local file systems seldom have troubles, so I only need to retry the s3 related operations used by ```local_s3_util``` in ```admin_handler.cpp```. For test, I pass all the unit tests.

mradwan (commented):
I see the retry logic is repeated amongst the 3 functions. Can we instead abstract the retry logic in a separate function and reuse it?

You can define a templatized function that takes in a lambda, runs it until it succeeds.
```
template<typename F>
void runWithRetries(F&& f) {
     while ... {
           auto result = f();
           if(f.IsSuccess()) {
                ...
           }
     }
}
```

There might be a bit of work to make sure all functions compatible with this but in worst case you can also add parameter bool& should_retry and let the function itself decide that.


Second thing is we should probably make the number of retries and retry policy (sleep ms) configurable by constructor of s3util instead of making it constant.

qiyanghe1998:
> template

qiyanghe1998:
> Great work Qiyang!
> 
> I see the retry logic is repeated amongst the 3 functions. Can we instead abstract the retry logic in a separate function and reuse it?
> 
> You can define a templatized function that takes in a lambda, runs it until it succeeds.
> 
> There might be something in folly already does that but might be complicated to incorporate
> 
> ```
> template<typename F>
> void runWithRetries(F&& f) {
>      while ... {
>            auto result = f();
>            if(f.IsSuccess()) {
>                 ...
>            }
>      }
> }
> ```
> 
> There might be a bit of work to make sure all functions compatible with this but in worst case you can also add parameter bool& should_retry and let the function itself decide that.
> 
Yeah, I can treat ```putObject``` and ```getObject``` in this way. For ```listAllObjects```, I want to retry from the previous marker which may be not that compatible with the templatized function. 

> Second thing is we should probably make the number of retries and retry policy (sleep ms) configurable by constructor of s3util instead of making it constant.

mradwan:
This means retry can be implemented inside listAllObjects, right? where the inner function is the one that handles the marker.

qiyanghe1998:
By https://stackoverflow.com/questions/45327453/aws-s3-how-to-set-maximum-number-of-retries-in-c and https://sdk.amazonaws.com/cpp/api/0.14.3/class_aws_1_1_client_1_1_default_retry_strategy.html, I change the ```aws_config``` to set up retry logic.

mradwan:
This looks great, even better and better! do you know if there is a way to test it?

PR #628, opened 2022-06-29
Enable IMDSv2 query for getAvailabilityZoneImpl

xia0pin9:
IMDSv2 uses session-oriented requests to mitigate several types of vulnerabilities that could be used to attempt to access the IMDS, protecting against malicious activities such as SSRF attacks. The change does not impact existing usage of "getAvailabilityZoneImpl" call.

Before:
```
➜  ~ curl --silent --max-time 10 --connect-timeout 5 http://169.254.169.254/latest/meta-data/placement/availability-zone
us-east-1a%
```
After:
```
➜  ~ TOKEN=$(curl -X PUT -H "X-aws-ec2-metadata-token-ttl-seconds: 300" http://169.254.169.254/latest/api/token 2> /dev/null) && curl --silent --max-time 10 --connect-timeout 5 -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/placement/availability-zone
us-east-1a%
```

### DIFF
--- a/common/availability_zone.cpp
+++ b/common/availability_zone.cpp
@@ -42,8 +42,10 @@ bool isAllowedAz(const std::string& az) {
 
 std::string getAvailabilityZoneImpl() {
   static const std::string command =
-    "curl --silent --max-time 10 --connect-timeout 5 "
-    "http://169.254.169.254/latest/meta-data/placement/availability-zone";
+      "TOKEN=$(curl -X PUT -H \"X-aws-ec2-metadata-token-ttl-seconds: 300\" "
+      "http://169.254.169.254/latest/api/token 2> /dev/null) && curl -H "
+      "\"X-aws-ec2-metadata-token: $TOKEN\" "
+      "http://169.254.169.254/latest/meta-data/placement/availability-zone";
 
   auto f = popen(command.c_str(), "re");
   if (f == nullptr) {
--- a/common/availability_zone.cpp
+++ b/common/availability_zone.cpp
@@ -42,10 +42,10 @@ bool isAllowedAz(const std::string& az) {
 
 std::string getAvailabilityZoneImpl() {
   static const std::string command =
-      "TOKEN=$(curl -X PUT -H \"X-aws-ec2-metadata-token-ttl-seconds: 300\" "
-      "http://169.254.169.254/latest/api/token 2> /dev/null) && curl -H "
-      "\"X-aws-ec2-metadata-token: $TOKEN\" "
-      "http://169.254.169.254/latest/meta-data/placement/availability-zone";
+    "TOKEN=$(curl -X PUT -H \"X-aws-ec2-metadata-token-ttl-seconds: 300\" "
+    "http://169.254.169.254/latest/api/token 2> /dev/null) && "
+    "curl --silent --max-time 10 --connect-timeout 5 -H \"X-aws-ec2-metadata-token: $TOKEN\" "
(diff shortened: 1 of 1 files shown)


jaricftw:
> No description provided.

hi @xia0pin9 can you add description on why you are making this change, and what you have done to test/verify it

xia0pin9:
> > No description provided.
> 
> hi @xia0pin9 can you add description on why you are making this change, and what you have done to test/verify it

updated

jaricftw:
> The change does not impact existing usage of "getAvailabilityZoneImpl" call.

Just to confirm: what happens if the curl command to fetch token fails? will that cause the entire command to fail and hence failing to retrieve AZ?

Additionally, I would like to see test plan for this change. E.g. the command output for the new command you listed in the description. And ideally also a private build: you can follow the instruction in https://w.pinadmin.com/display/KVSYS/Rocksplicator+Release+Guide#RocksplicatorReleaseGuide-DevTesting(PrivateBuild) to produce a private build, and ask in #kv-systems team to help with a deployment.

xia0pin9:
> > The change does not impact existing usage of "getAvailabilityZoneImpl" call.
> 
> Just to confirm: what happens if the curl command to fetch token fails? will that cause the entire command to fail and hence failing to retrieve AZ?
> 
> Additionally, I would like to see test plan for this change. E.g. the command output for the new command you listed in the description. And ideally also a private build: you can follow the instruction in https://w.pinadmin.com/display/KVSYS/Rocksplicator+Release+Guide#RocksplicatorReleaseGuide-DevTesting(PrivateBuild) to produce a private build, and ask in #kv-systems team to help with a deployment.

IMDS token endpoint has 1000 TPS (Transaction per Second) per instance, so the token fetch failure should be very rare, it will also break lots of existing services in that case. And if it indeed fail, it will cause the entire command to fail and failing to retrieve AZ info.

The output for both before/after commands are already listed in the description, they are the same (`us-east-1a`)

jaricftw:
> > > The change does not impact existing usage of "getAvailabilityZoneImpl" call.
> > 
> > 
> > Just to confirm: what happens if the curl command to fetch token fails? will that cause the entire command to fail and hence failing to retrieve AZ?
> > Additionally, I would like to see test plan for this change. E.g. the command output for the new command you listed in the description. And ideally also a private build: you can follow the instruction in https://w.pinadmin.com/display/KVSYS/Rocksplicator+Release+Guide#RocksplicatorReleaseGuide-DevTesting(PrivateBuild) to produce a private build, and ask in #kv-systems team to help with a deployment.
> 
> IMDS token endpoint has 1000 TPS (Transaction per Second) per instance, so the token fetch failure should be very rare, it will also break lots of existing services in that case. And if it indeed fail, it will cause the entire command to fail and failing to retrieve AZ info.
> 

Ideally we'd make it more resilient by breaking up the commands: fetch the token first, and if that fails, fall back to IMDSV1. 

But if we are confident about the reliability of the IMDSV2 calls, I think this simple change is okay.  Note that most of our services uses this call upon start-up, and if the command fails, the process will crash immediately. We'll probably need some retry mechanism as a follow up to make it more resilent -- if we ever observed such failures. 

> The output for both before/after commands are already listed in the description, they are the same (`us-east-1a`)
got it. Sorry, missed that earlier.

jaricftw (approved):
LGTM

jaricftw:
Some background on this change: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

PR #629, opened 2022-07-16
refactor backup from admin_handler to application_db

qiyanghe1998:
In order to enable incremental backup, I copy the createCheckpoint logic from admin_handler to application_db and change some headers. I plan to add some S3 related setting in application_db.

Here is the structure of the code, one admin_handler contains one application_db_manager and one application_db_manager contains many application_db. 

Previously, the backup is triggered by the request. All the method about how to deal with backup request is implemented in admin_handler. Now, we need to do incremental backup periodically, so I launch a thread in the constructor of admin_handler. Thus, it is not a good idea to invoke other methods defined in admin_handler. I let that thread invoke an incremental_backup method defined in application_db_manager where I plan to implement different kinds of scheduling strategies. Because backup is designed for each database, so I plan to implement those stuffs in application_db.

Follow my design, I need to pass some variables used in admin_handler to application_db (e.g. meta_db, s3_util, some gflags), I think that is kind of ugly. I will try to optimize it later. 

In another aspect, the previous backup request contains db_name (which db we needs to backup) and some s3 settings (s3_bucket, s3_backup_dir, limit_mbs). I add these settings to application_db right now. I am not sure whether it is a good idea to couple s3 settings and application_db since it will make db construction more complex and some db does not need backup. I think it is better to create another class for s3 settings and another bool variable to represent whether the db will be backuped.

Look forward to new ideas how to refactor!

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -57,6 +57,12 @@
 #include "rocksdb_replicator/rocksdb_replicator.h"
 #include "rocksdb_replicator/thrift/gen-cpp2/Replicator.h"
 #include "thrift/lib/cpp2/protocol/Serializer.h"
+#if __GNUC__ >= 8
+#include "folly/executors/CPUThreadPoolExecutor.h"
+#include "folly/system/ThreadName.h"
+#else
+#include "wangle/concurrent/CPUThreadPoolExecutor.h"
+#endif
 
 DEFINE_string(hdfs_name_node, "hdfs://hbasebak-infra-namenode-prod1c01-001:8020",
               "The hdfs name node used for backup");
@@ -126,6 +132,17 @@ DEFINE_int32(async_delete_dbs_frequency_sec,
 DEFINE_int32(async_delete_dbs_wait_sec,
              60,
              "How long in sec to wait between the dbs deletion");
+
+#if __GNUC__ >= 8
+using folly::CPUThreadPoolExecutor;
+using folly::LifoSemMPMCQueue;
+using folly::QueueBehaviorIfFull;
+#else
+using wangle::CPUThreadPoolExecutor;
+using wangle::LifoSemMPMCQueue;
+using wangle::QueueBehaviorIfFull;
+#endif
+
 namespace {
 
 const std::string kMetaFilename = "dbmeta";
@@ -445,7 +462,7 @@ AdminHandler::AdminHandler(
 
   if (FLAGS_enable_async_incremental_backup_dbs) {
     backup_manager_ = std::make_unique<ApplicationDBBackupManager>(db_manager_.get(), S3UploadAndDownloadExecutor(), 
-      meta_db_, &db_admin_lock_, FLAGS_rocksdb_dir, FLAGS_checkpoint_backup_batch_num_upload);
+      meta_db_.get(), &db_admin_lock_, FLAGS_rocksdb_dir, FLAGS_checkpoint_backup_batch_num_upload);
   }
 
   folly::splitTo<std::string>(
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -448,7 +448,7 @@ void deleteTmpDBs() {
 namespace admin {
 
 AdminHandler::AdminHandler(
-  std::unique_ptr<ApplicationDBManager> db_manager,
+  std::shared_ptr<ApplicationDBManager> db_manager,
   RocksDBOptionsGeneratorType rocksdb_options): AdminHandler(
       std::move(db_manager), 
       [rocksdb_options](const std::string& dataset, const std::string& db) {
@@ -456,10 +456,10 @@ AdminHandler::AdminHandler(
   }) {}
 
 AdminHandler::AdminHandler(
-    std::unique_ptr<ApplicationDBManager> db_manager,
+    std::shared_ptr<ApplicationDBManager> db_manager,
     RocksDBOptionsGenerator rocksdb_options)
   : db_admin_lock_()
-  , db_manager_(std::move(db_manager))
+  , db_manager_(db_manager)
   , rocksdb_options_(std::move(rocksdb_options))
   , s3_util_()
   , s3_util_lock_()
@@ -467,10 +467,14 @@ AdminHandler::AdminHandler(
   , allow_overlapping_keys_segments_()
   , num_current_s3_sst_downloadings_(0)
   , stop_db_deletion_thread_(false)
-  , stop_db_incremental_backup_thread_ (false) {
+  , stop_db_incremental_backup_thread_(false) {
   if (db_manager_ == nullptr) {
     db_manager_ = CreateDBBasedOnConfig(rocksdb_options_);
   }
+
+  backup_manager_ = std::make_unique<ApplicationDBBackupManager>(
+    std::move(db_manager), FLAGS_rocksdb_dir, FLAGS_checkpoint_backup_batch_num_upload);
+
   folly::splitTo<std::string>(
       ",", FLAGS_allow_overlapping_keys_segments,
       std::inserter(allow_overlapping_keys_segments_,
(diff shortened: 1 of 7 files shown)


qiyanghe1998:
The design changes! I add a ```backup_manager``` inside the ```admin_handler``` to deal with all kinds of incremental backup. It contains a ```db_manager``` and some other necessary things for backup. The backup method for ```db_manager``` and ```application_db``` are both implemented in ```backup_manager```. Right now, the thread launched in ```admin_handler```'s constructor will invoke the backup method in ```backup_manager``` periodically so that we can continue to complete backup logic inside the backup methods. 

There is another problem left: how to deal with s3 settings? Previously, all the s3 settings are piggybacked by the backup request. There is no corresponding request right now. We need to set up s3 settings initially. Thus, I add two constructor for ```backup_manager```, one for those the s3 settings are known, the other one for those the s3 settings are unknown.

rocksdb_admin/admin_handler.cpp
```diff
@@ -497,7 +510,29 @@ AdminHandler::AdminHandler(
   }
 
   if (FLAGS_enable_async_incremental_backup_dbs) {
-    LOG(INFO) << "incremental backup gflag is enabled";
+    db_incremental_backup_thread_ = std::make_unique<std::thread>([this] {
```
mradwan, line 513: Lets move this entire thread to be a property/code of backup manager. Ideally we dont have to add so many changes to admin handler.

We can simply add a function there called stop() which will can set the atomic bool and join the thread.

rocksdb_admin/admin_handler.cpp
```diff
@@ -510,6 +545,11 @@ AdminHandler::~AdminHandler() {
     stop_db_deletion_thread_ = true;
     db_deletion_thread_->join();
   }
+
+  if (FLAGS_enable_async_incremental_backup_dbs) {
```
mradwan, line 549: This can be moved in backup manager

rocksdb_admin/admin_handler.cpp
```diff
@@ -2182,4 +2222,41 @@ std::vector<std::string> AdminHandler::getAllDBNames() {
     return db_manager_->getAllDBNames();
 }
 
+void AdminHandler::setS3Config(
```
mradwan, line 2225: This function is not required please call backup manager directly.

rocksdb_admin/admin_handler.cpp
```diff
@@ -2182,4 +2222,41 @@ std::vector<std::string> AdminHandler::getAllDBNames() {
...
+    const std::string& s3_bucket,
+    const std::string& s3_backup_dir_prefix,
+    const std::string& snapshot_host_port,
+    const uint32_t limit_mbs,
+    const bool include_meta) {
+  backup_manager_->setS3Config(std::move(s3_bucket), std::move(s3_backup_dir_prefix), 
+    std::move(snapshot_host_port), limit_mbs, include_meta);
+}
+
+bool AdminHandler::checkS3Object(
```
mradwan, line 2235: What is the goal from this function, can we move it into backup manager as well?

rocksdb_admin/admin_handler.h
```diff
@@ -153,6 +152,21 @@ class AdminHandler : virtual public AdminSvIf {
...
 
+  // Set S3 config for the Backup Manager
+  void setS3Config(const std::string& s3_bucket,
+                   const std::string& s3_backup_dir_prefix,
+                   const std::string& snapshot_host_port,
+                   const uint32_t limit_mbs,
+                   const bool include_meta);
+
+  // Used for testing incremental backup 
+  bool checkS3Object(const uint32_t limit_mbs = 50,
```
mradwan, line 163: Dont test here. instead use aws s3 ls command to test.

mradwan (changes requested):
I have some comments.

Lets minimize the amount of code inside admin handler for the backup and move everything to backup manager.

So far I dont see any thing that strictly required inside admin handler please let me know if I understand incorrectly.

rocksdb_admin/admin_handler.cpp
```diff
@@ -399,16 +379,6 @@ bool DeserializeKafkaPayload(
   }
 }
 
-CPUThreadPoolExecutor* S3UploadAndDownloadExecutor() {
```
mradwan, line 402: Keep here and pass to BackupManager through constructor.

rocksdb_admin/admin_handler.cpp
```diff
@@ -440,28 +410,38 @@ void deleteTmpDBs() {
...
+  std::shared_ptr<ApplicationDBManager> db_manager,
   RocksDBOptionsGeneratorType rocksdb_options): AdminHandler(
       std::move(db_manager), 
       [rocksdb_options](const std::string& dataset, const std::string& db) {
         return rocksdb_options(dataset);
   }) {}
 
 AdminHandler::AdminHandler(
-    std::unique_ptr<ApplicationDBManager> db_manager,
+    std::shared_ptr<ApplicationDBManager> db_manager,
```
mradwan, line 421: Keep this as a unique_ptr and pass as raw ptr to backup manager.

Since admin handler has ownership and backup manager is also owned by admin handler this is also okay/safe.

rocksdb_admin/admin_handler.cpp
```diff
@@ -440,28 +410,38 @@ void deleteTmpDBs() {
...
     RocksDBOptionsGenerator rocksdb_options)
-  : db_admin_lock_()
-  , db_manager_(std::move(db_manager))
+  : db_admin_lock_(std::make_shared<common::ObjectLock<std::string>>())
```
mradwan, line 423: Avoid changing this to a pointer and pass as raw ptr to db manager

rocksdb_admin/admin_handler.cpp
```diff
@@ -440,28 +410,38 @@ void deleteTmpDBs() {
...
+  , db_manager_(db_manager)
   , rocksdb_options_(std::move(rocksdb_options))
   , s3_util_()
   , s3_util_lock_()
   , meta_db_(OpenMetaDB())
   , allow_overlapping_keys_segments_()
   , num_current_s3_sst_downloadings_(0)
-  , stop_db_deletion_thread_(false) {
+  , stop_db_deletion_thread_(false)
+  , executor_(new CPUThreadPoolExecutor(FLAGS_num_s3_upload_download_threads,
```
mradwan, line 432: avoid adding this.

mradwan: Why did we remove S3UploadAndDownloadExecutor and added it here instead?

rocksdb_admin/admin_handler.h
```diff
@@ -53,11 +68,11 @@ class AdminHandler : virtual public AdminSvIf {
  public:
   // TODO deprecate after getting rid of all callsites
   AdminHandler(
-    std::unique_ptr<ApplicationDBManager> db_manager,
+    std::shared_ptr<ApplicationDBManager> db_manager,
```
mradwan, line 71: Avoid changing this to shared_ptr, pass as raw ptr to backup manager instead.

rocksdb_admin/admin_handler.h
```diff
@@ -53,11 +68,11 @@ class AdminHandler : virtual public AdminSvIf {
...
     RocksDBOptionsGeneratorType rocksdb_options);
 
   AdminHandler(
-    std::unique_ptr<ApplicationDBManager> db_manager,
+    std::shared_ptr<ApplicationDBManager> db_manager,
```
mradwan, line 75: ditto

rocksdb_admin/admin_handler.cpp
```diff
@@ -2182,4 +2158,8 @@ std::vector<std::string> AdminHandler::getAllDBNames() {
     return db_manager_->getAllDBNames();
 }
 
+bool AdminHandler::backupAllDBsToS3() {
```
mradwan, line 2161: Wouldn't this cause a race condition?

mradwan (changes requested):
Qiyang, we agreed that having a call from admin handler and backup manager thread together will cause a race condition. Lets remove the function in backup manager.

rocksdb_admin/admin_handler.h
```diff
@@ -37,6 +36,22 @@
...
 #endif
 #include "rocksdb/status.h"
+#if __GNUC__ >= 8
+#include "folly/executors/CPUThreadPoolExecutor.h"
+#include "folly/system/ThreadName.h"
+#else
+#include "wangle/concurrent/CPUThreadPoolExecutor.h"
+#endif
+
+#if __GNUC__ >= 8
```
mradwan, line 46: we should not have using statement in header files, cause these statements are going to propagate to anyone who uses this header file.

rocksdb_admin/admin_handler.h
```diff
@@ -171,13 +189,14 @@ class AdminHandler : virtual public AdminSvIf {
...
   std::unique_ptr<ApplicationDBManager> db_manager_;
+  std::unique_ptr<ApplicationDBBackupManager> backup_manager_;
   RocksDBOptionsGenerator rocksdb_options_;
   // S3 util used for download
   std::shared_ptr<common::S3Util> s3_util_;
   // Lock for protecting the s3 util
   mutable std::mutex s3_util_lock_;
   // db that contains meta data for all local rocksdb instances
-  std::unique_ptr<rocksdb::DB> meta_db_;
+  std::shared_ptr<rocksdb::DB> meta_db_;
```
mradwan, line 199: Can we keep this unique_ptr and pass as raw ptr like the others?

rocksdb_admin/admin_handler.h
```diff
@@ -152,6 +167,9 @@ class AdminHandler : virtual public AdminSvIf {
 
   // Get all the db names held by the AdminHandler
   std::vector<std::string> getAllDBNames();
+  
+  // Used for testing incremental backup 
+  bool backupAllDBsToS3();
```
mradwan, line 172: Let's not add any API just for testing.
This function is also not implemented.

rocksdb_admin/admin_handler.cpp
```diff
@@ -462,6 +442,12 @@ AdminHandler::AdminHandler(
   if (db_manager_ == nullptr) {
     db_manager_ = CreateDBBasedOnConfig(rocksdb_options_);
   }
+
+  if (FLAGS_enable_async_incremental_backup_dbs) {
+    backup_manager_ = std::make_unique<ApplicationDBBackupManager>(db_manager_.get(), S3UploadAndDownloadExecutor(), 
+      meta_db_, &db_admin_lock_, FLAGS_rocksdb_dir, FLAGS_checkpoint_backup_batch_num_upload);
```
mradwan, line 448: &meta_db_

mradwan (changes requested):
Almost there, few more minor comments.

Thanks Qiyang for addressing all the feedback

mradwan (approved):
Perfect, thank you!

PR #630, opened 2022-08-01
[EASY] Fix bug introduced because folly::stringPrintf takes only char*

mradwan:
I introduced this bug by passing strings to folly::stringPrintf.
This diff fixes that problem, the bug haven't made it to production which is good.

Test plan:
To test this change I am testing backup/restore flows on dev cluster

### DIFF
--- a/rocksdb_admin/admin_handler.cpp
+++ b/rocksdb_admin/admin_handler.cpp
@@ -981,7 +981,8 @@ void AdminHandler::async_tm_backupDBToS3(
   common::Timer timer(kS3BackupMs);
   LOG(INFO) << "S3 Backup " << request->db_name << " to " << request->s3_backup_dir;
   auto ts = common::timeutil::GetCurrentTimestamp();
-  auto local_path = folly::stringPrintf("%s%s%d/", getS3TmpPath(), request->db_name.c_str(), ts);
+  const std::string s3_tmp_path = getS3TmpPath();
+  auto local_path = folly::stringPrintf("%s%s%d/", s3_tmp_path.c_str(), request->db_name.c_str(), ts);
   boost::system::error_code remove_err;
   boost::system::error_code create_err;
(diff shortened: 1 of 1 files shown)


rocksdb_admin/admin_handler.cpp
```diff
@@ -981,7 +981,8 @@ void AdminHandler::async_tm_backupDBToS3(
   common::Timer timer(kS3BackupMs);
   LOG(INFO) << "S3 Backup " << request->db_name << " to " << request->s3_backup_dir;
   auto ts = common::timeutil::GetCurrentTimestamp();
-  auto local_path = folly::stringPrintf("%s%s%d/", getS3TmpPath(), request->db_name.c_str(), ts);
+  const std::string s3_tmp_path = getS3TmpPath();
```
jaricftw, line 984: What if we let`getS3TmpPath` return `const char*` instead? Would that be less error-prone?

jaricftw (approved):

jaricftw:
did we not have unit test coverage for that

rocksdb_admin/admin_handler.cpp
mradwan: There are other consumers for getS3TmpPath.

I think a better solution is to avoid folly::stringPrintf in favor of more modern functions.

mradwan: If you dont mind I will do this in a follow up change

mradwan:
@jaricftw I believe because its gated behind a flag its not tested
https://github.com/pinterest/rocksplicator/blob/7b0d1bdc417cd83c2ed212e5b1af592a416aca9d/rocksdb_admin/tests/admin_handler_test.cpp#L483

We need to run the test with --enable_integration_test

rocksdb_admin/admin_handler.cpp
jaricftw: np, follow up is good

PR #631, opened 2022-08-02
[EASY] Fix build for rocksplicator

mradwan:
Unfortunately we need these kind of header guard when building inside cosmos. this diff adds them.


qiyanghe1998 (approved):
Checked. Thanks for fixing.

PR #632, opened 2022-08-02
implement backup descriptor for incremental backup

qiyanghe1998:
Follow Qiyang He's design for incremental backup, this diff adds the backup descriptor for each backup.

Generally, it will create a folder for each backup distinguished by the timestamp. Under each folder, there will be newly added SST files (including those from compaction), backup descriptor, CURRENT, MANIFEST and OPTION. The backup descriptor contains a map from SST filename to timestamp and the range of SST files. You can refer to https://docs.google.com/document/d/1aRO1PJ8qUN-yn6iUK7cisk7qSAsKFhnoyig0_T89jTo/edit#heading=h.oxp5c6sx5wxx for more details.

### DIFF
--- a/rocksdb_admin/application_db_backup_manager.cpp
+++ b/rocksdb_admin/application_db_backup_manager.cpp
@@ -45,12 +45,18 @@ DEFINE_string(s3_incre_backup_prefix, "tmp/backup_test/incremental_backup_test/"
 DEFINE_int32(incre_backup_limit_mbs, 100, "the rate limit for s3 client");
 
 DEFINE_bool(incre_backup_include_meta, false, "whether to backup meta data on s3");
-namespace admin {
+
+namespace {
 
 const std::string kMetaFilename = "dbmeta";
 const std::string kS3BackupMs = "s3_backup_ms";
 const std::string kS3BackupFailure = "s3_backup_failure";
 const int kS3UtilRecheckSec = 5;
+const std::string backupDesc = "backup_descriptor";
+
+}
+
+namespace admin {
 
 ApplicationDBBackupManager::ApplicationDBBackupManager(
     ApplicationDBManager* db_manager,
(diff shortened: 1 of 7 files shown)


rocksdb_admin/application_db_backup_manager.cpp
```diff
@@ -185,17 +250,86 @@ bool ApplicationDBBackupManager::backupDBToS3(const std::shared_ptr<ApplicationD
...
+      sst_id_max = std::max(file_num, sst_id_max);
+      sst_id_min = std::min(file_num, sst_id_min);
+    }
+  }
+
+  // Create the new backup descriptor file
+  std::string backup_desc_contents;
+  makeUpBackupDescString(file_to_ts, sst_id_min, sst_id_max, backup_desc_contents);
+  common::FileUtil::createFileWithContent(formatted_checkpoint_local_path, backupDesc, backup_desc_contents);
+  checkpoint_files.emplace_back(backupDesc);
```
mradwan, line 331: There is no reason to use emplace_back here, just use push_back

mradwan: Does this guarantee that backup descriptor will only be uploaded after all files get uploaded?

mradwan (commented):
I am still working on reviewing the unit tests. but posting these comments here to unblock Qiyang from addressing them.

qiyanghe1998:
Addressed all the comments. Could you have another look? @mradwan Thanks!

rocksdb_admin/application_db_backup_manager.cpp
```diff
@@ -142,20 +156,33 @@ std::shared_ptr<common::S3Util> ApplicationDBBackupManager::createLocalS3Util(
...
+// convert file_to_ts map, last_ts to a json value.
+void ApplicationDBBackupManager::makeUpBackupDescJson(const std::unordered_map<std::string, int64_t>& 
+    file_to_ts, int64_t last_ts, Json::Value& contents) {
+  contents.clear();
+  Json::Value temp_map(Json::objectValue);
+  for (auto& item : file_to_ts) {
+    temp_map[item.first] = item.second;
+  }
+  contents[fileToTs] = temp_map;
+  contents[lastBackupTs] = last_ts;
```
mradwan, line 168: Great!

rocksdb_admin/application_db_backup_manager.cpp
```diff
@@ -185,17 +210,90 @@ bool ApplicationDBBackupManager::backupDBToS3(const std::shared_ptr<ApplicationD
...
   auto local_s3_util = createLocalS3Util(FLAGS_incre_backup_limit_mbs, FLAGS_s3_incre_backup_bucket);
-  std::string formatted_s3_dir_path = folly::stringPrintf("%s%s/%d/", 
+  std::string formatted_s3_dir_path_upload = folly::stringPrintf("%s%s/%ld/", 
     ensure_ends_with_pathsep(FLAGS_s3_incre_backup_prefix).c_str(), db->db_name().c_str(), ts);
   std::string formatted_checkpoint_local_path = ensure_ends_with_pathsep(checkpoint_local_path);
+
+  std::unordered_map<std::string, int64_t> file_to_ts;
+  const string sst_suffix = ".sst";
+
+  if (last_ts >= 0) {
```
mradwan, line 234: In this case if machine restarts we wouldnt have last_ts.

Can you leave a comment that we need to handle this code doesn't handle this case well?

Thanks!

mradwan (approved):
Thanks for addressing the coments. I left one comment regarding leaving another comment if you can leave it that would be great.

Thank you!

qiyanghe1998:
Pushed a new version with backup starter

